Issue #10, opened 2025-10-12
Add automatic h3 (and others) partitioning based on a max size


cholmes:
We should let users just specify a maximum size per partition, or perhaps number of partitions (or ideally size on disk, but that's trickier) and automatically split based on h3 or quadkey or others. There should be some code to do this in https://github.com/opengeos/open-buildings

PR #23, opened 2025-10-19
feat(add): add H3 support with computed column abstraction

nlebovits:
**Related Issue**

#10 (but does _not_ close it)

**Description**
Adds H3 cell generation and partitioning. If H3 cell is missing before partitioning, it's automatically added. Resolution is configurable from 0 to 15 (default is 9), along with hive style partitioning.

 realized we should probably have a generic function to add additional columns like bbox, h3, s2, etc., so I created one and refactored the `add_bbox_column` and `add_bbox` to use it before adding H3 support. This will give us flexibility but keep boilerplate to a minimum.

Code now supports metadata for multiple spatial indices (bbox + H3).

Tests, readme, and changelog all updated.

**I have:**

- [x]  Confirmed that tests pass
- [x] Linted and formatted my code
- [x] Updated the changelog

### DIFF
--- a/README.md
+++ b/README.md
@@ -236,6 +236,8 @@ gpio add admin-divisions buildings.parquet output.parquet --dry-run
 
 The `partition` commands provide different options to partition GeoParquet files into separate files based on column values.
 
+**Smart Analysis**: All partition commands automatically analyze your partitioning strategy before execution, calculating statistics and providing recommendations. Use `--preview` for dry-run analysis without creating files. Use `--force` to override warnings, or `--skip-analysis` for performance.
+
 #### partition string
 
 Partition a GeoParquet file by string column values. You can partition by full column values or by a prefix (first N characters). This is useful for splitting large datasets by categories, codes, regions, etc.
@@ -269,7 +271,7 @@ Options:
 
 Example usage:
 ```bash
-# Preview partitions by first character of MGRS codes
+# Analyze and preview partition strategy (dry-run)
 gpio partition string input.parquet --column MGRS --chars 1 --preview
 
 # Partition by full column values
@@ -286,6 +288,8 @@ gpio partition string input.parquet output/ --column region --chars 1 --hive
 
 Partition a GeoParquet file by H3 hexagonal cells at a specified resolution. Automatically adds H3 column if it doesn't exist.
 
+By default, the H3 column is **excluded** from the output files (since it's redundant with the partition path), except when using Hive-style partitioning where it's included. Use `--keep-h3-column` to explicitly keep the column in all cases.
+
 ```
 $ gpio partition h3 --help
 Usage: gpio partition h3 [OPTIONS] INPUT_PARQUET [OUTPUT_FOLDER]
@@ -295,6 +299,10 @@ Usage: gpio partition h3 [OPTIONS] INPUT_PARQUET [OUTPUT_FOLDER]
   Creates separate GeoParquet files based on H3 cell prefixes at the
   specified resolution. If the H3 column doesn't exist, it will be
   automatically added before partitioning.
+
+  By default, the H3 column is excluded from output files (since it's
+  redundant with the partition path) unless using Hive-style partitioning.
+  Use --keep-h3-column to explicitly keep the column in all cases.
 
   Use --preview to see what partitions would be created without actually
   creating files.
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -240,6 +282,52 @@ gpio partition string input.parquet output/ --column mgrs_code --chars 2
...
+  Creates separate GeoParquet files based on H3 cell prefixes at the
+  specified resolution. If the H3 column doesn't exist, it will be
+  automatically added before partitioning.
+
+  Use --preview to see what partitions would be created without actually
+  creating files.
+
+Options:
+  --h3-name TEXT           Name of H3 column to partition by (default:
+                           h3_cell)
```
cholmes, line 312: Curious on the choice of h3_cell - have you seen that be widely used, or you just picked a sensible one? (Or did you just improve mine - looks like I said h3-cell, but I agree h3_cell is a better choice for wider database compatibility)

I ask because I want to 'standardize' it a bit with a vecorel extension - then we can do quick validation of a file advertises that it implements h3 vecorel extension and has the right column name. I think this is totally fine for now, just we may want to change the default in the future to match vecorel (but we will likely be the ones to make that choice)

README.md
```diff
@@ -240,6 +282,52 @@ gpio partition string input.parquet output/ --column mgrs_code --chars 2
...
+  --resolution INTEGER     H3 resolution for partitioning (0-15, default: 9)
```
cholmes, line 313: So I just tested this with a 66mb file and it looks like I got 6000+ parquet files and it took 20 minutes. We could consider lowering the default. But I think the main thing to do is to do more automatic / dynamic things here. #10 likely covers most of this, but once we have that mechanism in there then I think we probably calculate the default automatically for the data input, based on [best practices](https://github.com/opengeospatial/geoparquet/blob/main/format-specs/distributing-geoparquet.md)

geoparquet_io/cli/main.py
```diff
@@ -388,6 +390,89 @@ def add_bbox(
...
+@click.argument("input_parquet")
+@click.argument("output_parquet")
+@click.option("--h3-name", default="h3_cell", help="Name for the H3 column (default: h3_cell)")
+@click.option(
+    "--resolution",
+    default=9,
+    type=click.IntRange(0, 15),
+    help="H3 resolution level (0-15). Res 7: ~5km², Res 9: ~105m², Res 11: ~2m², Res 13: ~0.04m². Default: 9",
+)
+@click.option(
```
cholmes, line 403: So in the refactor we got these [decorators](https://github.com/cholmes/geoparquet-io/blob/main/geoparquet_io/cli/decorators.py) for the click commands, to not replicate so much code. But it looks like they weren't really used. Nothing needed for this PR, as it's consistent with how things are done, but just putting this in to remind me to open a ticket to actually make use of the decorators, so we can change things in one place.

geoparquet_io/cli/main.py
```diff
@@ -490,5 +575,78 @@ def partition_string(
     )
 
 
+@partition.command(name="h3")
```
cholmes, line 632: I love the idea of automatically doing the add h3 vs having to specifically add it. But I do think it'd be good to have an option about whether you include the column name in the output.

Like if I want to partition by h3 I think I often don't actually want to have an extra h3 column in there. If I'm first doing 'add' then it makes sense that I opted into it. But I'd at least like the option to not include it in the output.

And I lean towards the default being to not include it in the output files, but don't feel super strongly. At least if they don't have a hive partition - I suspect for hive partition you may need to include it, since the folder name is the attribute to be queried on.

README.md
cholmes: We also might proactively put in a warning if it's going to generate a ton of small files. Like right now we say:

```
% gpio partition h3 c-america.parquet c-h3        
Processing 722,109 features...
Successfully added H3 column 'h3_cell' (resolution 9) to: /var/folders/s6/ctf1vwb95cnb44vtlbg4r6c40000gp/T/h3_enriched_c-america.parquet
Partitioning by H3 cells at resolution 9 (column: 'h3_cell')
```

Could just report out how many unique partitions it'll be partitioned to, and perhaps elevate to a warning if it's going to be a ton (over 1000? Or something more dynamic based on best practices, like average file size)

Also we might have the default 'partition' be lower than the default 'add' - then if people want to partition further or search by it it's already there.

cholmes (approved):
Looks great - thanks for the contribution! 

Put in a few comments - feel free to update the PR, or to merge and ticket separately, or any combo thereof. Main things were:

- report / warn it's going to be a ton of small files
- option to not include h3 column in output file if you're partitioning and adding it automatically.

I also ticketed two things in this review - related but can be handled on their own.
 * #25 - we need to update the partitioned files bbox better. This was my bad, just noticed it while doing this.
 * #26 - sort by h3 - can have it work similar to stuff here, and I think could be nice to do by default when a user is partitioning by it.

README.md
nlebovits: No, honestly I just took your column name and figured snake case was preferable for consistency. Maybe a poll in CNG Slack or LinkedIn or something?

geoparquet_io/cli/main.py
nlebovits: This makes sense to me. At a minimum, it would mean a smaller file. So we'll have the default be not to add it, unless they're opting for hive partitioning.

Issue #11, opened 2025-10-12
parition by kdtree


cholmes:
Info at https://dewey.dunnington.ca/post/2024/partitioning-strategies-for-bigger-than-memory-spatial-data/

nlebovits:
Dewey's post mentions that the runtime for kdtree partitioning of 130 million building footprints in the US is ~10 minutes. I could see this becoming prohibitive for something like the full building footprints cell. Whoever gets to this (might be me) should add a warning output that, for datasets of, say, 50 million records or more, it's worth considering hierarchical partitioning strategies like country + kdtree or country + h3 cell + kdtree.

PR #30, opened 2025-10-21
feat(partition): add KD-tree spatial partitioning

nlebovits:
Implements KD-tree partitioning following the H3 implementation patterns:

- Add  with recursive CTE algorithm for binary partition IDs
- Add  for spatial partitioning by KD-tree cells
- CLI commands:  and
- Support iterations 1-20 (2^1 to 2^20 partitions), default 9 (512 partitions)
- Large dataset protection: warn/block datasets >50M rows, recommend hierarchical partitioning
- Column behavior: excluded by default, included for Hive (matches H3 pattern)
- Comprehensive test suite with 22 tests covering add, partition, and validation
- Refactor  recommendations to be generic (remove H3-specific language)

Algorithm based on: https://duckdb.org/2024/09/09/spatial-extension.html Performance scales with dataset_size × iterations.

**Related Issue(s):**

- #11 

**Description:**

**PR Checklist:**

- [x] Code is formatted
- [x] Tests pass
- [x] Changes are added to CHANGELOG.md

### DIFF
--- a/tests/test_kdtree.py
+++ b/tests/test_kdtree.py
@@ -14,7 +14,7 @@ class TestAddKDTreeColumn:
     """Test suite for add kdtree column command."""
 
     def test_add_kdtree_column_basic(self, buildings_test_file, temp_output_file):
-        """Test adding KD-tree column with default partitions (512)."""
+        """Test adding KD-tree column with auto-selection (default behavior)."""
         runner = CliRunner()
         result = runner.invoke(
             add,
@@ -22,16 +22,16 @@ class TestAddKDTreeColumn:
         )
         assert result.exit_code == 0
         assert os.path.exists(temp_output_file)
+        assert "Auto-selected" in result.output
 
         # Verify kdtree_cell column was added
         table = pq.read_table(temp_output_file)
         assert "kdtree_cell" in table.schema.names
 
-        # Verify binary strings are 10 characters (512 partitions = 9 iterations + starting '0')
+        # Verify binary strings are valid (length depends on auto-selection)
         kdtree_values = table.column("kdtree_cell").to_pylist()
         for value in kdtree_values:
             if value is not None:
-                assert len(value) == 10
                 assert all(c in "01" for c in value)
                 assert value.startswith("0")  # All start with '0'
 
@@ -116,7 +116,7 @@ class TestAddKDTreeColumn:
             ["kdtree", buildings_test_file, temp_output_file, "--verbose"],
         )
         assert result.exit_code == 0
-        assert "Adding column" in result.output or "Processing" in result.output
+        assert "Auto-selected" in result.output
 
 
 class TestPartitionKDTree:


geoparquet_io/core/partition_by_kdtree.py
```diff
@@ -0,0 +1,222 @@
...
+    Performance Note: Runtime scales with dataset size × iterations.
+    For datasets > 50M rows, consider hierarchical partitioning: partition by
+    a coarse key (country/region) first, then apply KD-tree within each partition.
+
+    Args:
+        input_parquet: Input GeoParquet file
+        output_folder: Output directory
+        kdtree_column_name: Name of KD-tree column (default: 'kdtree_cell')
+        iterations: Number of recursive splits (1-20, default: 9)
+                   iterations=5: 32 partitions, iterations=9: 512 partitions
```
cholmes: So the idea here is that the result will be a number of partitions equal to 2 to the iterations power? I admit I haven't really looked into the kdtree algorithm, is there no way to get like 750 partitions? 

It does feel pretty unintuitive to me to have to figure out what value to put here - most people won't know an ideal partition size, let alone doing the math to go from their data set to the right size. 

I think an ideal would be for a default iterations value to be more automatic. I think ideally a user doesn't provide any value and it picks an ideal partition size and figures out the iterations needed to reach that. 

The next thing I'd see them provide is either max (or approx?) number of rows per partition, or ideally mb size of partition (though the latter is more challenging. And so we'd have a default value for that of like 120k, in line with the geoparquet best practice recommendation. And then user can change that. Or if they really want they can set the iterations (likely mutually exclusive to max partition size).

geoparquet_io/core/partition_by_kdtree.py
```diff
@@ -0,0 +1,222 @@
...
+            f"\nThis dataset has {total_rows:,} rows. Computing KD-tree partition IDs{preview_note} "
+            f"requires processing\nthe entire dataset {iterations} times (runtime scales with "
+            f"dataset size × iterations).\n\n"
+            f"This would take an extremely long time and may crash your system.\n",
+            fg="yellow",
+        )
+        error_msg += click.style("\nRecommended approach:", fg="cyan", bold=True)
+        error_msg += click.style(
+            "\n  1. Use hierarchical partitioning: partition by a coarser key first "
+            "(country/region/state),\n     then apply KD-tree within each partition.\n",
```
cholmes: This recommendation strikes me as less useful unless we have the 'auto' as I recommended above. It'd automatically split out each country / state into 512 partitions, while most of them for most data sets likely don't need to be split up more. But if it is 'auto' then it'd just leave most of them alone, so the recommendation would make sense.

geoparquet_io/core/partition_by_kdtree.py
```diff
@@ -0,0 +1,222 @@
...
+    # Runtime scales with dataset_size × iterations, so warn/block for large datasets
+    # This applies to BOTH preview and actual partitioning since preview needs to compute values too
+    if not column_exists and not force and total_rows > 50_000_000:
+        preview_note = " (including preview)" if preview else ""
+        error_msg = click.style("\n⚠️  Large Dataset Warning\n", fg="yellow", bold=True)
+        error_msg += click.style(
+            f"\nThis dataset has {total_rows:,} rows. Computing KD-tree partition IDs{preview_note} "
+            f"requires processing\nthe entire dataset {iterations} times (runtime scales with "
+            f"dataset size × iterations).\n\n"
+            f"This would take an extremely long time and may crash your system.\n",
```
cholmes: I don't think I'd be quite so dire here, unless the chances of crashing people's system are quite high? I think the cool thing about kdtree is that it can provide a pretty ideal spatial partition, and I think many people will wait for that. I think just setting their expectations that it can take tens of minutes (or hours? Would be cool to provide some rough estimate). And also what settings they may need to make it work better.

Could also be nice to update some on progress, like on each iteration completed, so that people can get a sense that things are happening and how close it is to finishing.

geoparquet_io/core/partition_by_kdtree.py
nlebovits: I made a couple of changes to the PR:
- Switched from specifying the # of iterations to the # of partitions (2, 4, 8, etc.)
- Added auto-partition functionality that calculates the ideal number of partitions based on a specified number of rows (120,000 is recommended). Ideal partition sizes can either be approximated with a random sample of 100,000 rows or exactly, depending on user preference. In theory, we could also implement functionality to partition by a target size in MB, but I don't want to clutter up the tool.
- Added printing of partition sizes in rows + MBs

geoparquet_io/core/partition_by_kdtree.py
nlebovits: Auto has been implemented!

geoparquet_io/core/partition_by_kdtree.py
nlebovits: PR updated accordingly. I toned down the language + added some basic progress printing so people can see what # iteration they're on. It also runs waaaay faster if we determine partitions using random sampling instead of an exact count.

nlebovits:
> Nice work! So excited to see this made more accessible. Put in some suggestions on how to potentially make it even more user friendly.
> 
> One thing I've been thinking about for all partitions is to be able to put a basename in the parquet file. The partition files are all named like 0111101101.parquet, which makes sense. But I think it can be nice to let people call them all like fields_0111101101.parquet I think this is true with string partitions too, like fields_NL.parquet

I'm going to write this up as a separate ticket because it seems like functionality we'd want across all partitioning approaches, not just kd-tree. But agreed that it's useful.

Issue #29, opened 2025-10-21
Feature: Add modular `inspect` command for Geoparquet files


nlebovits:
Add an `inspect` command to easily examine Geoparquet files during or after processing.

## Use Cases
1. **Standalone inspection**: `inspect <file_path>` to examine existing files
2. **Inline inspection**: `--inspect` flag on I/O commands to view input/output during conversion

## Proposed Arguments
- `--pretty` - Pretty-printed data view
- `--stats` - Data distribution statistics
- `--bbox` - Bounding box information
- `--metadata` - File metadata
- `--head` / `--tail` - First/last N rows

This would make it much easier to quickly validate and explore Geoparquet data without external tools.

PR #31, opened 2025-10-21
feat(cli): add inspect command for fast file examination

nlebovits:
**Related Issue(s):**

- #29

**Description:**
Add  command with metadata display, data preview
(--head/--tail), column statistics (--stats), and JSON output (--json). 
Uses Rich library for formatted terminal output. 
Shows EPSG:4326 as default CRS when not explicitly specified in metadata.

**PR Checklist:**

- [x] Code is formatted
- [x] Tests pass
- [ ] Changes are added to CHANGELOG.md

### DIFF
--- a/geoparquet_io/cli/main.py
+++ b/geoparquet_io/cli/main.py
@@ -137,8 +137,8 @@ def check_row_group_cmd(parquet_file, verbose):
 # Inspect command
 @cli.command()
 @click.argument("parquet_file", type=click.Path(exists=True))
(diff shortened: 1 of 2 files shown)


nlebovits:
Added the ability to check compression type and parquet version. Noting that the head/tail thing is kind of odd. Click's `flag_value` para should allow optional integer args, but I was having trouble getting it working--it kept throwing "requires an argument" errors. I'm going to leave it for now and focus on bigger tickets, but perhaps it's something we figure out in the future.

Issue #16, opened 2025-10-19
More built-in country boundaries


cholmes:
Right now we just use overture - would be nice to put up a few different boundary files on source.coop and let users pick by name.

Also would be good to get metadata in vecorel about which was used.

nlebovits:
What would be arguments for choosing different boundary datasets? FAO probably makes sense as a default since it's the UN standard. That'll be the first one I put on Source. Maybe having examples of when someone would one to use each would be helpful for a guide.

Thoughts on me just uploading the FAO data with an added region column (if that's not already in the data)? We could have an option to partition by country with an optional flag like --include-regions. I have a hard time imagining a use case where someone would want the regions and _not_ the countries.

Issue #15, opened 2025-10-19
Additional admin level splitting


cholmes:
Would be good to go to at least admin level 2 in add / partition.

PR #38, opened 2025-10-24
Feature/add admin levels

nlebovits:
**Related Issue(s):**
- #15 
- #16

**Description:**
- Add partition admin: spatial join + partition by GAUL/Overture boundaries
- Add admin-divisions: add admin columns via spatial join
- Support hierarchical levels
- Add spatial extent filtering for remote datasets

**PR Checklist:**

- [x] Code is formatted
- [x] Tests pass
- [x] Changes are added to CHANGELOG.md

### DIFF
--- a/docs/guide/add.md
+++ b/docs/guide/add.md
@@ -133,15 +133,29 @@ gpio add admin-divisions buildings.parquet output.parquet --dataset overture \
 
 Two remote admin boundary datasets are supported:
 
-| Dataset | Columns Added | Description |
-|---------|--------------|-------------|
-| `gaul` (default) | `admin:continent`, `admin:country`, `admin:department` | GAUL L2 Admin Boundaries - worldwide |
-| `overture` | `admin:country`, `admin:region` | Overture Maps Divisions (219 countries, 3,544 regions) - [docs](https://docs.overturemaps.org/guides/divisions/) |
+| Dataset | Standard | Columns Added | Description |
+|---------|----------|---------------|-------------|
+| `gaul` (default) | GAUL naming + ISO 3166-1 alpha-3 | `admin:continent`, `admin:country`, `admin:department` | FAO Global Administrative Unit Layers (GAUL) L2 - worldwide coverage with standardized naming |
+| `overture` | **Vecorel compliant** (ISO 3166-1/2) | `admin:country_code`, `admin:subdivision_code` | Overture Maps Divisions with ISO 3166 codes (219 countries, 3,544 regions) - [docs](https://docs.overturemaps.org/guides/divisions/) |
+
+### Vecorel Compliance (Overture Dataset Only)
+
+The `overture` dataset follows the [Vecorel administrative division extension](https://vecorel.org/administrative-division-extension/v0.1.0/schema.yaml) specification with standardized ISO codes:
+
+- **`admin:country_code`** (REQUIRED): ISO 3166-1 alpha-2 country code (e.g., "US", "AR", "DE")
+- **`admin:subdivision_code`**: ISO 3166-2 subdivision code WITHOUT country prefix (e.g., "CA" not "US-CA")
+
+The tool automatically transforms Overture's native region codes (e.g., "US-CA") to strip the country prefix for Vecorel compliance.
+
+**Note:** The GAUL dataset uses FAO's standardized naming system but is NOT Vecorel compliant:
+- Has ISO 3166-1 alpha-3 codes (e.g., "TZA"), but Vecorel requires alpha-2 (e.g., "TZ")
+- Uses GAUL's standardized naming for subnational units, not ISO 3166-2 codes
+- Columns: `admin:continent` (continent name), `admin:country` (GAUL country name), `admin:department` (GAUL L2 name)
 
 ### Notes
 
-- Uses the [administrative division extension](https://github.com/fiboa/administrative-division-extension) from [fiboa](https://github.com/fiboa)
-- GAUL dataset: [source.coop GAUL L2](https://data.source.coop/nlebovits/gaul-l2-admin/)
+- **Overture dataset**: Vecorel compliant with ISO 3166-1 alpha-2 and ISO 3166-2 codes
+- **GAUL dataset**: FAO standardized naming system - [source.coop GAUL L2](https://data.source.coop/nlebovits/gaul-l2-admin/)
 - Performs spatial intersection to assign admin divisions based on geometry
 - Requires internet connection to access remote datasets
 - Uses spatial extent filtering and bbox columns for optimization
--- a/geoparquet_io/core/add_admin_divisions_multi.py
+++ b/geoparquet_io/core/add_admin_divisions_multi.py
@@ -230,13 +230,19 @@ def add_admin_divisions_multi(
     subquery_cols_str = ", ".join(subquery_cols)
 
     # Build SELECT clause for admin columns
-    # Create column aliases based on level names
+    # Create column aliases based on level names with optional transformations
     admin_select_parts = []
     for i, (level, col) in enumerate(zip(levels, partition_columns)):
-        # Use level name as the output column name with admin: prefix
-        output_col_name = f"admin:{level}"
-        # Handle struct field access (e.g., names['primary']) vs simple column names
-        if "[" in col or "(" in col:
+        # Get the output column name (may be customized for Vecorel compliance)
+        output_col_name = dataset.get_output_column_name(level)
+
+        # Check if a column transformation is needed (e.g., stripping country prefix)
+        col_transform = dataset.get_column_transform(level)
+
+        if col_transform:
+            # Apply transformation - the transform is a full SQL expression
+            admin_select_parts.append(f'{col_transform} as "{output_col_name}"')
+        elif "[" in col or "(" in col:
             # SQL expression - reference the alias from subquery
             admin_select_parts.append(f'b._col_{i} as "{output_col_name}"')
         else:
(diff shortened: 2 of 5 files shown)


docs/guide/add.md
```diff
@@ -96,23 +96,55 @@ gpio add kdtree input.parquet output.parquet --verbose
...
+
+| Dataset | Columns Added | Description |
+|---------|--------------|-------------|
+| `gaul` (default) | `admin:continent`, `admin:country`, `admin:department` | GAUL L2 Admin Boundaries - worldwide |
+| `overture` | `admin:country`, `admin:region` | Overture Maps Divisions (219 countries, 3,544 regions) - [docs](https://docs.overturemaps.org/guides/divisions/) |
+
+### Notes
 
-Default countries source: [source.coop admin boundaries](https://data.source.coop/cholmes/admin-boundaries/countries.parquet)
+- Uses the [administrative division extension](https://github.com/fiboa/administrative-division-extension) from [fiboa](https://github.com/fiboa)
```
cholmes: We should switch this to [vecorel admin extension](https://github.com/vecorel/administrative-division-extension).

docs/guide/add.md
```diff
@@ -96,23 +96,55 @@ gpio add kdtree input.parquet output.parquet --verbose
...
 ```
 
-Uses the [administrative division extension](https://github.com/fiboa/administrative-division-extension) from [fiboa](https://github.com/fiboa).
+### Datasets
+
+Two remote admin boundary datasets are supported:
+
+| Dataset | Columns Added | Description |
+|---------|--------------|-------------|
+| `gaul` (default) | `admin:continent`, `admin:country`, `admin:department` | GAUL L2 Admin Boundaries - worldwide |
```
cholmes: I recently realized that colons are a bad option for hive partitioning, so I think we should consider a different delimiter. See https://github.com/vecorel/administrative-division-extension/issues/4 for a bit about it.

We should also add these new to the vecorel admin extension. Ticket is at https://github.com/vecorel/administrative-division-extension/issues/2 - we can add on there. I don't think we need to wait on that update for a merge - this library can run ahead, but I think it'd be good to have them match, but that can just be update vecorel extension to what we do here.

One of the benefits of that is it'll provide a clear place to document what these mean. Like I'm not sure how 'department' is different from 'region' (from overture) just from looking at it. With a vecorel extension we can just point people at that, and explain there.

docs/guide/add.md
```diff
@@ -96,23 +96,55 @@ gpio add kdtree input.parquet output.parquet --verbose
...
+| `overture` | `admin:country`, `admin:region` | Overture Maps Divisions (219 countries, 3,544 regions) - [docs](https://docs.overturemaps.org/guides/divisions/) |
+
+### Notes
 
-Default countries source: [source.coop admin boundaries](https://data.source.coop/cholmes/admin-boundaries/countries.parquet)
+- Uses the [administrative division extension](https://github.com/fiboa/administrative-division-extension) from [fiboa](https://github.com/fiboa)
+- GAUL dataset: [source.coop GAUL L2](https://data.source.coop/nlebovits/gaul-l2-admin/)
```
cholmes: Should we put these in a common place? I'm totally fine if we moved the overture one under yours (probably worth running it again and document what version of overture it was run on).

docs/guide/add.md
```diff
@@ -96,23 +96,55 @@ gpio add kdtree input.parquet output.parquet --verbose
...
+- Performs spatial intersection to assign admin divisions based on geometry
+- Requires internet connection to access remote datasets
```
cholmes, line 160: Is this now required? My original implementation just used a local one. I like that it can now work without a user having to figure that out, but I think it'd be good to prefer local, as it'll generally be faster.

One idea is that we can have a default cache location for the admin files (let users set it differently if they want, and provide an argument or environment variable). So it'd just fetch it for them, and if it's in the cache then an internet connection isn't required.

docs/guide/partition.md
```diff
@@ -76,22 +76,56 @@ If KD-tree column doesn't exist, it's automatically added.
...
-# Custom admin column
-gpio partition admin input.parquet output/ --column iso_code
+# Hive-style multi-level (creates continent=Africa/country=Kenya/department=Accra/)
+gpio partition admin input.parquet output/ --dataset gaul \
+    --levels continent,country,department --hive
 
-# Hive-style
-gpio partition admin input.parquet output/ --hive
+# Overture Maps by country and region
+gpio partition admin input.parquet output/ --dataset overture --levels country,region
```
cholmes, line 117: What happens if I want both gaul and overture? Like continent, country, department and region? My guess is that the second one run will overwrite the first one for the country? But the ones that area in the first and not the second will stay as the first? 

I think that can be fine, but we should document it - I think some people may be interested in getting multiple admin levels.

README.md
```diff
@@ -108,8 +111,11 @@ gpio sort hilbert input.parquet sorted.parquet
 # Partition by H3 cells
 gpio partition h3 large.parquet output_dir/ --resolution 7
 
-# Partition by country
-gpio partition admin buildings.parquet by_country/
+# Partition by admin boundaries with spatial extent filtering
+gpio partition admin buildings.parquet by_admin/ --dataset gaul --levels continent,country
```
cholmes, line 115: I do think it'd be nice to still have the ability to point directly at a country file, vs only having the hard coded ones. Like if someone wants to try out geonames, or there's an update to one of the files that we don't have. Like just let dataset take a path (or have a mutually exclusive dataset-path flag)

docs/guide/add.md
nlebovits: In my first implementation, I had a local file option and caching, but I ended up stripping them out for this PR.

The local file option has slightly different needs than the remote files, because the schema has to be flexible (user should be able to pass a path and column(s) of interest) and the dataset should be in different formats. I was wondering if I should first work on the ticket that converts vector data to parquet and _then_ work on the local files thing--the intermediate conversion step isn't necessarily required, but it might speed up joins to have the admin data well formatted and indexed. Just don't know if the overhead of read --> convert --> join is worth it versus just slower read --> slower join (maybe with geopandas or something).

Caching definitely speeds up the processing but it has somewhat complicated logic because of remote dataset versioning and the spatial filtering. I thought it warranted it's own ticket. For instance, with the current logic, a user would only cache the spatially filtered part of the admin boundaries dataset. So, before using the cached data, we'd have to check 1) if what was in the cache fully covered the new data being passed to the command, and 2) if what was in the cache was the latest version (in the case of Overture Maps). Current read time from remote is in the range of a couple of seconds, at least for small datasets, so I didn't want to bother with it yet, but it should be implemented at some point.

So, tl;dr: both of these are good but should be their own tickets bc of complexity.

docs/guide/partition.md
nlebovits: Maybe the simplest option is to allow people to optionally designate column names or prefixes, like "gaul_country" or "overture_country", and then just document composition of the two joins in a row (add first admin boundaries --> add second admin boundaries --> partition).

README.md
nlebovits: Yeah, re: my comment above, I'll open a new ticket to add this functionality as an additional method in the admin dataset class.

docs/guide/add.md
nlebovits: Yeah, I'd be fine to centralize stuff in a common location. It gives the datasets a bit more credibility if they're not hosted under a personal account, too. What's the best place?

Which Overture run are you talking about? The code currently pulls Overture boundaries directly from their s3 bucket.

cholmes: I was talking about https://source.coop/cholmes/admin-boundaries - didn't realize you changed the code to pull directly from their boundaries. But it's probably a better route to stream directly from them - I pulled them out since I needed them locally originally, then I uploaded them since I wanted to use that same dataset automatically. I remember the admin boundaries being a bigger dataset than needed for just the countries. 

As for a common place, the most flexible neutral place on source coop is probably https://source.coop/tge-labs - we can make https://source.coop/tge-labs/admin-boundaries and put GAUL and any others there. I can set that up if that sounds good. I think we could perhaps put Overture there even if we pull from s3 (which I think is better since it'll stay up to date), just so 'normal' users can more easily access a countries file - I found it unintuitive to get the right incantation to pull out just the country borders

docs/guide/add.md
cholmes: Ok, cool. Happy to have them in their own tickets and have this one merge

And for caching I was thinking of doing it more simply - just cache the full dataset in the default case and then subset from the local copy. And then just offer a mode that doesn't use the local cache. It's not a huge dataset, so I think caching on the first call (with a line informing the user that it's happening) is fine. I was also just thinking of not worrying about automated versioning - could let users just delete their cached one if they wanted a new copy.

docs/guide/partition.md
cholmes: Sure, that approach sounds good.

docs/guide/add.md
nlebovits: Looks like Overture's default country and region datasets are compliant with Vecorel with only a slight change (dropping the country prefix from the region column). I went ahead and made that the default admin boundaries dataset. GAUL is not compliant, although we could create some kind of mapping in the future if we thought it was worth the effort. They're using a different version of ISO3 for the country names.

cholmes: > GAUL is not compliant, although we could create some kind of mapping in the future if we thought it was worth the effort. They're using a different version of ISO3 for the country names.

Yeah, I think it'd be nice to translate any input into the consistent vecorel format. But I have some desire to tweak the vecorel format, and also want to fully implement vecorel metadata and validation. So I think it's good for now, and we can come back to it later - for now it's fine that it tries to align.

Once the vecorel thing gets solid I could also see just having vecorel-compliant versions of the admin boundaries in the tge-labs source coop repo. Then the geoparquet-io code doesn't have to do anything fancy, just point at the one that's already in the right format.

Issue #34, opened 2025-10-23
Feat: make file a STAC item/collection


nlebovits:
Add tooling to allow the user to convert GeoParquet files to valid STAC items or collections, both as an independent module and as a flag where appropriate (e.g., `partition`).

## Background
We're already collecting most required STAC metadata during processing (geometry, bbox, projection, etc.). The main additions needed are:
- `id` (can derive from filename)
- `description` (user-provided or auto-generated)
- `datetime` (can default to UTC processing time or file timestamp)

## Implementation approach
- Cannibalize/adapt code from [fio-stac](https://github.com/developmentseed/fio-stac) (Development Seed, deprecated)
- Add `pystac` as dependency
- Generate static STAC catalogs (JSON files for upload to S3/Source Cooperative)
- Extract GeoParquet `geo` metadata to populate STAC properties

## Handling partitioned data
- **Single file** → Single STAC Item
- **Partitioned file** (country, H3, KD-tree, etc.) → STAC Collection + Item per partition
  - Collection: Overall dataset metadata
  - Items: One per partition with partition-specific bbox, geometry, and asset href

## Auto-generating links and assets
User provides bucket prefix (e.g., `--bucket s3://source.coop/my-org/roads/`), and tool automatically constructs:
- Item JSON `self` links: `s3://source.coop/my-org/roads/usa.json`
- Asset `href` values: `s3://source.coop/my-org/roads/usa.parquet`
- Public URLs where applicable (converting S3 URIs to HTTPS URLs for Source Cooperative, etc.)

Optional `--public-url` flag for when bucket has different public-facing URL:
```bash
--bucket s3://my-bucket/path/ --public-url https://data.source.coop/my-org/
```

This approach works consistently for both single files and partitioned outputs without requiring manual href specification.

## Required STAC Item fields
- `type`: "Feature"
- `stac_version`: "1.0.0"
- `id`: Derived from filename or dataset name + partition key
- `geometry`: GeoJSON geometry from GeoParquet spatial extent
- `bbox`: Bounding box array [min_lon, min_lat, max_lon, max_lat]
- `properties.datetime`: ISO 8601 timestamp (processing time or file timestamp)
- `links`: Array with `self` link (auto-generated from bucket prefix)
- `assets`: Dictionary with at least one asset pointing to the GeoParquet file
  - `href`: Auto-generated from bucket prefix + filename
  - `type`: Always `"application/vnd.apache.parquet"`
  - `roles`: Always `["data"]`
[...]

PR #57, opened 2025-11-08
feat(stac): add STAC Item and Collection generation

nlebovits:
## Description
Add  command to generate STAC metadata from GeoParquet files:
- Single file → STAC Item JSON
- Partitioned directory → STAC Collection + Items
- Automatic PMTiles overview detection
- Public URL mapping for S3→HTTPS conversion
- Projection info (CRS, geometry types) from GeoParquet metadata

Add  command for STAC validation:
- STAC spec compliance via pystac.validation
- Required fields checking
- Asset href resolution (local files)
- Best practices validation

_Currently open against #56 utils it's merged into `main`, then will switch base branch to `main` here._

## Related Issue(s)
- #34 

## Checklist
- [x] Code is formatted
- [x] Tests pass

### DIFF
--- a/tests/test_stac_check.py
+++ b/tests/test_stac_check.py
@@ -4,6 +4,7 @@ import json
 import os
 
 import click
+import pystac
 import pytest
 
 from geoparquet_io.core.stac_check import check_stac, validate_stac_file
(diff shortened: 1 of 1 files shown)


nlebovits:
@cholmes the suggestion I provide in this PR on how to convert geoparquet to pmtiles is not correct, since you can't actually do that, lol. I saw [your comment](https://gist.github.com/cholmes/d086d7d5ca4ac0148aa2f1259fd813d8) on using `gpq` to do the conversion as geoparquet --> geojson --> pmtiles. Ultimately, it'd be great if tippecanoe could somehow support pmtiles, since temporarily converting parquet to geojson just to generate pmtiles seems less than ideal, especially for large datasets. There's [a whole discussion on this here](https://github.com/felt/tippecanoe/issues/193), which I think you may have seen. But in the short term, how do you want to handle this here? 

I've also been considering whether it would make sense to add tippecanoe and gpq as optional deps here to allow users to directly generate pmtiles if they're missing, but I'm leaning away from it. The current implementation is to just look for a pmtiles file in the root of the STAC item/collection and warn if it's not there, along with a suggestion for how to make one. I could see a _separate_ library that bundles gpio, gpq, tippecanoe, and maybe keiko's in-line viewer thingy to generate quick thumbnails--maybe that's for the cloud-native geonode project?

PR #75, opened 2025-11-29
Add `gpio select` command for creating output with a subset of fields.

cholmes:
## Description
 Adds a new select command that allows users to create output files with a subset of fields from the input, inspired by GDAL's --fields option.

  Features:
  - Select specific fields with --fields "id,name,category"
  - Exclude fields with --fields "temp,debug" --exclude
  - Handle missing fields gracefully with --ignore-missing-fields
  - Support for quoted field names with spaces, commas, and escaped quotes
  - Geometry column automatically preserved to maintain valid GeoParquet format
  - Full support for remote files (S3, GCS, Azure) and compression options

  Example usage:
  #### Select specific fields
  gpio select input.parquet output.parquet --fields "id,name,geometry"

  #### Exclude unwanted fields
  gpio select input.parquet output.parquet --fields "temp_data,debug_info" --exclude

  #### Handle varying schemas across files
  gpio select input.parquet output.parquet --fields "id,optional_field" --ignore-missing-fields

## Technical Details

Decided to base off of GDAL select, even though it's not quite the ergonomics I'd pick from scratch, as it seems like it could be nice for users switching between the two.

## Related Issue(s)
- #

## Checklist
- [x] Code is formatted
- [ ] Tests pass (unsure, didn't work locally)

### DIFF
--- a/geoparquet_io/core/select.py
+++ b/geoparquet_io/core/select.py
@@ -2,6 +2,8 @@
 Core functionality for selecting/excluding fields from GeoParquet files.
 """
 
+from __future__ import annotations
+
 import click
 import fsspec
 import pyarrow.parquet as pq


cholmes:
I decided I don't like this approach. Going to try an `extract` command that does this plus bbox / geometry plus filter (where clause).

PR #289, opened 2026-03-13
Support converting and exporting files without geometry columns

cholmes:
## Summary

This PR adds comprehensive support for working with non-geospatial data in gpio, making it a more versatile tool for any Parquet workflow.

### Key Features

1. **Convert non-geo files to optimized Parquet** - CSV or Parquet files without geometry columns now convert to optimized Parquet (ZSTD compression, proper row groups) instead of failing
2. **Export non-geo Parquet to other formats** - Plain Parquet files can now be exported to CSV, JSON, Shapefile, and GeoPackage
3. **Friendly error messages** - Clear guidance when users try operations that don't make sense (e.g., inspect on CSV, FlatGeobuf without geometry)
4. **Consistent overwrite protection** - All output formats now require `--overwrite` flag

### Converting to Parquet (input without geometry)
```bash
$ gpio convert data.csv output.parquet
No geometry column detected. Converting as plain Parquet without GeoParquet metadata.
✓ Converted to optimized Parquet (no geometry)
```

### Exporting from Parquet (no geometry)
| Format | Behavior |
|--------|----------|
| **CSV** | Warns, outputs without WKT column |
| **JSON** | Warns "not valid GeoJSON", outputs plain JSON array |
| **Shapefile/GeoPackage** | Warns about no spatial features, proceeds |
| **FlatGeobuf** | Clear error: "FlatGeobuf requires geometry data" |

### Inspect Command Improvements
```bash
# Wrong file type
$ gpio inspect data.csv
Error: The 'inspect' command only works with Parquet files...

# Misnamed file (CSV with .parquet extension)  
$ gpio inspect fake.parquet
Error: The file 'fake.parquet' has a .parquet extension but is not a valid Parquet file...
```

### Consistent Overwrite Protection
All formats (CSV, JSON, FlatGeobuf, GeoPackage, Shapefile) now:
- Refuse to overwrite existing files by default
- Show helpful error: `Use --overwrite to replace it`

## Test Plan
- [x] 110+ lines of new tests for non-geo Parquet conversion
- [x] 4 tests for inspect error messages
- [x] 4 tests for no-geometry format exports
- [x] 3 tests for overwrite protection consistency
- [x] All existing tests pass

## Summary by CodeRabbit

* **New Features**
  * Per-command --overwrite support for conversion/export commands; writers (CSV, GeoJSON, GDAL formats) respect overwrite and warn on existing outputs.
  * Option to allow converting inputs without geometry into plain Parquet/CSV/JSON; GeoJSON accepts an optional description.

* **Bug Fixes**
  * Friendlier Parquet-related error messages and upfront inspect validation for misnamed or non-Parquet inputs.
  * Formats that require geometry now produce clear errors; non-geo formats emit informative warnings.

* **Tests**
  * Expanded tests for no-geometry conversions, overwrite protection, and inspect/extension validation.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -98,8 +98,8 @@ geoparquet_io/
 | `validate.py` | GeoParquet file validation against specification r... | 2854 |
 | `inspect_utils.py` | Utilities for inspecting GeoParquet files. | 1548 |
 | `duckdb_metadata.py` | DuckDB-based Parquet metadata extraction. | 1277 |
+| `convert.py` |  | 1259 |
 | `extract.py` | Extract columns and rows from GeoParquet files. | 1225 |
-| `convert.py` |  | 1189 |
 | `arcgis.py` | ArcGIS Feature Service to GeoParquet conversion. | 1161 |
 | `metadata_utils.py` | Utilities for extracting and formatting GeoParquet... | 1077 |
 | `partition_common.py` |  | 908 |


yharby:
One small thing I noticed while going through the codebase is that filesystem/path handling is currently mixed between `os.path` and `pathlib`.

From a quick scan I did:

```
Pattern                      Files   Occurrences
------------------------------------------------
from pathlib import Path     32      47
os.path.* usage              29      ~176
```

So both styles are already in use, but `os.path` still appears more frequently.

Since cross-platform path handling (Windows vs POSIX) can get messy, it might be worth deciding on a consistent direction for future code. Modern Python tends to favor `pathlib` because it’s OS-agnostic and gives a cleaner API (`Path / "file"` instead of `os.path.join`, etc.).

Would it make sense to:

* Prefer **`pathlib` for new code**, and
* Add a short note in **CLAUDE.md** like *“Prefer `pathlib` over `os.path` for new filesystem/path code”*?

Not suggesting a big refactor right now, but having a guideline could help keep things consistent going forward.

PR #404, opened 2026-04-24
fix: add wkb conversion to read partition

smohiudd:
## Description
In `read_partition` DuckDB converts GeoParquet WKB columns to native GEOMETRY type. Exporting directly via ``.arrow()`` serializes DuckDB's geometry format (not standard WKB), so downstream ``ST_GeomFromWKB`` calls fail with "WKB type 'UNKNOWN' is not supported". This PR checks if there is GEOMETRY type in the DuckDB schema and if yes casts GEOMETRY columns back to WKB via `_wrap_query_with_wkb_conversion` helper before returning the Table.

This bug was encountered when trying to read a partition and then apply Hilbert sort:
```
gpio.read_partition('s3://bucket/data/year=2026/month=02/day=*/*.parquet').sort_hilbert()
```
Resulting in this error:
```
Error: Could not parse WKB input:WKB type 'UNKNOWN' is not supported! (type id: 0, SRID: 0)
(You can use TRY_CAST instead to replace unsupported geometries with NULL)
```

## Breaking Changes
No

## Related Issue(s)
- #

## Checklist
- [x] Tests added/updated
- [ ] Documentation updated (README, guides, CLI reference)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of geometry columns when reading partitioned GeoParquet datasets so spatial columns are correctly converted on read.

* **Tests**
  * Added an end-to-end test ensuring partitioned reads followed by a Hilbert-curve sort preserve row counts, geometry column presence, and spatial validity.

### DIFF
--- a/geoparquet_io/api/table.py
+++ b/geoparquet_io/api/table.py
@@ -213,7 +213,10 @@ def read_partition(
         >>> table = gpio.read_partition('partitioned_output/')
         >>> table = gpio.read_partition('data/quadkey=*/*.parquet')
     """
-    from geoparquet_io.core.duckdb_utils import get_duckdb_connection
+    from geoparquet_io.core.duckdb_utils import (
+        _wrap_query_with_wkb_conversion,
+        get_duckdb_connection,
+    )
     from geoparquet_io.core.partition.reader import build_read_parquet_expr
     from geoparquet_io.core.remote import needs_httpfs
     from geoparquet_io.core.streaming import find_geometry_column_from_table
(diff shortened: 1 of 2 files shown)


nlebovits:
Hey @smohiudd thanks for the PR! Would you mind adding a test? Otherwise, lgtm.

smohiudd:
> Hey @smohiudd thanks for the PR! Would you mind adding a test? Otherwise, lgtm.

Done!

PR #429, opened 2026-05-06
fix(wfs): 10x faster extraction via Python HTTP download

### LINKED ISSUE #428: WFS extraction is catastrophically slow due to double-download and suboptimal parsing
## Summary

WFS extraction for large datasets is orders of magnitude slower than it should be. A 3.8M feature extraction that should take ~10-15 minutes is estimated to take **56+ hours**.

## Test Case

```bash
# Wallonia INSPIRE Buildings - 3.8M features
gpio wfs https://geoservices.wallonie.be/geoserver/inspire_bu/ows \
  --layer inspire_bu:BU.Building_building_emprise \
  -o buildings.parquet
```

### 1. Double Download in `_fetch_wfs_page_duckdb`

**File:** `geoparquet_io/core/wfs.py` lines 1087-1123

The function downloads the entire GeoJSON **twice**:

```python
# First download - just to check if empty
count_query = f"""
    SELECT len(features) AS cnt
    FROM read_json_auto('{safe_url}', maximum_object_size=536870912)
"""
count_result = con.execute(count_query).fetchone()

# Second download - to actually parse
query = f"""
    WITH features AS (
        SELECT unnest(features) AS feature
        FROM read_json_auto('{safe_url}', maximum_object_size=536870912)
    ...
"""
```

For 3.8M features (~9GB GeoJSON), this means downloading **18GB** instead of 9GB.

**Fix:** Use try/except instead of pre-checking, or cache the first result.

### 2. Suboptimal Parsing Approach

Current approach:
1. `read_json_auto` to parse raw JSON
2. `UNNEST` features array
3. `ST_GeomFromGeoJSON` on each geometry
4. `unnest(props)` to expand properties

Better approach using `ST_Read`:
```sql
SELECT * FROM ST_Read('https://wfs-url?outputFormat=application/json')
```

`ST_Read` uses GDAL which:
- Handles GeoJSON natively with streaming
- Auto-detects schema
- No manual geometry conversion needed
- Likely much more memory-efficient

## Observed Performance

- **WFS server throughput:** ~720 kbps (~90 KB/s)
- **Per-feature size:** ~2.4 KB (GeoJSON)
- **Total data:** ~9 GB for 3.8M features
- **Single download time:** ~28 hours at observed rate
- **With double-download bug:** ~56 hours

After 51 minutes:
- Only 14.4 MB received
- 1.6 GB memory usage
- Still on first (count) query

## Request

1. **Fix the double-download bug** — immediate priority
2. **Benchmark ST_Read vs read_json_auto** — may provide significant speedup
3. **Add profiling for this specific dataset** — use the Wallonia WFS as a benchmark case:
   - Track: download time, memory usage, parsing time
   - Compare: current approach vs ST_Read vs parallel pagination
4. **Consider progress reporting** — for multi-GB downloads, users need feedback

## Environment

- geoparquet-io: latest main
- DuckDB: with spatial + httpfs extensions
- Python: 3.11
- OS: Ubuntu 22.04

nlebovits:
## Summary

Fixes #428 — WFS extraction is catastrophically slow due to DuckDB's httpfs being inefficient for streaming large JSON over HTTP.

### Root Cause Analysis

The issue reported "double-download" as the culprit, but profiling revealed a deeper problem:

| Approach | 1000 features | Rate |
|----------|---------------|------|
| DuckDB httpfs streaming (old) | **31.4s** | 32 f/s |
| Python httpx → local parse (new) | **3.2s** | 310 f/s |

**DuckDB's httpfs is ~10x slower than Python's httpx** for streaming JSON. The parsing itself is instant (0.04s for 1000 features).

### The Fix

Replace DuckDB httpfs streaming with:
1. Download GeoJSON using Python httpx (with retry logic)
2. Write to temp file
3. Parse locally with DuckDB's spatial extension
4. Clean up temp file

This also eliminates the "double-download" bug as a side effect — no need for the separate count query.

### Benchmark Results

**Single-worker mode (default):**

| Features | Before | After | Speedup |
|----------|--------|-------|---------|
| 1,000 | 31.4s | 3.2s | **10x** |
| 5,000 | ~158s | 9.1s | **17x** |
| 50,000 | ~525s | 105.6s | **5x** |
| 100,000 | ~1050s | 140.7s | **7.5x** |

**With parallel workers (`--workers` flag):**

| Workers | 50k features | Rate | Speedup |
|---------|--------------|------|---------|
| 1 | 105.6s | 474 f/s | 1.0x |
| 4 | 36.4s | 1,374 f/s | 2.9x |
| 8 | 19.4s | 2,576 f/s | 5.4x |
| 10 | 18.4s | 2,711 f/s | **5.7x** |

### Impact on Original Issue

For 3.8M features (Wallonia INSPIRE Buildings):

| Configuration | Estimated Time |
|--------------|----------------|
| Before (buggy) | **56+ hours** |
| After (fix, 1 worker) | ~94 minutes |
| After (fix, 10 workers) | **~23 minutes** |

### Breakdown

At 100k features, the bottleneck is clearly network:
- Download: 157.4s (94%)
- Parse: 9.4s (6%)

The WFS server throughput is the limiting factor (~1.4 MB/s). Parallel pagination helps by making concurrent requests.

### Recommendations

For large datasets (100k+ features), use `--workers 4-8`:
```bash
gpio extract wfs https://example.com/wfs --layer my_layer -o output.parquet --workers 8
```

## Test plan

- [x] All 120 WFS tests pass
- [x] Tested with Wallonia INSPIRE Buildings WFS (real-world case from issue)
- [x] Verified empty response handling works
- [x] Benchmarked at 100k features

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **Improvements**
  * Enhanced WFS data fetching with more reliable download and parsing workflow.
  * Added parallel processing for paginated requests, improving performance.
  * Improved progress messages now include feature count awareness when using pagination.
  * Better error handling and coordinate reference system validation.
  * Refined logging for clearer visibility into the fetch process.

### DIFF
--- a/geoparquet_io/core/wfs.py
+++ b/geoparquet_io/core/wfs.py
@@ -5,7 +5,7 @@ This module provides functionality to download features from OGC WFS services
 and convert them to GeoParquet format. Supports WFS 1.0.0, 1.1.0, and 2.0.0.
 
 Key features:
-- DuckDB-native HTTP streaming for fast extraction (10x+ faster than Python HTTP)
+- Fast extraction via Python httpx download + DuckDB local parsing
 - Server-side bbox filtering
 - CRS negotiation with EPSG variant handling
 - Hilbert curve sorting and bbox column generation
@@ -1053,12 +1053,11 @@ def _infer_column_types(table: pa.Table) -> pa.Table:
 
 def _fetch_wfs_page_duckdb(url: str) -> pa.Table:
     """
-    Fetch a WFS GeoJSON page directly using DuckDB's httpfs.
+    Fetch a WFS GeoJSON page and parse with DuckDB.
 
-    This is MUCH faster than Python HTTP because:
-    - DuckDB streams the HTTP response (no Python memory buffering)
-    - JSON parsing happens in C++ (faster than Python json)
-    - Geometry conversion happens in-database (no temp files)
+    Downloads via Python httpx then parses locally. This is ~10x faster than
+    DuckDB's httpfs streaming for JSON because httpx handles HTTP more
+    efficiently for large responses.
 
     Args:
         url: Full WFS GetFeature URL with all parameters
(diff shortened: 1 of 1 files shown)


nlebovits:
Consolidated into #431 which unifies all WFS improvements into a single PR.

PR #452, opened 2026-06-03
fix(add): use bbox pre-filter for GeoParquet 2.0 spatial joins

cholmes:
## Summary

- GeoParquet 2.0 and parquet-geo-only files were skipping the bbox pre-filter in spatial joins (`add admin-divisions`, `add country-codes`), falling back to brute-force `ST_Intersects` on every feature pair — causing OOM errors on datasets with >100K features
- Now derives bounds from native geometry via `ST_XMin`/`ST_XMax`/`ST_YMin`/`ST_YMax` to build the same bbox pre-filter that GeoParquet 1.x files get from their explicit bbox column
- Extracts shared `build_bbox_condition()` helper into `core/add/spatial_join.py` to eliminate duplicated bbox condition logic between `admin_divisions.py` and `country_codes.py`

## Root Cause

When `get_bbox_advice()` detected native geometry (`skip_bbox_prefilter=True`), both modules set `input_bbox_col = None` and the spatial join query fell through to the no-optimization branch — a full cartesian-ish `LEFT JOIN ... ON ST_Intersects(...)`. The assumption was that DuckDB's native row group statistics would handle filtering, but those only help with row group pruning during reads, not with optimizing the join itself.

## Test plan

- [x] Existing tests pass (11 tests covering admin divisions and country codes)
- [x] Pre-commit hooks pass
- [x] Import linter passes
- [ ] Manual test with GeoParquet 2.0 file against `gpio add admin-divisions`

## Summary by CodeRabbit

* **Chores**
  * Improved spatial-join performance by smarter bounding-box pre-filtering that adapts to available input geometry characteristics.
  * Updated dry-run/export messaging to clearly indicate when native geometry bounds are being used for bbox pre-filtering.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -82,14 +82,13 @@ jobs:
         run: uv run bandit -r geoparquet_io/ -c pyproject.toml
 
       - name: Dependency audit
-        # CVE-2026-3219: pip tar/ZIP handling (CVSS 4.6). Fix in pip 26.1.
-        # Self-expiring ignore: removed automatically when pip 26.1+ available.
+        # PYSEC-2026-196: pip 26.1.1 vulnerability. Fix in pip 26.1.2.
         run: |
-          PIP_VERSION=$(uv run pip --version | awk '{print $2}' | cut -d. -f1,2)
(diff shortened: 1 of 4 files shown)


cholmes:
Did some additional testing and doing this actually hurts performance, instead of helping it, since duckdb can't rely on its native stuff as much.. Going to close this, and try another PR that goes in the other direction.

PR #453, opened 2026-06-04
fix(add): deduplicate border-straddling features in spatial joins

cholmes:
## Summary

- When input polygons straddle the border between two admin divisions, the spatial join was producing one output row per matching boundary — duplicating input features (e.g., 32,705 input rows → 57,938 output rows)
- Now deduplicates by default using `QUALIFY ROW_NUMBER() OVER (PARTITION BY row_id ORDER BY ST_Area(ST_Intersection(...)) DESC NULLS LAST)`, keeping only the match with the largest overlap area
- Adds `--all-matches` CLI flag to preserve the old behavior when users want all matching boundaries
- Applies the same fix to both `admin_divisions.py` and `country_codes.py`
- Upgrades DuckDB from 1.5.0 to 1.5.3 (fixes GEOMETRY + window function compatibility needed for the dedup query)

## Test plan

- [x] 9 new tests in `test_spatial_join_dedup.py` covering:
  - Query builder generates dedup SQL by default (QUALIFY, ROW_NUMBER, ST_Area)
  - Query builder omits dedup SQL with `deduplicate=False`
  - Dedup SQL correctly excludes internal `_gpio_row_id` column
  - Dedup works alongside bbox optimization
  - Synthetic DuckDB data: border-straddling polygon keeps largest overlap match
  - Unmatched features preserved via LEFT JOIN
  - CLI dry-run shows dedup SQL by default
  - CLI `--all-matches` flag omits dedup SQL
- [x] All 7 existing dry-run tests pass (no regressions)
- [x] Full test suite: 2670 passed, 0 regressions

## Summary by CodeRabbit

* **New Features**
  * Added an `--all-matches` option to keep every matching boundary instead of deduplicating to the best overlap. This is available in admin-boundary and country-code enrichment flows.
  * Improved spatial matching behavior for inputs with native geometries, including clearer dry-run output about bbox filtering.

* **Tests**
  * Added broader test coverage for deduplication, all-matches mode, join behavior, and dry-run SQL output.


cholmes:
I don't like the solution here, as it's memory/compute intensive, and I think most of the duplication was due to reading both the countries and states/provinces and matching against most of them. So want to fix some of that root problem, and then maybe revisit the border issue.

Issue #401, opened 2026-04-24
[Bug]: conversion from GeoPackage to GeoParquet has non-reproducible error


AndresKasekamp:
### Bug Description

When I run 
`gpio convert gpkg_name.gpkg parquet_name.parquet --layer "layer_name"`
it crashes sometimes. I tried converting all the layers in a GeoPackage to GeoParquet, and I am getting various errors.
However, it is non-reproducible, meaning that sometimes the conversion for a layer works fine, other times it crashes unexpectedly.

### Steps to Reproduce

1. Download GeoPackage Estonia Topographic Database from Estonia's Geoportal [GeoPackage in a zip file](https://geoportaal.maaamet.ee/index.php?lang_id=1&plugin_act=otsing&andmetyyp=ETAK&dl=1&f=ETAK_EESTI_GPKG.zip&page_id=609)
2. Unzip the file
3. Use `ogrinfo `to get the layer names
4. Use the layer names in a loop to convert all the GeoPackage layers to GeoParquet

Error pics:

<img width="862" height="320" alt="Image" src="https://github.com/user-attachments/assets/4d6fdcdf-021d-4cac-a0af-742a67ba410a" />
<img width="1462" height="283" alt="Image" src="https://github.com/user-attachments/assets/c1ab51ad-14e1-47c0-8e17-e4a4becce402" />
<img width="1370" height="319" alt="Image" src="https://github.com/user-attachments/assets/bc31f99e-db4a-4773-b603-cb47c018b312" />
<img width="1052" height="334" alt="Image" src="https://github.com/user-attachments/assets/cc6d4480-eb0c-45ee-81fb-98785dd0fb2c" />
<img width="1386" height="340" alt="Image" src="https://github.com/user-attachments/assets/85f22d0e-0a35-4d15-8893-c80690b841dc" />

### Expected Behavior

Successful conversion of GeoPackage layers to GeoParquet format.

### Actual Behavior

```shell
I got various errors, which sometimes occurred, other times not, for a layer in a GeoPackage:

- `Invalid Input Error: Unsupported geometry type in WKB`
- `Segmentation fault (core dumped)` 
- `munmap_chunk(): invalid pointer`
```

### Environment

```markdown
- OS: WSL2 on Windows 11
- gpio version: geoparquet-io, version 1.1.0b1
- Python version: Python 3.14.3
- Installation method: uv
```

### Sample Data (Optional)

GeoPackage from Estonia's geoportal:
[...]

PR #472, opened 2026-06-07
fix(ci): repair recurring slow-tests failures on main

cholmes:
## Why merges to `main` keep failing despite green PRs

The `slow-tests` job runs only on **push / schedule** — never on `pull_request`. So every `slow`/`network` test is skipped on PR checks and only executes *after* merge. A PR can be fully green and still break `main`. This PR fixes the recurring failures and adds a way to verify slow-tests *before* merge.

### 1. `test_pip_audit_passes` drift (most frequent)
`tests/test_security.py::test_pip_audit_passes` duplicated the dedicated `security` job's audit, but with its own CVE-ignore list that drifted out of sync. The workflow ignores `PYSEC-2026-196` (pip 26.1.1, fix 26.1.2); the pytest copy still only ignored a stale CVE and ran bare `pip-audit`, returning 1 → failure on every push.

**Fix:** removed the duplicate runtime audit from pytest; `slow-tests` always skips `test_security.py`. The `security` job is the single source of truth (it already handles CVE ignores *and* blocks PRs).

### 2. Estonia GeoPackage segfault (ubuntu)
`test_sequential_layer_conversion_no_crash` (regression for #401) segfaulted the pytest-xdist worker on Linux (`worker gw0 crashed`) and hit per-layer WKB read errors locally.

**Fix:** run the sequential conversion in a **fresh subprocess** — a native crash becomes a clean, diagnosable failure (non-zero return code) instead of killing the worker, and a fresh process avoids GDAL/DuckDB handle races with sibling tests sharing the worker. Per-layer read errors are tolerated (the regression guarded against is a hard crash, not individual bad layers).

### 3. Flaky DuckDB extension download
The *Pre-install DuckDB extensions* step failed intermittently on transient network errors from `extensions.duckdb.org`.

**Fix:** retry with backoff before failing.

## Also
`slow-tests` can now be opted into per-PR via the **`run-slow-tests`** label, so risky changes can be verified before merge. This PR carries that label so its own fixes run through the slow suite.

## Not included (deferred)
`test_benchmark_suite.py::test_chain_filter_reproject_partition_runs` also fails on `main` — but it's a **CRS null-vs-default** bug (reprojecting `EPSG:3857 → EPSG:4326` leaves the stale 3857 in geo metadata, in `_assemble_and_apply_geo_metadata`). That's being handled on the dedicated `fix/crs-null-vs-default` branch, so it's intentionally left out here to avoid conflicts.

## Summary by CodeRabbit

* **Tests**
  * Improved test resilience with retry/backoff for external extension setup.
  * Refactored layer-conversion test to run in an isolated worker subprocess and require at least one successful layer conversion.
  * Removed live dependency-audit as a unit test; security audit runs in a dedicated CI job.

* **Chores**
  * CI workflow: allow running slow tests via label and broadened conditions for label-triggered runs.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -121,8 +121,22 @@ jobs:
         run: uv sync --all-extras
 
       - name: Pre-install DuckDB extensions
+        shell: bash
         run: |
-          uv run python -c "import duckdb; con = duckdb.connect(); con.execute('INSTALL spatial'); con.execute('INSTALL httpfs'); con.execute('INSTALL h3 FROM community')"
+          # DuckDB extension downloads (extensions.duckdb.org) occasionally fail
+          # with a transient network error, which broke otherwise-green builds.
+          # Retry with backoff before giving up.
+          install="import duckdb; con = duckdb.connect(); con.execute('INSTALL spatial'); con.execute('INSTALL httpfs'); con.execute('INSTALL h3 FROM community')"
+          for attempt in 1 2 3 4 5; do
+            if uv run python -c "$install"; then
+              echo "DuckDB extensions installed (attempt $attempt)."
+              exit 0
+            fi
+            echo "Extension install failed (attempt $attempt). Retrying in $((attempt * 10))s..." >&2
+            sleep $((attempt * 10))
+          done
+          echo "Failed to install DuckDB extensions after 5 attempts." >&2
+          exit 1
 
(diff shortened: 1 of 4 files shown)


cholmes:
Merging in, as it improves tests, though I'm doing a bit more to try to make things a bit more resilient in the slow tests.

PR #480, opened 2026-06-10
fix(partition): single-pass COPY PARTITION_BY instead of per-value re-scan (#478)

### LINKED ISSUE #478: partition commands re-scan full input per partition value — O(N×data); use single-pass COPY PARTITION_BY
## Summary

`gpio partition string` / `gpio partition admin` re-scan the **entire input once per partition value**. The cost is `O(num_partitions × input_size)`, so partitioning a large dataset into many partitions becomes infeasible. Partitioning should be a **single streaming pass** over the input (DuckDB `COPY … (FORMAT parquet, PARTITION_BY (col))`).

## Impact (real-world case)

Partitioning the Fields of the World global field-boundary dataset by admin country:

- Input: **~3.2 billion rows / ~220 GB** GeoParquet (1000 files), **195** distinct countries (5,204 country×region combos).
- `gpio partition string --column admin:country_code` over this input performs **195 full scans of 220 GB ≈ 43 TB of reads** (country×region would be 5,204 scans).
- Observed: it produced ~4 GB of output in ~100 s and was on track for **hours-to-days**; never completed. (Made dramatically worse on a network filesystem, but the asymptotic blow-up is the real problem and bites on local disk too.)

A single-pass `PARTITION_BY` reads the input **once** and routes rows to all partition writers concurrently — minutes instead of days.

## Root cause

`geoparquet_io/core/partition/common.py`:

- `partition_by_column()` gets the distinct values, then **loops** over them (around line 877):
  ```python
  partition_values = _get_unique_partition_values(con, input_url, column_expr, column_name, verbose)
  for row in partition_values:
      _process_partition_value(con, input_url, row[0], ...)
  ```
- `_process_partition_value()` (line 713) builds a per-value query that **re-reads the whole input** and filters:
  ```python
  partition_query = f"""
      SELECT {select_clause}
      FROM '{input_url}'        -- full glob, re-scanned every call
      WHERE "{column_name}" = '{partition_value}'
  """
  write_parquet_with_metadata(con, partition_query, output_filename, original_metadata=partition_metadata, ...)
  ```
  The input is not clustered/sorted by the partition column, so DuckDB cannot prune row groups — each iteration scans all input.
- `_get_unique_partition_values()` (line 614) adds one more full `SELECT DISTINCT … FROM '{input_url}'` scan.

Net: **(1 + N) full scans** of the input for N partition values.

## Affected commands

- `gpio partition string` (confirmed)
- `gpio partition admin` (confirmed — its preview reported a join over 1.08M features and partition writes; same `partition_by_column` path after the admin join)
- Please check whether the cell-id partitioners (`partition a5/h3/s2/quadkey/kdtree`) share `partition_by_column`; if so they have the same issue.

## Reproduction

```bash
# Any GeoParquet with a high-cardinality string column works; bigger = more obvious.
gpio partition string big_dataset/*.parquet out/ --column admin:country_code -v
# Watch: it logs "Found N unique partition values" then writes them one by one,
# each re-reading the full input. Wall-clock scales with N × input_size.
```

## Proposed solution

Replace the per-value loop with a **single** DuckDB `COPY … PARTITION_BY` pass.

```sql
COPY (
    SELECT {select_clause}        -- excludes/keeps partition col per keep_partition_column
    FROM read_parquet('{input_url}')
) TO '{output_dir}'
  (FORMAT parquet, PARTITION_BY ({column_expr}),
   COMPRESSION '{compression}', COMPRESSION_LEVEL {level},
   ROW_GROUP_SIZE {rows}, OVERWRITE_OR_IGNORE {overwrite});
```

Design points to handle:

1. **Metadata preservation (the main subtlety).** `COPY` writes plain Parquet and drops the GeoParquet `geo` key and any extra KV metadata (e.g. fiboa/vecorel `collection`). Today this is preserved per-file by `write_parquet_with_metadata` (which also recomputes a per-partition `bbox` via `_strip_bbox_from_metadata`). Options, in order of preference:
   - **`KV_METADATA`**: if the installed DuckDB supports `KV_METADATA` together with `PARTITION_BY`, pass the serialized `geo` (+ passthrough keys like `collection`) so every partition file is written correct in one pass. Needs verification per DuckDB version.
   - **Post-pass stamp**: after the single COPY, walk the output files and apply `geo`/passthrough metadata (constant) — still 1 extra read+write per file, but no longer N× over the whole input. Reuse the existing metadata builder.
   - Decide whether to keep the nice per-partition `bbox` recompute. If `KV_METADATA` is used, the `bbox` would be global/constant unless recomputed in the post-pass; per-partition `covering` (the bbox struct columns) is still correct since those are real columns.

2. **Output layout / naming.** DuckDB `PARTITION_BY` writes Hive dirs `col=value/data_*.parquet`. Preserve gpio's current outputs:
   - `--hive`: map DuckDB's `col=value/` dirs to gpio's expected `{column}={value}/…` (and `{column}_prefix={value}` for `--chars`).
   - flat (default): flatten to `{value}.parquet` / `{prefix}_{value}.parquet` (`--prefix`). A post-COPY move/rename of the single `data_0.parquet` in each dir is cheap (rename, same FS).
   - Apply `sanitize_filename()` to values as today.

3. **Partition column inclusion.** DuckDB excludes `PARTITION_BY` columns from the data files by default; gpio has `keep_partition_column`. Re-select the column into the projection (aliased) if `keep_partition_column` is set, or add it back.

4. **`--chars` prefix partitioning.** Partition on the expression `LEFT("col", n)` (already built as `column_expr`) — `PARTITION_BY (column_expr)` works on expressions, so this maps directly.

5. **Many partitions / memory.** A single COPY with thousands of partitions keeps many writers open. DuckDB handles this with per-partition buffering/flushing, but validate memory on high-cardinality cases (our case: 5,204) and expose/respect `memory_limit` and `partition_buffer_size` if needed.

6. **Preview / analysis path.** Keep `--preview` and the strategy analysis (they only need the cheap `DISTINCT`/`GROUP BY` aggregate, not the per-value writes).

7. **Backward compatibility.** Preserve all current flags/behavior: `--hive`, `--prefix`, `--chars`, `--compression(-level)`, `--row-group-size(-mb)`, `--geoparquet-version`, `--overwrite`, `keep_partition_column`, and return value (`partition_count`).

## Alternative considered (rejected)

Pre-sort the input by the partition column (`gpio sort`) so the existing per-value `WHERE` benefits from Parquet row-group predicate pushdown (≈1 scan total across all values). Keeps the loop, but adds a full extra read+write to sort, still issues N `COPY`s, and depends on pushdown actually pruning. A single `PARTITION_BY` pass is simpler and strictly less I/O.

## Acceptance criteria

- Single pass over the input regardless of partition count (verify with `--show-sql`: one `COPY … PARTITION_BY`, not N queries).
- Correctness: row totals reconcile exactly (sum of partitions == input rows); no duplication/loss.
- Metadata: every partition retains a valid `geo` (crs omitted where appropriate, correct `encoding`/`geometry_types`/`covering`) **and** passthrough KV keys like fiboa/vecorel `collection`; `vec validate` passes on partitions.
- Output naming identical to current behavior for both flat and `--hive`, incl. `--prefix` and `--chars`.
- Scales: partition billions of rows / hundreds–thousands of partitions in a single read pass.
- Existing partition tests still pass; add a test asserting only one scan/COPY is issued and that totals reconcile on a multi-value fixture.

## Context

Discovered while producing an admin-partitioned, fiboa/vecorel-valid version of the FTW global dataset (`gpio add admin-divisions --vecorel` → partition by `admin:country_code`, splitting only countries >2 GB by `admin:subdivision_code`). The `add admin-divisions` step (post the maritime-overlap land-filter fix) is correct and fast; partitioning is the remaining blocker purely due to the per-value re-scan.

cholmes:
## Summary

Fixes #478. Partition commands re-scanned the **entire input once per partition value** — an `O(num_partitions × input_size)` anti-pattern that made partitioning large datasets into many partitions infeasible (the FTW case: ~195 country partitions of a 220 GB input ≈ **43 TB** of reads, never completed).

Both affected code paths now read the input **exactly once** via a single streaming DuckDB `COPY … PARTITION_BY` into a staging dir, then rewrite each (small) staging partition into its final file with correct per-partition metadata.

## Changes

- **`partition_by_column`** (`core/partition/common.py`) — used by `partition string` **and** the cell-id partitioners (`a5`/`h3`/`s2`/`quadkey`/`kdtree`): replaced the `_get_unique_partition_values` + `_process_partition_value` per-value loop with one COPY + per-file finalize. New shared helpers: `_build_staging_select`, `_run_partitioned_copy`, `_iter_staging_partitions`, `_finalize_partition_file`.
- **`partition_by_admin_hierarchical`** (`core/partition/admin_hierarchical.py`): replaced the per-combination loop over the enriched temp table with one `COPY … PARTITION_BY` (multi-level → nested dirs), reusing the shared helpers. Keeps vecorel column-retention and the `ensure_vecorel_columns` nullability rewrite (DuckDB can't write non-nullable Parquet columns).

## Behavior preserved

Output layout, naming, and all flags (`--hive`, `--prefix`, `--chars`, `--compression(-level)`, `--row-group-size(-mb)`, `--geoparquet-version`, `--overwrite`, `keep_partition_column`) are unchanged. Tight per-partition `bbox`/`geometry_types`/`covering` and passthrough KV (vecorel `collection`) are preserved because each partition is rewritten through the existing `write_parquet_with_metadata` path. The aliased partition key (`__gpio_part`) lets DuckDB drop the split column while keeping/dropping the original independently.

## Single-scan, verified

`-v` shows exactly one `COPY … PARTITION_BY`:

```
Single-pass partition COPY (one scan of input):
COPY (SELECT *, LEFT("fsq_place_id", 1) AS __gpio_part FROM '…' WHERE "fsq_place_id" IS NOT NULL)
  TO '…/.gpio_partition_staging' (FORMAT PARQUET, PARTITION_BY (__gpio_part),
  COMPRESSION SNAPPY, GEOPARQUET_VERSION 'NONE', OVERWRITE_OR_IGNORE)
```

## Tests

New `tests/test_partition_single_pass.py`:
- **Reconciliation**: `sum(partition rows) == input rows` for flat / `--hive` / `--chars` / keep+drop column.
- **Single scan**: spies the DuckDB connection — asserts exactly **one** `PARTITION_BY` statement and **zero** per-value `WHERE "col" =` writes.
- **Metadata**: each partition has valid `geo`, **tight** (per-partition) bbox, and passthrough `collection` KV.

All existing partition tests pass (`test_partition`, `test_partition_by_string`, `test_hive_partition`, `test_partition_admin_vecorel` incl. the local nullability integration test, cell-id suites). Full fast suite: 2795 passed. ruff / xenon / vulture / import-linter clean.

## Acceptance criteria (#478)

- [x] Single pass over input regardless of partition count
- [x] Row totals reconcile exactly (no dup/loss)
- [x] Per-partition `geo` valid; passthrough KV (`collection`) retained; vecorel non-nullable columns preserved
- [x] Output naming identical for flat and `--hive` (incl. `--prefix`, `--chars`)
- [x] Existing partition tests pass; added single-scan + reconciliation tests

## Summary by CodeRabbit

* **New Features**
  * Partitioning uses a single-pass staging + finalize flow for faster, streaming partition writes

* **Bug Fixes**
  * Errors on filename collisions from sanitized partition values
  * Values that sanitize to empty now use `_empty`
  * Warn when admin hierarchy is incomplete
  * Row order within partitions is no longer guaranteed
  * `--no-overwrite` reports only partitions written in the current run

* **Documentation**
  * Changelog updated with execution model and behavioral guarantees

* **Tests**
  * Added comprehensive tests covering single-pass partitioning, metadata, collisions, nulls, and overwrite behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,17 +80,28 @@ This is the first beta release of geoparquet-io 1.0, featuring major new spatial
 
 ### Fixed
 
-- Partition commands now read the input **once** instead of re-scanning it per
-  partition value (#478). `gpio partition string`, `gpio partition admin`, and
-  the cell-id partitioners (`a5`/`h3`/`s2`/`quadkey`/`kdtree`) all shared an
+- Partition commands now route all rows in a **single** `COPY … PARTITION_BY`
+  scan instead of re-scanning the input per partition value (#478).
+  `gpio partition string`, `gpio partition admin`, and the cell-id partitioners
+  (`a5`/`h3`/`s2`/`quadkey`/`kdtree`) all shared an
   `O(num_partitions × input_size)` per-value loop that made partitioning large
   datasets into many partitions infeasible (e.g. ~195 country partitions of a
-  220 GB input meant ~43 TB of reads). They now use a single streaming DuckDB
-  `COPY … PARTITION_BY` pass into a staging dir, then rewrite each (small)
-  partition into its final file with correct per-partition metadata. Output
-  layout, naming, flags, and per-partition `geo`/`bbox`/`covering` metadata
-  (plus passthrough KV like vecorel `collection` and non-nullable vecorel
-  columns) are unchanged.
+  220 GB input meant ~43 TB of reads). They now stream into a staging dir, then
+  rewrite each (small) partition into its final file with correct per-partition
+  metadata. Output layout, naming, flags, and per-partition
+  `geo`/`bbox`/`covering` metadata (plus passthrough KV like vecorel
+  `collection` and non-nullable vecorel columns) are unchanged.
+  - Distinct partition values that sanitize to the **same** filename (e.g.
+    `"São Paulo"` / `"São-Paulo"`) now raise a clear error instead of silently
+    dropping or overwriting rows; values that sanitize to empty fall back to
+    `_empty` rather than `.parquet`.
+  - `gpio partition admin` warns when features match a coarser admin level but
+    are missing a finer one (they cannot be placed in any partition).
+  - Note: row order *within* a partition is no longer guaranteed (sort each
+    output afterward if needed); with `--no-overwrite` the returned/printed
+    partition count reflects only files actually written this run; and when
+    partition analysis runs (the default) the input is read once more for that
+    cheap aggregate before the COPY.
 - Distinguish an explicit `crs: null` (CRS *unknown*) from an omitted `crs` key
   (defaults to OGC:CRS84), per the GeoParquet spec:
   - Reading a file with `crs: null` now logs a warning (once per input) from the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -80,6 +80,17 @@ This is the first beta release of geoparquet-io 1.0, featuring major new spatial
 
 ### Fixed
 
+- Partition commands now read the input **once** instead of re-scanning it per
+  partition value (#478). `gpio partition string`, `gpio partition admin`, and
+  the cell-id partitioners (`a5`/`h3`/`s2`/`quadkey`/`kdtree`) all shared an
+  `O(num_partitions × input_size)` per-value loop that made partitioning large
+  datasets into many partitions infeasible (e.g. ~195 country partitions of a
+  220 GB input meant ~43 TB of reads). They now use a single streaming DuckDB
+  `COPY … PARTITION_BY` pass into a staging dir, then rewrite each (small)
+  partition into its final file with correct per-partition metadata. Output
+  layout, naming, flags, and per-partition `geo`/`bbox`/`covering` metadata
+  (plus passthrough KV like vecorel `collection` and non-nullable vecorel
+  columns) are unchanged.
 - Distinguish an explicit `crs: null` (CRS *unknown*) from an omitted `crs` key
   (defaults to OGC:CRS84), per the GeoParquet spec:
   - Reading a file with `crs: null` now logs a warning (once per input) from the
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -80,17 +80,28 @@ This is the first beta release of geoparquet-io 1.0, featuring major new spatial
 
 ### Fixed
 
-- Partition commands now read the input **once** instead of re-scanning it per
-  partition value (#478). `gpio partition string`, `gpio partition admin`, and
-  the cell-id partitioners (`a5`/`h3`/`s2`/`quadkey`/`kdtree`) all shared an
+- Partition commands now route all rows in a **single** `COPY … PARTITION_BY`
+  scan instead of re-scanning the input per partition value (#478).
+  `gpio partition string`, `gpio partition admin`, and the cell-id partitioners
+  (`a5`/`h3`/`s2`/`quadkey`/`kdtree`) all shared an
   `O(num_partitions × input_size)` per-value loop that made partitioning large
   datasets into many partitions infeasible (e.g. ~195 country partitions of a
-  220 GB input meant ~43 TB of reads). They now use a single streaming DuckDB
-  `COPY … PARTITION_BY` pass into a staging dir, then rewrite each (small)
-  partition into its final file with correct per-partition metadata. Output
-  layout, naming, flags, and per-partition `geo`/`bbox`/`covering` metadata
-  (plus passthrough KV like vecorel `collection` and non-nullable vecorel
-  columns) are unchanged.
+  220 GB input meant ~43 TB of reads). They now stream into a staging dir, then
+  rewrite each (small) partition into its final file with correct per-partition
+  metadata. Output layout, naming, flags, and per-partition
+  `geo`/`bbox`/`covering` metadata (plus passthrough KV like vecorel
+  `collection` and non-nullable vecorel columns) are unchanged.
+  - Distinct partition values that sanitize to the **same** filename (e.g.
+    `"São Paulo"` / `"São-Paulo"`) now raise a clear error instead of silently
+    dropping or overwriting rows; values that sanitize to empty fall back to
+    `_empty` rather than `.parquet`.
+  - `gpio partition admin` warns when features match a coarser admin level but
+    are missing a finer one (they cannot be placed in any partition).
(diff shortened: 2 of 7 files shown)


cholmes:
Pushed `8e42860` addressing an adversarial code review of this PR. Most are hardening (collision detection, staging-dir isolation, quoted identifiers, etc. — see the commit message), but **one is a real correctness fix to what this PR previously produced**, worth calling out:

### 🐞 Alias column leaked into every output file

Before this commit, `read_parquet('…/__gpio_part=value/*.parquet')` in the finalize step **auto-detected hive partitioning and re-added the dropped `__gpio_part` alias as a data column**. So every partition output file silently carried a stray `__gpio_part` (and `__gpio_part0/1` for admin) column that shouldn't exist. No existing test caught it — the column-path tests only checked the *real* column was present/absent, and the vecorel admin test only checked the admin columns it deliberately keeps.

Fix: read staging back with `hive_partitioning=false`, plus regression assertions for both drivers that the alias never appears in output.

### Also a correctness fix
- Distinct partition values that sanitize to the same filename (e.g. `"São Paulo"` / `"São-Paulo"`, common in `partition string`/`admin` on place names) used to silently drop or overwrite rows. They now raise a clear `PartitionError`.

### Hardening (no behavior change for the happy path)
- Shared primitives extracted into `core/partition/staging.py` (public seam + `PartitionWriteOptions` instead of `**write_kwargs`).
- Staging dir is now a unique `mkdtemp` **sibling** of the output dir (can't leak into a remote upload, no concurrent-run clash) with incremental per-partition cleanup.
- Admin: extent computed once (not per level); dead `DISTINCT` scan removed; warn when a feature matches a coarse level but misses a finer one.
- `quote_identifier` on `column_name` / `input_geom_col` / admin columns.
- Docstrings/CHANGELOG: corrected the "reads once" claim, documented row-order non-determinism and `--no-overwrite` count semantics.

Two review findings were intentionally **not** applied: keeping `threads=1` (serves one-file-per-partition + bounded memory), and keeping the connection-level `SET preserve_insertion_order=false` (the enrichment join runs on that connection before the COPY, so it's load-bearing, not redundant).

Full fast suite (2845) + all partition tests pass.

PR #493, opened 2026-06-13
feat(pmtiles): expose tippecanoe tile-size/drop-densest flags (#492)

### LINKED ISSUE #492: Expose tippecanoe tile-size / drop-densest flags in `pmtiles create` (hardcoded combo defeats feature dropping)
## Summary

`gpio pmtiles create` hardcodes four tippecanoe flags with no way to
override them. Two of them interact in a way that silently defeats
size-based feature dropping, producing very large low-zoom tiles for
dense datasets. There is no CLI option, Python API parameter, env var,
or config hook to change this behavior.

## Where

`geoparquet_io/core/pmtiles.py`, `_build_tippecanoe_command`
(unconditional, ~lines 216–219):

```python
cmd.append("--simplify-only-low-zooms")
cmd.append("--no-simplification-of-shared-nodes")
cmd.append("--no-tile-size-limit")
cmd.append("--drop-densest-as-needed")
```

`create_pmtiles_from_geoparquet` (the public API) takes a fixed set of
named args — no `tippecanoe_args` / `**kwargs` / passthrough. The CLI
`pmtiles create` surfaces only `layer / min-zoom / max-zoom / bbox /
where / include-cols / precision / src-crs / attribution /
layer-by-column`. Confirmed against `main` (v1.3.0): no env var or
override path exists anywhere in `pmtiles.py`.

## The footgun

`--drop-densest-as-needed` only drops features to bring a tile back
**under the tile-size limit**. `--no-tile-size-limit` removes that
limit — so drop-densest has nothing to trigger on (only the 200k
features/tile hard cap remains). The two flags shipped together mean:
for dense data, low-zoom tiles keep essentially every feature and grow
without bound.

These are reasonable *lossless / max-fidelity* defaults, but they're
wrong for a national- or global-overview web map, and right now there's
no way to opt out without forking or bypassing the library.

## Real-world impact

A Moldova field-boundary dataset — 628,685 polygons × 14 numeric
attributes — tiles into a **230 MB** PMTiles archive. At national zoom
(z6) the overview tile carries a huge fraction of all 628k features, so
first paint and panning are slow. With a tile-size limit in place,
`--drop-densest-as-needed` would actually fire and bound those low-zoom
tiles.

## Proposed solution

Expose the hardcoded flags as overridable options while keeping the
current values as defaults (no behavior change for existing users):

- `--tile-size-limit / --no-tile-size-limit` (default: no limit, as today)
- `--drop-densest-as-needed / --no-drop-densest-as-needed` (default: on)
- optionally `--maximum-tile-bytes N` to set an explicit cap

Thread them: CLI option → `create_pmtiles_from_geoparquet(...)` kwarg →
`_build_tippecanoe_command(...)`. Add a test asserting the flags appear
/ are omitted in the built command for each toggle.

A lower-effort alternative would be a single escape-hatch
`--tippecanoe-arg` (repeatable) passthrough, but explicit named options
are friendlier and keep the validation surface controlled.

## Workaround today

Bypass the wrapper with the pipeline the library's own
`TippecanoeNotFoundError` message documents:

```
gpio extract geoparquet in.parquet --include-cols … \
  | gpio convert geojson - --precision 5 \
  | tippecanoe -P -o out.pmtiles -l fields -Z6 -z14 \
      --drop-densest-as-needed --extend-zooms-if-still-dropping
```

nlebovits:
## Summary

`gpio pmtiles create` hardcoded four tippecanoe production flags with no way to override them. Two of them interact in a way that silently defeats size-based feature dropping:

- `--no-tile-size-limit` removes the per-tile size cap
- `--drop-densest-as-needed` only drops features to bring a tile back **under** that cap

Shipped together, drop-densest has nothing to trigger on, so low-zoom tiles on dense (national/global) data grow essentially unbounded. There was no CLI option, Python API parameter, or env var to change this (issue #492).

## What changed

All four flags are now individually toggleable, plus a new explicit byte cap, threaded end-to-end through **CLI → Python API → core**:

| Flag | CLI option | API kwarg | Default |
|------|-----------|-----------|---------|
| `--simplify-only-low-zooms` | `--simplify-only-low-zooms / --no-...` | `simplify_only_low_zooms` | `True` |
| `--no-simplification-of-shared-nodes` | `--no-simplification-of-shared-nodes / --simplification-of-shared-nodes` | `no_simplification_of_shared_nodes` | `True` |
| `--no-tile-size-limit` | `--no-tile-size-limit / --tile-size-limit` | `no_tile_size_limit` | `True` |
| `--drop-densest-as-needed` | `--drop-densest-as-needed / --no-...` | `drop_densest_as_needed` | `True` |
| `--maximum-tile-bytes N` | `--maximum-tile-bytes` | `maximum_tile_bytes` | `None` |

**Defaults reproduce the historical command exactly** — pure opt-in, zero behavior change for existing users.

`--maximum-tile-bytes` takes precedence over `--no-tile-size-limit` (the two are contradictory; the explicit cap is what gives drop-densest a limit to drop against).

### Fixing the footgun

```bash
# Re-enable the size limit so drop-densest actually drops on dense data
gpio pmtiles create dense.parquet tiles.pmtiles --tile-size-limit --max-zoom 14

# Or set an explicit cap
gpio pmtiles create dense.parquet tiles.pmtiles --maximum-tile-bytes 500000
```

## Tests

- 5 build-command tests (each toggle off + `maximum_tile_bytes` precedence/suppression)
- 3 CLI-level tests asserting options parse and thread through to core
- Existing default test guards against regression
- 31 pmtiles + 40 geojson-stream tests pass; all commit-stage pre-commit hooks pass

## Docs

Added a "Tuning tile generation (dense data)" subsection to `docs/guide/geojson.md` with the footgun warning and CLI + Python tabbed examples.

Closes #492

## Summary by CodeRabbit

* **New Features**
  * Added new PMTiles generation tuning options to the CLI and API (simplification controls, densest-feature dropping toggle, tile-size limiting toggle).
  * Added an explicit per-tile byte cap option and a force overwrite flag.
* **Documentation**
  * Documented dense-data tuning behavior, including how tile-size limiting interacts with densest-feature dropping, with CLI and Python examples.
* **Tests**
  * Added unit and CLI tests validating option pass-through, default behavior, precedence, and `--force` handling.
* **Chores**
  * Updated ignore files to exclude local cache/tool directories (including `.serena/` and Menard cache artifacts).


nlebovits:
### Follow-up commit (`06c70f9`)

Two additions on top of the original flag work:

**1. Refactored `_run_pipeline` (xenon F → B).** It was cyclomatic-complexity **F (42)**, over the project's `xenon --max-absolute=E` gate (pre-existing on `main`, but worth clearing now that this PR touches the file). Decomposed into focused helpers — `_log_pipeline`, `_spawn_gpio_chain`, `_collect_upstream_errors`, `_pump_features`, `_run_with_layer_injection`, `_run_simple` — leaving `_run_pipeline` a thin orchestrator (now **B/8**). Behavior preserved exactly: helpers share the process list so cleanup still terminates every subprocess on failure, stderr is still drained before waiting, and upstream errors surface in the same order. `_spawn_gpio_chain` self-cleans on partial-spawn failure to keep the original leak-free guarantee.

**2. Added `--force` / `-f`** to `pmtiles create` (and the `create_pmtiles` API), passing tippecanoe's `--force` to overwrite an existing output file.

`xenon` now passes clean for `pmtiles.py`. 35 pmtiles tests pass (4 new: build-command force on/off + CLI threading for `--force`/`-f`).

> Note: repo-wide `xenon` still reports one unrelated **F** — `core/common.py:537 _promote_numeric_type` — which predates this branch and lives in a different module. Out of scope here; happy to clear it in a separate PR.

PR #507, opened 2026-06-16
feat: default geometry repair with ST_MakeValid (issue #506)

### LINKED ISSUE #506: Invalid geometry handling: investigate and default to ST_MakeValid across the pipeline
## Summary

geoparquet-io needs consistent, robust handling of invalid geometry across the entire pipeline. Currently invalid geometry causes silent data quality issues or crashes at various points.

## Request

1. **Audit** how invalid geometry is currently handled in:
   - WFS extraction (`gpio extract wfs`)
   - Format conversion (`gpio convert`)
   - PMTiles generation (`gpio pmtiles create`)
   - Any other relevant operations

2. **Default to ST_MakeValid** - repair geometry automatically unless user opts out with a `--no-repair-geometry` flag (or similar)

3. **Warn** when invalid geometry is encountered and repaired, with feature count

## Observed behavior

When extracting from IDECOR WFS (Córdoba, Argentina), some layers contain invalid polygons (self-intersections, unclosed rings). These pass through extraction silently, then crash PMTiles generation:

```
TopologyException: unable to assign free hole to a shell at -64.509654999999995 -31.433138
Error: tippecanoe failed with exit code 110
```

DuckDB spatial confirms 3 invalid geometries in the extracted parquet:
```sql
SELECT id, ST_IsValid(geometry) FROM 'layer.parquet' WHERE NOT ST_IsValid(geometry)
-- Returns 3 rows
```

## Why default to repair

- Most users want valid output, not crash-on-invalid
- ST_MakeValid is the standard solution (PostGIS, DuckDB spatial, GDAL all support it)
- Invalid geometry in source data is common in government/municipal datasets
- Users who need to preserve invalid geometry exactly can opt out

## Implementation notes

DuckDB spatial provides `ST_MakeValid()` which could be applied during the pipeline. The repair should happen early (at extraction or conversion) so downstream operations work reliably.

nlebovits:
## Summary

Invalid geometry (self-intersections, unclosed rings — common in government/municipal data) is now **repaired automatically** with `ST_MakeValid` across the entire pipeline. This fixes the reported crash:

```
TopologyException: unable to assign free hole to a shell
Error: tippecanoe failed with exit code 110
```

- Adds `--repair-geometry/--no-repair-geometry` flag (default **on**) to: `convert`, `convert geojson`, `extract wfs/arcgis/bigquery/carto/geoparquet`, `pmtiles create`
- Emits exact-count warnings: `Repaired 3 invalid geometries`
- Opt-out still warns: `Left unrepaired 3 invalid geometries (--no-repair-geometry)`
- Only invalid geometry is touched; valid geometry is byte-identical, NULL preserved, bboxes stay correct

## Test plan

- [ ] 18 unit tests + 4 integration tests added (2968 total passing)
- [ ] Pre-commit hooks pass (ruff, import-linter, doc-sync, no-click-echo, duckdb-antipatterns)
- [ ] End-to-end: invalid GeoJSON → PMTiles succeeds (was crashing with TopologyException)
- [ ] Opt-out (`--no-repair-geometry`) preserves invalid geometry and warns

Closes #506

🤖 Generated with [Claude Code](https://claude.ai/code)

## Release Notes

* **New Features**
  * Geometry repair is now enabled by default across conversion, extraction, GeoJSON generation, and PMTiles creation.
  * Added a CLI toggle `--repair-geometry/--no-repair-geometry` and a matching Python option `repair_geometry` (default `True`) to control behavior.
  * When disabled, invalid geometries are preserved and a warning reports how many were not repaired.

* **Documentation**
  * Updated the convert guide with details and opt-out behavior.

* **Tests**
  * Added unit and end-to-end integration coverage for repair behavior, warnings, opt-out, and regression scenarios.


nlebovits:
## Bug: DuckDB segfault during geometry repair

The repair feature detects invalid geometries correctly but crashes during Arrow conversion. This is a DuckDB spatial extension bug, not a logic error in our code.

### Reprex

```bash
# Extract layer with repair enabled (default) - CRASHES
gpio extract wfs \
  "https://idecor-ws.mapascordoba.gob.ar/geoserver/idecor/wfs" \
  "idecor:Carlos_Paz_Peligrosidad" \
  /tmp/test.parquet --verbose

# Output shows repair working, then segfault:
# Repaired 3 invalid geometries
# Fatal Python error: Segmentation fault
```

### Root cause

The crash occurs in `repair_arrow_table_geometry()` at line 167 when calling `.arrow().read_all()` after `ST_MakeValid`. The repair SQL executes successfully - the crash is in DuckDB's Arrow export of the repaired geometry.

```python
# This works (repair + fetch as Python objects):
con.execute("SELECT ST_AsWKB(ST_MakeValid(...))").fetchall()  # OK

# This crashes (repair + export to Arrow):
con.execute("SELECT ST_AsWKB(ST_MakeValid(...))").arrow().read_all()  # SEGFAULT
```

### Workaround needed

Either:
1. Report upstream to DuckDB and wait for fix
2. Find alternative repair path that doesn't hit the Arrow export bug
3. Wrap in try/except and fall back to unrepaired table with warning

DuckDB version: 1.5.1

nlebovits:
## ✅ Segfault resolved — and the earlier root-cause analysis was wrong

Followed up on the SIGSEGV reported above. The crash is real and reproducible, but the **previous diagnosis (a DuckDB Arrow-export bug) is falsified**. Fixed in 8b343ad.

### What the earlier comment claimed
> The crash is in DuckDB's Arrow export of the repaired geometry. `.fetchall()` works; `.arrow().read_all()` segfaults.

Tested that exact dichotomy on the real IDECOR data:

| Test | Result |
|------|--------|
| Full repair query → `.fetchall()` | **SIGSEGV** (not just `.arrow()`) |
| Full repair query → `.arrow().read_all()` | SIGSEGV |
| `ST_MakeValid` over all 64 rows (fetch *and* arrow) | ✅ fine |
| Synthetic bowtie through identical code path | ✅ fine |

So it is **not** Arrow export, and **not** `ST_MakeValid` itself. None of the three proposed workarounds (report upstream / Arrow-free path / fall back to unrepaired) were necessary.

### Actual root cause
A DuckDB 1.5.1 **vectorized conditional-execution bug**. The crash requires all three together:
1. The OR-form predicate with a **raw-blob `IS NULL`** term: `"geometry" IS NULL OR ST_IsValid(TRY(ST_GeomFromWKB("geometry")))`
2. `ST_MakeValid(...)` in the **opposite CASE branch**
3. The **exact valid/invalid row layout** — the raw-blob `IS NULL` term mis-aligns the selection vector fed to `ST_MakeValid`, reading out of bounds.

It is data-layout dependent: no subset of the 64 IDECOR rows reproduces it. Narrowing term-by-term, either dropping the redundant `col IS NULL` term or inverting to the AND-form eliminates the crash (deterministic, 3/3).

### Fix
Rewrote the repair predicate to the **logically-equivalent AND-form** (repair in `THEN`, passthrough in `ELSE`) in `repair_arrow_table_geometry` and the `repair_query_geometry` BLOB branch:

```sql
CASE WHEN parsed IS NOT NULL AND NOT ST_IsValid(parsed)
     THEN ST_AsWKB(ST_MakeValid(parsed)) ELSE col END
```

The native-GEOMETRY path (`repair_geometry_sql`) was verified safe and left unchanged — the trigger requires the raw-blob term.

### Verification
- `gpio extract wfs … idecor:Carlos_Paz_Peligrosidad` now exits 0, writes 64 features, **0 invalid geometries**.
- Regression guard: real 64-row WKB fixture (`tests/data/idecor_carlospaz_invalid_wkb.parquet`) exercised in a **subprocess** so a revert fails the test cleanly instead of crashing the pytest worker.
- All 18 geometry-repair tests pass.

PR #521, opened 2026-06-26
Add `gpio reduce-precision` command

jatorre:
## Description

Adds a `reduce-precision` command that snaps geometry coordinates to a fixed grid via DuckDB's `ST_ReducePrecision`. This shrinks the dominant geometry column substantially (~37% at ~0.11 m on EPSG:4326 data, more at coarser grids), lowering file size and the bytes a query must read. The native CRS is preserved — it never reprojects.

## Technical Details

Pipeline order is load-bearing and was verified empirically against DuckDB 1.5.1 / GEOS:

1. **Repair before reduce.** `ST_ReducePrecision` raises `TopologyException` on invalid input, so `ST_MakeValid` runs *first* (guarded, byte-identical for valid geometry). With `--no-repair-geometry` the reduce is wrapped in `TRY` so invalid input degrades to NULL instead of aborting the whole file.
2. **Reduce.** Snap to `--grid` (CRS units). On valid input GEOS preserves validity, so no second repair is needed.
3. **Drop empty.** Precision reduction routinely collapses slivers to `POLYGON EMPTY` / `LINESTRING EMPTY`; these (and pre-existing null/empty) are dropped by default with a count warning. `--keep-empty` retains them.

A stored bbox covering column is regenerated from the reduced geometry so it never goes stale (a stale bbox would mis-prune) — mirroring how `convert reproject` regenerates bbox. `--grid` is **required**: the operation is lossy and the right value is CRS-dependent (`1e-6` ≈ 0.11 m on EPSG:4326; `0.1`/`1.0` for metre-based projected CRS).

- Core: `core/reduce_precision.py` — `reduce_precision_table()` (Arrow in/out) + `reduce_precision()` (file/stdin via the shared `execute_transform`).
- CLI: top-level `gpio reduce-precision` (single-file, supports stdin/stdout, remote I/O).
- Python API: `Table.reduce_precision()` and `ops.reduce_precision()`.
- Reuses the existing `repair_geometry_sql` helper and the `--repair-geometry/--no-repair-geometry` decorator.

Complexity grade A; module coverage 97%; import-linter clean (core stays Click-free).

## Breaking Changes

## Related Issue(s)
- N/A

## Checklist
- [x] Tests added/updated

## Summary by CodeRabbit

* **New Features**
  * Added a new “reduce precision” command for snapping geometry coordinates to a fixed grid.
  * Added matching support in the Python API and fluent table interface.
  * Added a new user guide page and updated command references.
* **Bug Fixes**
  * Reduced geometries can now be repaired before processing, and empty results can be kept or dropped as needed.
  * Stored bounding-box metadata is regenerated after precision reduction.
* **Tests**
  * Added coverage for CLI, Python API, geometry validity, metadata handling, and empty-geometry behavior.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -64,6 +64,7 @@ geoparquet_io/
 | `gpio partition` | a5, admin, h3, kdtree, quadkey, s2, string | Commands for partitioning GeoParquet files |
 | `gpio pmtiles` | create | PMTiles generation commands |
 | `gpio publish` | stac, upload | Commands for publishing GeoParquet data (STAC metadata, cloud uploads) |
+| `gpio reduce-precision` |  | Reduce geometry coordinate precision by snapping to a fixed grid |
 | `gpio skills` |  | List and access LLM skills for gpio |
 | `gpio sort` | column, hilbert, quadkey | Commands for sorting GeoParquet files |
 <!-- END GENERATED: cli-commands -->
--- a/docs/api/python-api.md
+++ b/docs/api/python-api.md
@@ -623,6 +623,25 @@ WGS84, use `assume_crs84=True` to treat them as OGC:CRS84 and write the default
 table = gpio.read('unknown_crs.parquet').reproject(assume_crs84=True)
 ```
 
+#### `reduce_precision(grid, repair=True, drop_empty=True)`
+
+Reduce geometry coordinate precision by snapping to a fixed grid (DuckDB's
+`ST_ReducePrecision`). Shrinks the geometry column substantially; the native CRS
+is preserved (no reprojection). Invalid geometry is repaired *before* reducing,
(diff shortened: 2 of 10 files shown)


cholmes:
Love this, but I feel like it's not so appropriate for a top-level command - I like to keep those smaller in number and then group into sub-commands. Could we put it in a new sub-group? Like `tools` or `processing` or `process`

cholmes:
Note there's also a potential command to reduce decimal precision, like cut significant digits, that I think is also worth having. I actually thought that's what this was when I just read the title. It can be useful for decreasing the size of parquets, as most coordinates have way more significant digits than is meaningful. There's also reducing them for other columns too. No need to do those now, but would be good to at least have the name be a bit more meaningful, like reduce-geom-precision or something.

PR #530, opened 2026-06-28
fix(spatial): make add/partition CRS-aware for non-CRS84 input (#525)

### LINKED ISSUE #525: Spatial operations are not CRS-aware: admin joins error and grid keying silently wrong for non-CRS84 input
## Summary

Several spatial operations assume the input GeoParquet is in OGC:CRS84 / EPSG:4326 (lon-lat degrees) and do not reproject. On real-world data in another CRS this produces **two failure modes**:

- **Class A — spatial-predicate joins**: a hard `ST_Intersects` CRS error.
- **Class B — lon/lat grid keying**: silently *wrong* cells, no error.

Both are reproducible with this public file (Netherlands buildings, **EPSG:28992**, projected metres):
`https://data.source.coop/cholmes/portolan-nl/kadaster/panden/bag-light.parquet`

## Class A — spatial-predicate joins error on CRS mismatch

The centroid-in-polygon join against the OGC:CRS84 admin boundaries is done with no reprojection, so DuckDB 1.5 refuses it:

```
Binder Error: Cannot call function 'ST_Intersects' with geometries of different
coordinate reference systems (CRS). First geometry type is in 'OGC:CRS84' which
is not compatible with 'EPSG:4258'.
```

Affected (all share the same pattern):
- `gpio add admin-divisions` — `geoparquet_io/core/add/admin_divisions.py`
- `gpio partition admin` — `geoparquet_io/core/partition/admin_hierarchical.py`
- `gpio process aggregate admin` — `geoparquet_io/core/process/aggregate/by_admin.py`

`admin_hierarchical.py` only does `SET geometry_always_xy = true` and assumes "compatible CRS"; there is no `ST_Transform`. Any input not labelled OGC:CRS84 (e.g. EPSG:4258, the standard EU geographic CRS) errors.

Repro:
```
gpio process aggregate admin --level region \
  https://data.source.coop/cholmes/portolan-nl/kadaster/panden/bag-light.parquet out.parquet
```

## Class B — lon/lat grid keying is silently wrong for projected CRS

The grid keying calls `<grid>_lonlat_to_cell(ST_X(ST_Centroid(geom)), ST_Y(ST_Centroid(geom)), …)`, which expects **degrees**. For a projected CRS (e.g. EPSG:28992, coordinates in the ~0–280,000 metre range) the X/Y are fed in as if they were lon/lat, producing meaningless/out-of-range cells. No error is raised.

Affected:
- `gpio add a5` / `add h3` / `add s2` / `add quadkey` — `geoparquet_io/core/add/{a5,h3,s2,quadkey}.py`
- the corresponding `gpio partition {a5,h3,s2,quadkey}`
- `gpio process aggregate a5` — `geoparquet_io/core/process/aggregate/by_a5.py`

Repro:
```
gpio add a5 \
  https://data.source.coop/cholmes/portolan-nl/kadaster/panden/bag-light.parquet out.parquet
# completes without error but a5 cells are wrong (metres interpreted as degrees)
```

## Proposed fix

Introduce a shared "normalize input geometry to the operation's expected CRS" utility and apply it everywhere:

- **Admin joins**: reproject the input centroid to the admin boundaries' CRS (OGC:CRS84) before `ST_Intersects`.
- **Grid keying**: reproject the input geometry/centroid to EPSG:4326 / OGC:CRS84 before `*_lonlat_to_cell`.

Mechanics (DuckDB 1.5 confirmed working):
- Detect the input CRS from the `GEOMETRY('<crs>')` column type (or geo metadata).
- If the CRS is present and differs from the target, `ST_Transform(geom, '<src>', '<target>')` (session has `geometry_always_xy = true`).
- If the geometry is CRS-less (e.g. from an in-memory Arrow table via `ST_GeomFromWKB`), treat it as already the target CRS — `ST_Intersects` accepts a CRS-less geometry against a CRS-bearing one.
- If the CRS already matches the target, no transform.

One shared helper + tests, applied across `add`, `partition`, and `process aggregate`. Suggest a single PR.

## Notes

- Related: #524 (auto-resolution) is in the same grid/spatial area; both could be addressed when reworking these modules, but are independent.
- `process aggregate {a5,admin}` is new and currently carries the **same** CRS assumptions as the existing `partition`/`add` equivalents — i.e. no regression, just the same gap surfaced on real data.

cholmes:
## Summary

Fixes #525.

Several spatial operations assumed the input GeoParquet was OGC:CRS84 / EPSG:4326 (lon-lat degrees) and never reprojected. On real-world data in another CRS this produced two failure modes:

- **Class A — admin spatial-predicate joins**: a hard DuckDB error `Cannot call function 'ST_Intersects' with geometries of different coordinate reference systems`.
- **Class B — lon/lat grid keying**: silently *wrong* cells (projected metres fed to `*_lonlat_to_cell` as if they were degrees), no error.

## What changed

A shared "normalize geometry to the operation's expected CRS" helper in `core/crs_utils.py`:
- `crs_string_for_transform(crs)` → `"AUTH:CODE"` for a non-default identifiable CRS, else `None`.
- `transform_geom_sql(geom_expr, source_crs, target_crs="OGC:CRS84")` → wraps the expression in `ST_Transform` when needed, else returns it unchanged (no-op for CRS84 / CRS-less / default).
- `source_crs_string(parquet_file)` → detects a file's CRS as a transform string (reuses the existing `extract_crs_from_parquet`).

Applied across:
- **Grid keying** — `add a5/h3/s2/quadkey` reproject to lon/lat before centroid keying (`partition {a5,h3,s2,quadkey}` route through `add`, so they're covered too). `add quadkey` now **reprojects** a known projected CRS instead of rejecting it; its bbox-midpoint path is skipped when reprojecting (the stored bbox is in the source CRS).
- **Admin joins** — `add admin-divisions` and `partition admin` reproject the input to the admin CRS (OGC:CRS84) before `ST_Intersects`, compute the extent filter over the reprojected geometry, and skip the stored-bbox pre-filter when the CRS differs. `build_spatial_join_condition` gained an optional pre-built input-geom expression to support this.

## Hot path is unchanged (no regression)

CRS84 / CRS-less input makes the helper a no-op, so the common path is byte-for-byte what it was. Guarded structurally (the join SQL keeps its bbox pre-filter and contains no `ST_Transform` for CRS84 input) and measured:

| path | wall-clock (`add a5`, 400k rows) |
|---|---|
| baseline `main` (CRS84) | ~2.08–2.11s |
| this branch (CRS84) | ~2.13s |
| this branch (projected, reprojects per row) | ~2.02–2.10s |

No meaningful overhead on either path — the per-row `ST_Transform` is dwarfed by cell computation.

## Testing

- New `tests/test_crs_normalize.py` (21 tests, written failing first): helper unit tests; **Class B** (projected `fields_5070` fixture keyed via `add a5/h3/s2/quadkey` matches a reprojected-to-CRS84 copy); **Class A** (the join builders assign rows for projected input, vs 0 without the fix); and hot-path no-op guards.
- Full fast suite: 2999 passed, 8 skipped (one pre-existing flaky commitizen test under `xdist`, passes isolated and on baseline).
- All quality gates pass (ruff, codespell, xenon, import-linter, vulture).
- **End-to-end on the issue's real EPSG:28992 file** (`bag-light.parquet`, 11.4M rows): `add admin-divisions` now completes and assigns 11.34M/11.37M features to "Netherlands (Kingdom of the)" (border slivers to Belgium/Germany) — previously it errored at `ST_Intersects`. Grid keying produces correct lon/lat cells.

## Follow-up

The `process aggregate {a5,admin}` modules carry the same assumption but live on the `feature/process-aggregate` branch (not on `main`). They get the same helper wired in via a separate follow-up commit on that branch, per the agreed scope.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed spatial grid indexing and admin spatial joins to be CRS-aware, reprojecting non-default inputs before spatial processing.
  * Ensured quadkey generation and keying for grids (A5/H3/S2/quadkey) work correctly for projected CRS inputs.
  * Improved admin hierarchical enrichment/partitioning CRS handling and bbox pre-filter behavior.

* **Documentation**
  * Updated changelog and the `add` guide to clearly describe CRS handling rules for grid cell generation and administrative joins.

* **Tests**
  * Added extensive CRS detection/reprojection regression coverage for grids, joins, and query behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -89,6 +89,15 @@ This is the first beta release of geoparquet-io 1.0, featuring major new spatial
   the resolution closest to the target partition count. One fix covers
   `a5`/`h3`/`s2`/`quadkey`; it falls back to the old global estimate when the
   data can't be probed.
+- **Spatial operations are now CRS-aware (#525).** `gpio add`/`partition` for the
+  lon/lat grids (`a5`, `h3`, `s2`, `quadkey`) and the admin spatial joins
+  (`add admin-divisions`, `partition admin`) assumed OGC:CRS84 input. On data in
+  another CRS this either errored (admin joins: `ST_Intersects` CRS mismatch) or
+  silently produced wrong cells (grids: projected metres keyed as degrees). A
+  shared helper now detects the input CRS and reprojects to the operation's
+  expected CRS (lon/lat for grids, OGC:CRS84 for admin) before the spatial work;
+  CRS84/CRS-less input is untouched, so the common path is unchanged. `add quadkey`
+  no longer rejects a known projected CRS — it reprojects instead.
 - Partition commands now route all rows in a **single** `COPY … PARTITION_BY`
   scan instead of re-scanning the input per partition value (#478).
   `gpio partition string`, `gpio partition admin`, and the cell-id partitioners
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -89,6 +89,15 @@ This is the first beta release of geoparquet-io 1.0, featuring major new spatial
   the resolution closest to the target partition count. One fix covers
   `a5`/`h3`/`s2`/`quadkey`; it falls back to the old global estimate when the
   data can't be probed.
+- **Spatial operations are now CRS-aware (#525).** `gpio add`/`partition` for the
+  lon/lat grids (`a5`, `h3`, `s2`, `quadkey`) and the admin spatial joins
+  (`add admin-divisions`, `partition admin`) assumed OGC:CRS84 input. On data in
+  another CRS this either errored (admin joins: `ST_Intersects` CRS mismatch) or
+  silently produced wrong cells (grids: projected metres keyed as degrees). A
+  shared helper now detects the input CRS and reprojects to the operation's
+  expected CRS (lon/lat for grids, OGC:CRS84 for admin) before the spatial work;
+  CRS84/CRS-less input is untouched, so the common path is unchanged. `add quadkey`
+  no longer rejects a known projected CRS — it reprojects instead.
 - Partition commands now route all rows in a **single** `COPY … PARTITION_BY`
   scan instead of re-scanning the input per partition value (#478).
   `gpio partition string`, `gpio partition admin`, and the cell-id partitioners
--- a/docs/guide/add.md
+++ b/docs/guide/add.md
@@ -98,6 +98,12 @@ If your file already has a bbox column but lacks covering metadata (e.g., from e
 
 ## H3 Hexagonal Cells
 
+!!! note "Input CRS"
+    The grid commands (`h3`, `s2`, `a5`, `quadkey`) key on lon/lat. Input in a
+    non-CRS84 CRS is reprojected to lon/lat automatically before cell assignment,
+    so projected data (e.g. national grids in metres) keys correctly with no
+    manual reprojection. Input already in OGC:CRS84 / EPSG:4326 is untouched.
+
 Add [H3](https://h3geo.org/) hexagonal cell IDs based on geometry centroids:
 
 === "CLI"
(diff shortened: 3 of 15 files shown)


cholmes:
I've got a couple more fixes from adversarial review coming in for this one.

Issue #545, opened 2026-07-01
Validate #540: bbox pre-filter is a no-op/pessimization on DuckDB 1.5.1; rationale is stale (perf claim not reproducible)


nlebovits:
## Summary

Hand-testing of #540 (merged as 5d3de1c) on **DuckDB 1.5.1** shows the PR is **safe to keep** — but its central *performance rationale* is **not reproducible on 1.5.1**, and the bbox pre-filter it deliberately retains for the 1.x path is now a mild **pessimization** rather than a safeguard. This ticket records the evidence and proposes low-risk follow-ups.

Context: #540 was merged without hand-testing (see the FTW/Overture discussion). This is the follow-up validation. All results below are DuckDB **1.5.1**, the version pinned in `pyproject.toml`.

Related: #538 (parent), #540 (merged), #462 (rejected direction), #460 / #457 (the original Overture hang).

## What #540 actually changed

Despite the framing, **#540 does not remove the bbox pre-filter.** The only behavioral change is the status/dry-run **message** (`Using native geometry with DuckDB SPATIAL_JOIN...` instead of the misleading "No bbox available, degraded fallback…"), plus docs and a regression test. The 1.x explicit-bbox path is byte-for-byte unchanged. So the regression risk people were worried about (rip out the pre-filter → 20-min hang / OOM on FTW/Overture) **was not taken by this PR.** Nothing merged here is a regression.

### 1. Message fix — ✅ correct and works
Verified live:
- `fields_gpq2_crs84_zstd.parquet` (native, no bbox) → prints `Using native geometry with DuckDB SPATIAL_JOIN...`
- `fields_pgo_crs84_bbox_snappy.parquet` (1.x + bbox) → prints the bbox-optimized message
- genuine no-bbox 1.x → still gets the old "No valid bbox column" warning (correct)

### 2. The "bbox AND ⇒ BLOCKWISE_NL_JOIN, ~27–73× slower" claim — ❌ does not reproduce on 1.5.1
`build_spatial_join_condition`'s docstring and the #462 rejection rest on: *"the moment any bbox-overlap test is ANDed in front, the plan drops to a `BLOCKWISE_NL_JOIN` (O(n·m)) … ~27–73× slower."*

[...]

Issue #538, opened 2026-06-30
perf(add): lean on DuckDB SPATIAL_JOIN for admin-divisions/country-codes; fix misleading native-geometry message; re-evaluate bbox pre-filter (supersedes #462)


nlebovits:
## Summary

This supersedes #462. While investigating that issue I benchmarked the actual join behavior on **DuckDB 1.5.1** and found that #462's core premise is **inverted**, and its proposed fix would make things *substantially worse*. This ticket records the corrected understanding and re-scopes the work to the changes that are actually safe and valuable.

#462 assumed:
> the format we most want to promote (native geo types) is currently the slow path … derive a bbox pre-filter on the fly from native geometry and AND the 4-comparison overlap test in front of `ST_Intersects`.

On DuckDB 1.5.1 the opposite is true: a **bare `ST_Intersects(...)` ON clause is recognized by DuckDB's `SPATIAL_JOIN` operator and is dramatically faster**, while ANDing any bbox-overlap test in front of it (whether derived inline or from explicit bbox struct columns) **defeats `SPATIAL_JOIN` and forces a `BLOCKWISE_NL_JOIN`** (O(n·m)).

The native-geometry path *already* emits a bare `ST_Intersects` ON clause (`input_bbox_col` is `None`, so `build_spatial_join_condition` returns just the intersects predicate), so for inputs/shapes where the optimizer triggers it is **already on the fast path**. No on-the-fly bbox derivation is needed — and it is actively harmful.

---

## Evidence (DuckDB 1.5.1)

200k point input `LEFT JOIN` 4k buffered-polygon admin set, identical result row counts (250,946 — LEFT semantics preserved across all three):

| Strategy | Join operator | Time |
|---|---|---|
| Bare `ST_Intersects` (today's native-geometry ON clause) | **`SPATIAL_JOIN`** | **0.17s** |
| Explicit bbox struct cols + `ST_Intersects` (#462's "fast" path / #460) | `BLOCKWISE_NL_JOIN` | 4.65s |
| Derived inline `ST_X/YMin/Max` bbox + `ST_Intersects` (#462's proposed fix) | `BLOCKWISE_NL_JOIN` | 12.35s |

`EXPLAIN` confirms `SPATIAL_JOIN` is chosen for a bare `ST_Intersects` and **still triggers** through every real query shape this command generates:
- `LEFT JOIN` to an admin **subquery with a `WHERE`** filter (the `_build_admin_subquery` shape),
[...]

Issue #462, opened 2026-06-07
perf(add): optimize admin-divisions spatial join for GeoParquet 2.0 / native Parquet geo types


cholmes:
Follow-on to #461.

## Background

PR #461 restored a plain streaming `LEFT JOIN` for `gpio add admin-divisions` to fix the OOM/hang on large inputs. As part of that work, the spatial-join condition (`build_spatial_join_condition` in `core/duckdb_utils.py`) only emits the cheap 4-comparison **bbox numeric pre-filter** when *both* sides expose a bbox covering column. When they don't, it falls back to a bare `ST_Intersects(target_geom, input_geom)` against every candidate pair.

For **GeoParquet 2.0** and **Parquet-with-native-geospatial-types** inputs, `get_bbox_advice(..., operation="spatial_filtering")` deliberately sets `skip_bbox_prefilter = True` (`core/common.py:1965`), so `input_bbox_col` is passed as `None` into the join. The design assumption is that DuckDB will use native Parquet geometry row-group statistics automatically and that the explicit bbox pre-filter is redundant.

## Problem

In practice, for these native-geometry inputs the join takes the "no bbox" path and is **substantially slower** than the equivalent run on a file that carries an explicit `bbox` struct column:

1. The user sees the warning `No bbox columns available, using full geometry intersection...` (`core/add/admin_divisions.py:434`, also `country_codes.py:519`), which is confusing on a GeoParquet 2.0 file that *does* have spatial info — it reads like the tool failed to find data it should have.
2. Without a per-row bbox numeric pre-filter in the `ON` clause, DuckDB runs full `ST_Intersects` on far more candidate pairs. Native Parquet column statistics help prune **row groups on a single scan**, but they do not give the join the cheap per-row xmin/xmax/ymin/ymax overlap test that makes the explicit-bbox path fast.

Net result: the format we most want to promote (native geo types) is currently the slow path for this command.

## Goal

Make the admin-divisions (and country-codes) spatial join on GeoParquet 2.0 / native-Parquet-geo inputs at least as fast as the explicit-bbox-column path — without reintroducing the OOM behavior fixed in #461.

## Directions to investigate

[...]

PR #546, opened 2026-07-01
test(add): guard native spatial-join stays on SPATIAL_JOIN at scale (#545, Fix C)

nlebovits:
## Fix C for #545

Adds a plan-level regression guard so a future change that reintroduces the #460 Overture hang fails CI **loudly** instead of hanging in production.

### What
New `TestSpatialJoinOperatorPlan` in `tests/test_duckdb_utils.py` (test-only, no production code changed):

- **`test_native_predicate_plans_as_spatial_join`** — fast, runs on every PR. Builds the exact native-path query the `add` commands emit (`build_spatial_join_condition` with no bbox → bare `ST_Intersects`, in the real `LEFT JOIN` shape), runs `EXPLAIN`, and asserts the plan uses `SPATIAL_JOIN` with no `BLOCKWISE_NL_JOIN` / `NESTED_LOOP_JOIN` / `CROSS_PRODUCT`.
- **`test_native_predicate_plans_as_spatial_join_at_scale`** — `@pytest.mark.slow` (CI slow job). Same assertion at **1.2M input features × 45k target polygons**, the faithful ≥1M-scale check requested in the issue.

`EXPLAIN` is plan-only, so the 1.2M-row case runs in <1s.

### Why this is the right guard
- Tied to the shared `build_spatial_join_condition` + real `LEFT JOIN` shape, so it covers both `admin-divisions` and `country-codes`.
- **Verified it has teeth:** feeding the assertion helper the bbox-AND predicate on realistic (varying) bbox data produces `BLOCKWISE_NL_JOIN` and the helper raises. So the guard genuinely fails if the fast path degrades.

### Context: #545's central finding does not reproduce
Issue #545 claims that on DuckDB 1.5.1 "all three predicate shapes stay on `SPATIAL_JOIN`". Against the real GAUL cache with the **`LEFT JOIN`** the code actually emits, the bbox-AND pre-filter drops to `BLOCKWISE_NL_JOIN` (it stays on `SPATIAL_JOIN` only under `INNER JOIN`, which the code does not use). The bare/native predicate stays on `SPATIAL_JOIN` at every scale (1k → 1.2M, ~15s at 1.2M, no hang). This guard locks in that bare fast path — the one that dropping the 1.x bbox pre-filter (Fix B) would rely on.

### Note for Fix B
Operator choice for the bbox-AND shape is **data-dependent** (degenerate/constant bboxes constant-fold back to `SPATIAL_JOIN`; realistic varying bboxes drop to `BLOCKWISE_NL_JOIN`). Any before/after benchmark for Fix B must use realistic data or it will measure a mirage.

### Test plan
- [x] `pytest tests/test_duckdb_utils.py` — 40 passed (incl. both new tests)
- [x] ruff check + format clean
- [x] marker routing confirmed: fast case in the default job, at-scale case in the `slow or network` job
- [ ] Follow-up: live remote-Overture ≥1M benchmark (in progress) to green-light Fix B

Draft until the remote-Overture validation lands.


nlebovits:
## Live remote-Overture validation (the #460 premise) — done

Ran against the real Overture release **2026-06-17.0**, DuckDB 1.5.1. Target = `theme=divisions/type=division_area` = **1,067,718 features (~6.05 GB WKB, native geometry + bbox covering)**.

**Operator at true remote ≥1M cardinality is identical to local** (`EXPLAIN`, plan-only so no full download), using the exact ON clauses `build_spatial_join_condition` emits, in the real `LEFT JOIN` shape:

| predicate | remote 1.07M target | local GAUL 45k |
|---|---|---|
| bare `ST_Intersects` (Fix B) | `SPATIAL_JOIN` | `SPATIAL_JOIN` |
| bbox-overlap AND (current 1.x) | `BLOCKWISE_NL_JOIN` | `BLOCKWISE_NL_JOIN` |

➡️ **#460's premise — "SPATIAL_JOIN does not reliably trigger" for the bare predicate on remote — does not hold on DuckDB 1.5.1.** Remote scan does not change the optimizer's operator choice. This is the green-light for Fix B's safety: dropping the bbox term moves the 1.x path from `BLOCKWISE_NL_JOIN` to the indexed `SPATIAL_JOIN`, remote included.

**No hang:** the bare path runs as real indexed work (streams the remote admin, builds the R-tree), not an O(n·m) hang.

**One caveat worth carrying into Fix B (not a blocker):** at full unpruned ≥1M scale the admin side is ~6 GB of geometry, and bare `SPATIAL_JOIN` holds/indexes that side in memory (a memory/speed tradeoff vs streaming `BLOCKWISE_NL_JOIN`). This is **already the native-path behavior today** and is mitigated by the production connection config — `admin_divisions.py` runs these joins with `threads=1` + `temp_directory` spill + `preserve_insertion_order=false` (for DuckDB #8270). Same config applies to both predicate shapes, so Fix B doesn't change the memory discipline; it just extends the fast operator to the 1.x path.

Net: Fix B is safe re: operator selection at remote Overture scale. Recommend proceeding, documenting the memory/extent-pruning behavior in the docstring rather than the current "SPATIAL_JOIN doesn't trigger on remote" rationale.

nlebovits:
## Fix B added to this branch

Second commit implements Fix B on top of the Fix C guard.

**Change:** `build_spatial_join_condition` now always emits a bare `ST_Intersects`. The per-row bbox-overlap term (AND'd in front for the 1.x path) is removed — under the `LEFT JOIN` these commands emit it forced a `BLOCKWISE_NL_JOIN`, whereas bare stays on the indexed `SPATIAL_JOIN` (validated against the remote ≥1M Overture admin, no hang).

**Retained:** the separate admin/countries-side **extent** pre-filter (WHERE clause bounding the target to the input bbox) — the actually-beneficial use of the stored bbox covering. Verified in dry-run SQL across native, 1.x-with-bbox, and reprojected (EPSG:31287) inputs: `ON ST_Intersects(...)` is bare, the `WHERE "…_bbox".xmin <= …` extent filter is intact.

**Messaging:** `spatial_join_strategy` messages no longer claim a join-level bbox pre-filter; `BBOX_PREFILTER` now denotes the extent pre-filter. Removed the now-dead `input_bbox_col`/`target_bbox_col` params from the join builders.

**Scope note:** the partition path (`admin_hierarchical.py`) has its own inline bbox pre-filter and is intentionally left for a separate, scoped follow-up — the issue's Fix B targets `build_spatial_join_condition` + the two `add` commands.

**Verification:**
- [x] Full fast suite: 3131 passed, 6 skipped
- [x] Slow at-scale SPATIAL_JOIN guard: passed
- [x] ruff / ruff-format / codespell / no-click-echo / duckdb-antipatterns / doc-sync hooks: pass
- [x] vulture, xenon (real thresholds), import-linter: pass
- [x] Docs updated (`docs/guide/add.md`) — removed the stale "#462 to be implemented / bbox-overlap 5-10x faster" claim
- [ ] Live remote-Overture ≥1M execution: operator confirmed `SPATIAL_JOIN` via EXPLAIN; full execution is memory-bound at 6GB admin and handled by the production `threads=1`+spill config (see prior comment)

nlebovits:
Closing. The ≥1M-feature validation that #545 said was still owed came back against the production change on this branch.

Despite the "Fix C, test-only" title, this branch's diff to `duckdb_utils.py` (+48/−71) also **removes** the bbox pre-filter (Fix B) — it deletes the `input_bbox_col` / `target_bbox_col` params and the 4-comparison pre-filter block, leaving a bare `ST_Intersects`. That is the change the validation disproved:

- **Fix B (remove the pre-filter → bare `ST_Intersects`)** — ❌ rejected. At FTW/Overture scale (~2M real admin polygons) the bare predicate drives DuckDB's `SPATIAL_JOIN` into an in-memory, non-spilling R-tree that OOMs — the same "effectively hangs" failure as #460. The bbox term keeps the LEFT JOIN on a streaming `BLOCKWISE_NL_JOIN` with bounded memory. So the pre-filter is a **memory-safety necessity**, not the "mild pessimization" the small-scale (#538 / 504k-row) timings suggested.
- **Fix C (guard asserting native joins stay on `SPATIAL_JOIN` / flagging `BLOCKWISE_NL_JOIN` as the bad #460 signature)** — ❌ rejected. Post-validation that signature is backwards for large real polygons; the guard would invite removing the pre-filter again.
- **Fix A (docstring / user-message correction)** — ✅ kept, moved to a focused docs-only PR: #551.

The pre-existing OOM risk on the no-bbox / native-geometry paths (which never had the pre-filter) is split out to #550. Full reasoning trail in #545.

PR #573, opened 2026-07-14
Add `--optimize-for web` for browser-ready GeoParquet output

yharby:
## Description

Adds a `--optimize-for web` profile to `gpio convert geoparquet` (and the matching Python API) that produces a GeoParquet 2.0 file tuned for browser map viewers such as hyparquet plus deck.gl or MapLibre. The profile is a single, fixed contract. It writes native Parquet `GeospatialStatistics`, keeps a covering `bbox` column for viewport pruning, byte-targets the row-group size so each row group maps to one HTTP range request, emits a Parquet page index, and uses ZSTD compression. Everything is written through the memory-safe streaming writer, so inputs larger than RAM still convert (DuckDB spills the Hilbert sort to disk).

The motivation is that server-side readers (DuckDB, warehouses) already get pushdown from the full Parquet footer regardless of layout, but browser viewers reading over HTTP range requests need fetch-sized row groups, native stats, and a covering bbox to prune efficiently. This profile encodes those best practices behind one flag instead of leaving users to hand-tune several options.

## Technical Details

New single source of truth in `core/web_profile.py` defines what `web` means (GeoParquet 2.0, native stats, covering bbox, streaming strategy, page index) and the fetch-unit sizing math.

- **Byte-targeted row groups.** `resolve_web_row_group_rows()` sizes row groups from a target compressed byte budget (default 8 MB, tunable via `--row-group-size-mb`), clamped to 10k..200k rows, with a footer guard capping total row groups at 1000 so the initial footer download stays small. An explicit `--row-group-size` wins outright.
- **Native GeospatialStatistics.** New `has_parquet_native_geo_stats()` detector in `metadata_utils.py` reads row group 0's geometry leaf via pyarrow and normalizes the bbox order to `[xmin, ymin, xmax, ymax]`.
- **Page index and data page size.** `ParquetWriteSettings` gains `write_page_index` and `data_page_size` fields plus a `for_web_profile()` factory. These are forwarded into the pyarrow writer kwargs.
- **Write path routing.** DuckDB `COPY TO` cannot emit a page index, so `write_parquet_with_metadata()` forces the strategy path (bypassing the plain-COPY fast path) whenever `write_settings` is provided, and routes web writes to the streaming strategy. The streaming strategy now honors `write_settings` for both `write_from_query` and `write_from_table`, including detecting an existing bbox column so the geo metadata advertises it.
- **Covering bbox on v2.0.** A `force_include_bbox` flag threads through the convert paths so the bbox column is kept even for version 2.0, which would normally drop it.
- **Cheap row counting.** On the file path the row count comes from a light source-level scan (`read_parquet`, `read_csv_auto`, or `ST_Read`) rather than the fully built query, to avoid re-running Hilbert sort, geometry repair, and CRS transforms a second time. On the in-memory Table path it uses Arrow `nbytes` as the sizing proxy, documented as an uncompressed signal.
- **Write-strategy contract.** `BaseWriteStrategy` and all four strategies accept a `write_settings` parameter. The three non-streaming strategies accept and ignore it, with docstrings explaining why.
- **stdout no-op warning.** Streaming `convert` to stdout warns that the web layout is dropped when re-emitted as Arrow IPC, and tells the user to write to a `.parquet` file instead.

Python API surface. New top-level `convert_geoparquet(input, output, optimize_for="web", ...)` one-shot function, exported from `geoparquet_io` and `geoparquet_io.api`. `Table.write(..., optimize_for="web")` applies the same fixed contract, adding a bbox column via `add_bbox_table` when the table lacks one.

Docs. New guide at `docs/guide/web-optimized-geoparquet.md` (added to `mkdocs.yml` nav) with CLI and Python examples, a reader contract section, and the page-index rationale. Python API reference updated.

Dependency. Bumps the pyarrow floor from `>=23.0.1` to `>=24`, required for native `GeospatialStatistics` support.

## Breaking Changes

No changes to existing APIs or command behavior. The `--optimize-for` flag and `optimize_for` parameter are new and default to `None`, so existing calls are unaffected. Note the pyarrow floor moves to `>=24`, which affects environments pinned to pyarrow 23.

## Related Issue(s)
- N/A

## Checklist
- [ ] Integration tests for changes that cross module/format/service boundaries
- [ ] All tests pass (`uv run pytest`) 
- [ ] At least one adversarial review (AI counts, rubber stamps don't)

Added tests. `tests/test_web_profile.py`, `tests/test_convert_web_optimized.py`, `tests/test_api_web_optimize.py`, `tests/test_metadata_native_geo_stats.py`, `tests/test_parquet_writer_page_index.py`.


yharby:
Over the next few days, I'll build a simple frontend reader and optimize the output parquet file for the best approach on both sides ... client-side reading via [hyparquet](https://github.com/hyparam/hyparquet) (not DuckDB-wasm or parquet-wasm) ... with the draft outputting different datasets at varying complexity and zoom levels. I think we can recommend this for the first version of the Portolan spec.

yharby:
## Real-world test + hyparquet check

Ran `--optimize-for web` on a 362 MB Finnish buildings file (5.65M rows) and read the hyparquet 1.26.2 source to see what a browser reader actually uses.

### What the profile did well
- Row groups went from **1,380 tiny ones to 46 fetch-sized ones** (~11 MB each). This is the real win, far fewer HTTP range requests.
- Added a page (offset) index, a covering `bbox` struct, and kept native geo stats. File is valid and viewer-ready.

### The catch
The output grew **362 MB to 492 MB (+130 MB)**. Cause, the input already had flat `xmin/ymin/xmax/ymax` columns, and the profile added a second `bbox` struct with the same numbers. So the bbox is stored twice.

### Does that struct help the browser reader?
I checked the hyparquet code. For row-group pruning it uses **only standard Parquet column min/max stats** on whatever column the app filters on. It does **not** read native `GeospatialStatistics`, and it does **not** read the GeoParquet `covering` field (both appear nowhere in its pruning path). It prunes the same way on a flat `xmin` column or a `bbox.xmin` struct field.

So on this file the extra struct buys hyparquet nothing. The flat columns already had the stats it needs.

### Suggested follow-up
When the input already has bbox columns, reuse them (mark them as covering, or drop and replace) instead of appending a duplicate. That removes the ~100 MB of redundant data. The consolidation + stats are what matter, not a second bbox copy.

Note, this is one reader at one version. A clean input with no bbox column still benefits from the profile adding one.

cholmes:
Seems like we should try to PR hyparquet to actually make use of this? And does DuckDB WASM make use of it? 

I think I'm ok to add it to geoparquet-io to help push things forward, but it seems like we should have clients that actually use it.

I also do think we should ultimately push these into the geoparquet distributing best practices, and as I mentioned on slack also get the bbox covering back as an option in geoparquet 2.0 (or we can make it an 'extension').

cholmes:
I do think it'd be good to get a really good performance testing suite where we can compare different variables. Like https://github.com/geoparquet/geoparquet-testing but https://github.com/geoparquet/geoparquet-performance-testing - have a bunch of files that have different row group sizes / covering / partitions. 

But I'm definitely up to do 'something' here and recommend it in Portolan - just want to reach higher and higher conviction in our recommendations, and also make it easy to experiment on new recommendations. That work across a bunch of different file sizes / types vs just testing on one or two.

yharby:
Thank you @cholmes Good points. I've actually already started experimenting with the output Parquet from this PR using **hyparquet**. I put together a small demo where you can load any remote GeoParquet URL and test browser-side rendering and filtering:

https://yharby.github.io/geoparquet-overviews/

For example:
https://yharby.github.io/geoparquet-overviews/?url=https%3A%2F%2Fdata.source.coop%2Fyoussef-harby%2Fgeoparquet-overviews%2Fgeoparquet-io-web-profile-test%2Foverture%2Fbarcelona%2Fbuildings_building.parquet&x=2.10038&y=41.37674&z=10.73

My goal is to use this as a quick win and to push towards this to compare different layouts (row group sizes, covering bbox, etc.) and measure what actually improves client-side performance. I also agree that having a dedicated performance testing suite would be really valuable, especially if we want to turn these findings into Portolan recommendations and eventually GeoParquet best practices/specs

Issue #584, opened 2026-07-21
fix(validate): empty geometries counted as 'outside bbox' in stats/bbox containment checks


cholmes:
`native_geo_stats_contains_data_*` fails all 12 valid `data/encodings/*` corpus files plus `geometrycollection.parquet` with "1 of 3 geometries fall outside geospatial statistics".

Root cause: fixtures contain `POINT EMPTY` (plus a NULL row). In `_check_native_geo_stats_contains_data` (validate.py:1586-1600) the containment CASE uses `ST_XMin(geom) >= xmin AND ...`; for an EMPTY geometry the extent functions return NULL, so the row is silently counted as not-within. Empty geometries have no extent and per the Parquet geospatial-statistics semantics should be excluded from bbox containment (same for `bbox_contains_data_*`, which shows the same false positive on rewritten files).

Fix: exclude empties (`AND NOT ST_IsEmpty(...)`) in both containment queries.

PR #590, opened 2026-07-21
test(validate): regression coverage for empty-geometry bbox exclusion

Sanjays2402:
## PR Title

Add regression coverage for the empty-geometry bbox exclusion

## Description

Empty geometries have no extent, so DuckDB returns NULL for their bounds. Main already excludes them from both metadata-bbox containment queries (the fix for #584 landed via #602 and was reworked by #662), so this PR is now tests-only: it adds regression coverage pinning the `NOT ST_IsEmpty` predicate in both generated query paths.

## Technical Details

Tests assert the WKB/GEOMETRY bbox query and the native geospatial-statistics query both filter `ST_IsEmpty` rows before counting, and that a file whose rows are all NULL/EMPTY yields SKIPPED (not a vacuous PASS) for the native-stats check.

## Breaking Changes

## Related Issue(s)
- #584 (already fixed on main by #602; this adds the regression tests)

## Checklist
- [ ] Integration tests for changes that cross module/format/service boundaries — N/A, tests-only
- [x] All tests pass (`uv run pytest tests/test_validate_empty_geometries.py`: 3 passed)
- [x] CodeRabbit comments fixed or answered
- [ ] Documentation updated (README, guides, CLI reference) — N/A, tests-only


cholmes:
Thank you so much for the PR / contribution! As I mentioned in #600 in the comment at the end we've been slammed, but will try to review / merge / cut a release with your changes soon! We appreciate your work.

Sanjays2402:
Rechecked head 55fa7711: it adds the `NOT ST_IsEmpty` predicate to both the bbox containment query and the native geospatial-statistics query, with focused coverage for each path. There is nothing further needed from my side; the branch is ready for review/merge.

cholmes:
I've updated this branch to current main (merge, conflicts resolved). One note on the resolution: main has since shipped the empty-geometry exclusion itself — #602 landed the `NOT ST_IsEmpty` filter for #584, and #662 reworked those queries to use `quote_identifier()`. So `validate.py` now takes main's version, and what this PR contributes is the regression test coverage, which I adapted to main's current query shape (quoted identifiers, and the newer zero-non-empty-rows → SKIPPED semantics, with an added test for that case). All three tests pass locally.

cholmes:
Adversarial review findings (post-merge head):

**1. Scope is now tests-only.** With main already carrying the fix (via #602/#662), this PR's value is the regression guard. The `Closes #584` in the description is stale — that issue was closed by #602 — so the description should be updated to say it adds regression coverage for the empty-geometry exclusion.

**2. The tests are white-box string assertions.** They assert `NOT ST_IsEmpty(...)` appears in the generated SQL rather than observing behavior on a real file with empty geometries. That's acceptable as a cheap regression guard, but it will break on any harmless refactor of the query text (as this merge itself demonstrated — the original assertions no longer matched main's quoting). A behavioral test — a small parquet fixture with one empty + one real geometry, asserting the check passes — would survive refactors. Fine to land as-is; noting the trade-off.

**3. `_RecordingConnection` bypasses `_is_bbox_valid` and remote checks implicitly** — if those code paths reorder, the mock's `queries[-1]` assumption could silently test the wrong query. A targeted `assert "parquet_metadata" not in containment_query` would make the guard self-checking.

Nothing blocking — with the description updated this is a reasonable tests-only merge.

PR #606, opened 2026-07-25
Fix misleading 'Batch size 0 too large' on non-paged ArcGIS requests

### LINKED ISSUE #485: extract arcgis raises misleading 'Batch size 0 too large' when the count query returns non-JSON (WAF block pages misdiagnosed, no User-Agent sent)
## Summary

When an ArcGIS server answers any **non-paged** request (the `returnCountOnly=true` count query or the layer-info request) with a non-JSON body, gpio raises `BatchTooLargeError` with `batch_size=0`, producing

```
Error: Batch size 0 too large for https://sigma.madrid.es/hosted/.../0: Server returned non-JSON response (likely HTML error page)
```

This is misleading in several ways. There is no batch on a count query, so the adaptive fallback from #382 can never apply. The real diagnostics (HTTP status code, Content-Type, first bytes of the body) are discarded. And because `BatchTooLargeError` is treated as not retryable with the same params, a transient WAF block page kills the extract on the first request with a message that sends the user hunting in the wrong direction.

Verified present at `4ccb4f6` and unchanged on current `main` (`4ff5dab`).

## Real-world incident

Full-catalog extraction of 16 layers from `https://sigma.madrid.es/hosted/rest/services/` (Madrid city ArcGIS) running in GitHub Actions. Every layer failed instantly with the error above. The same gpio command at the same commit succeeded from a residential machine minutes later (305 features, no errors), and small `--limit 5` smoke extracts from a GitHub runner had succeeded 3 minutes earlier. The server's WAF was serving an HTML block page to the runner's datacenter IP. gpio reported it as a batch-size problem with batch size zero.

## Root cause

1. `geoparquet_io/core/arcgis.py`, `get_feature_count()` calls `_make_request("GET", query_url, params=params)` without a `batch_size` argument. Same for the layer-info request.
2. `geoparquet_io/core/http_retry.py`, `make_request_with_retry()` catches `json.JSONDecodeError` and raises

```python
raise BatchTooLargeError(
    url=url,
    batch_size=batch_size or 0,   # <- count query has no batch, becomes 0
    reason="Server returned non-JSON response (likely HTML error page)",
) from e
```

3. `BatchTooLargeError.__str__` renders `Batch size 0 too large for ...`.

The comment in `http_retry.py` says "This is NOT retryable with same params - batch is too large", which is only true for the paged feature fetch. For the count and layer-info requests the assumption does not hold, non-JSON there means an auth wall, a WAF block page, a proxy error page, or an HTML 500, all of which deserve a different error and are often transient.

## Suggested fixes

1. Only raise `BatchTooLargeError` when the failing request actually carried a batch size (`batch_size is not None`). Otherwise raise `RemoteAccessError` (or a new error) that includes the HTTP status code and Content-Type, e.g. `Server returned text/html (HTTP 200) instead of JSON for the feature-count request, possibly a WAF or proxy block page`.
2. Include status code and Content-Type in the `BatchTooLargeError` reason as well, it costs nothing and makes WAF blocks distinguishable from genuine payload-limit truncation.
3. Send an identifying User-Agent. gpio currently sends none, so httpx's default `python-httpx/x.y` goes out, which is prime WAF bait from datacenter IPs. Something like `geoparquet-io/<version> (+https://github.com/geoparquet/geoparquet-io)` in `make_request_with_retry`'s default headers, ideally overridable, would make gpio a better-behaved client and likely avoids some of these blocks entirely.
4. Optionally allow one retry with backoff when a non-paged request returns non-JSON, since block pages and proxy errors are frequently transient.

## Reproduction

Hard to reproduce on demand since it needs a server that serves an HTML page to some clients, but the error path is unit-testable, call `make_request_with_retry(..., parse_json=True, batch_size=None)` against anything returning HTML and observe `Batch size 0 too large`.

Sanjays2402:
Closes #485

## Description

The ArcGIS count and layer-info queries call `make_request_with_retry` without a `batch_size`, so any non-JSON body (WAF block page, proxy error page, HTML 500) raised `BatchTooLargeError` with `batch_size=0` and printed `Batch size 0 too large for <url>`. There is no batch on those requests, so the adaptive batch fallback can never apply and the message points at the wrong cause.

## Technical Details

`make_request_with_retry` now raises `RemoteAccessError` on a non-JSON response when `batch_size is None`; the paged feature fetch keeps raising `BatchTooLargeError` with its real batch size. Both messages now include the response Content-Type and HTTP status, which makes a block page distinguishable from genuine payload truncation. Response bodies are still not included.

Scope note: this PR covers suggestions 1 and 2 from the issue. The User-Agent header (3) and the retry-on-non-JSON change (4) are behavioural changes to every outbound request and are left as follow-ups.

## Breaking Changes

## Related Issue(s)
- #485

## Checklist
- [ ] Integration tests for changes that cross module/format/service boundaries — N/A, unit-level error mapping
- [x] All tests pass (`uv run pytest`) — `tests/test_arcgis.py` error/retry/timeout tests pass; new test fails without the fix and passes with it
- [ ] CodeRabbit comments fixed or answered — none yet
- [ ] Documentation updated (README, guides, CLI reference) — N/A, error message only

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error handling for non-JSON responses from remote services.
  - Reports clearer details, including response type and HTTP status.
  - Prevents non-paged requests from being incorrectly treated as oversized batches.
  - Preserves the requested batch size when reporting paged request failures.

- **Tests**
  - Added regression coverage for HTML and other non-JSON responses in both request scenarios.

### DIFF
--- a/geoparquet_io/core/http_retry.py
+++ b/geoparquet_io/core/http_retry.py
@@ -150,13 +150,27 @@ def make_request_with_retry(
                 try:
                     return cast(dict[str, Any], response.json())
                 except json.JSONDecodeError as e:
-                    # Server returned non-JSON (likely HTML error page)
+                    # Server returned non-JSON (likely HTML error page).
+                    # Don't include response content in error - may contain
+                    # sensitive tokens; the status code and Content-Type are
+                    # safe and make a WAF/proxy block page distinguishable
+                    # from a genuine payload-size truncation.
+                    content_type = response.headers.get("content-type", "unknown")
+                    detail = (
+                        f"Server returned non-JSON response "
+                        f"(Content-Type: {content_type}, HTTP {response.status_code}), "
+                        f"likely an HTML error or block page"
+                    )
+                    if batch_size is None:
+                        # No batch was requested (e.g. a feature-count or
+                        # layer-info query), so reducing the batch size cannot
+                        # help and BatchTooLargeError would be misleading.
+                        raise RemoteAccessError(url, detail) from e
                     # This is NOT retryable with same params - batch is too large
-                    # Don't include response content in error - may contain sensitive tokens
                     raise BatchTooLargeError(
                         url=url,
-                        batch_size=batch_size or 0,
-                        reason="Server returned non-JSON response (likely HTML error page)",
+                        batch_size=batch_size,
+                        reason=detail,
                     ) from e
             else:
                 return bytes(response.content)
--- a/tests/test_arcgis.py
+++ b/tests/test_arcgis.py
@@ -442,6 +442,53 @@ class TestTimeout:
 
         assert mock_client.get.call_args.kwargs["timeout"] == 300.0
 
+    def test_non_json_without_batch_size_is_remote_access_error(self):
+        """A non-paged request that gets HTML must not report a batch problem.
+
+        Regression test for #485: the count and layer-info queries carry no
+        batch size, so a WAF/proxy block page was reported as
+        "Batch size 0 too large", hiding the real cause.
+        """
+        from geoparquet_io.core.exceptions import BatchTooLargeError, RemoteAccessError
+        from geoparquet_io.core.http_retry import make_request_with_retry
+
+        mock_client = MagicMock()
+        mock_response = Mock()
+        mock_response.json.side_effect = json.JSONDecodeError("x", "<html>", 0)
+        mock_response.raise_for_status = Mock()
+        mock_response.status_code = 200
+        mock_response.headers = {"content-type": "text/html"}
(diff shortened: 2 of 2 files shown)


cholmes:
@Sanjays2402 - apologies for the super slow response! We really appreciate your contribution, it's just @nlebovits and I have both been slammed, and are right now prioritizing https://portolan-sdi.org which builds on gpio and hopefully will bring a larger audience of users. But we'll try to get to a review of this as soon as we can.

Thank you!

nlebovits:
LGTM!

Issue #625, opened 2026-08-05
Inconsistency between CRS handling in Python API and CLI


fnattino:
Hi there, thanks a lot for this fantastic library! 

I noticed an inconsistency between how the CRS is handled during conversions to GeoParquet via the documented Python API and how it is handled via the CLI - not sure whether it's a bug or I am missing something.

Specifically, a dataset originally in a different CRS seems to be wrongly labelled as OGC:CRS84 when conversion is carried out via the Python API (the correct CRS is instead kept when the CLI is used).

Here is a minimal reproducible example:

```python
import geopandas as gpd
import geoparquet_io as gpio
import shapely

# create a geopackage file in EPSG:28992
geometry = [shapely.from_wkt("POINT (0 0)")]
gdf = gpd.GeoDataFrame(geometry=geometry, crs="EPSG:28992")
gdf.to_file("tmp.gpkg")

# convert to geoparquet via the Python API
gpio.convert("tmp.gpkg").write("tmp.parquet")

# check the output - the CRS is now wrongly set as OGC:CRS84
gpio.read("tmp.parquet").info()
## Table: 1 rows, 1 columns
## Geometry: geometry
## CRS: OGC:CRS84 (default)
## Bounds: [0.000000, 0.000000, 0.000000, 0.000000]
## GeoParquet: 1.1

# now the same conversion with the function internally used by the CLI
gpio.core.convert.convert_to_geoparquet("tmp.gpkg", "tmp.parquet")

# check the output - the CRS is correct
gpio.read("tmp.parquet").info()
## Table: 1 rows, 2 columns
## Geometry: geom
## CRS: EPSG:28992
## Bounds: [0.000000, 0.000000, 0.000000, 0.000000]
## GeoParquet: 1.1
```

cayetanobv:
Hit this in production via portolan-cli 1.0.0b0, which converts through the Python Table API (`gpio.convert(src)` → `.add_bbox()` / `.sort_hilbert()` → `.write()`).

Use case: a 28-collection catalog mirroring Junta de Extremadura open data (SITEX), sources in EPSG:25829/25830, both Shapefile and GeoPackage. Every output parquet came out `"crs": null`, and everything downstream trusted that: STAC collection bboxes were published in meters, catalog extent aggregation dropped all 28 collections, and PMTiles generation failed (tippecanoe exit 110 — every clamped coordinate collapses to one point).

[...]

PR #644, opened 2026-08-19
Fix Table API dropping the source CRS between convert() and write()

cayetanobv:
## Description

`gpio.convert('data.gpkg').write('out.parquet')` wrote `"crs": null` (= OGC:CRS84) for projected sources: `convert()` received the detected CRS from `read_spatial_to_arrow` and discarded it. The CLI/function path preserved CRS, which is the inconsistency reported in #625. Downstream this shipped wrong bboxes and broken PMTiles in portolan-cli, which converts through this API.

The Table now carries the detected CRS as a fallback for `.crs` (CRS embedded in the Arrow table still wins), `_write_geoparquet` picks it up through the existing `input_crs` plumbing, and chained operations (`add_bbox`, `sort_hilbert`, …) propagate it via a `_wrap` helper — they previously also dropped the auto-version hint, which now survives too.

## Technical Details

- `Table.__init__` gains an optional `crs` hint; `convert()` passes `detected_crs`.
- `Table.crs`: embedded CRS first, hint as fallback.
- 17 `return Table(...)` sites in chained methods now go through `self._wrap(...)`.
- No writer changes: all write strategies already accepted `input_crs`.

Verified on the real data from the report follow-up (Junta de Extremadura, EPSG:25829 shapefile and EPSG:25830 GeoPackage): plain, chained, and parquet-passthrough writes all keep their EPSG now.

## Breaking Changes

## Related Issue(s)
- #625

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved coordinate reference system (CRS) information when converting and writing GeoParquet files.
  * Maintained CRS metadata across chained table operations, slicing, transformations, and metadata updates.
  * Ensured detected CRS information remains available when source metadata lacks embedded CRS details.
  * Embedded CRS metadata continues to take precedence when available.
  * Updated CRS information correctly when reprojecting data to a new coordinate system.

### DIFF
--- a/geoparquet_io/api/table.py
+++ b/geoparquet_io/api/table.py
@@ -486,14 +486,24 @@ class Table:
         self._auto_version_hint: str | None = None
         self._crs_hint: dict | str | None = crs
 
-    def _wrap(self, table: pa.Table, geometry_column: str | None) -> Table:
+    _KEEP_CRS_HINT = object()
+
+    def _wrap(
+        self,
+        table: pa.Table,
+        geometry_column: str | None,
+        crs: dict | str | None | object = _KEEP_CRS_HINT,
+    ) -> Table:
         """Build a derived Table, carrying the read-time hints forward.
 
         Chained operations (add_bbox, sort_hilbert, ...) produce new Arrow
         tables that still contain no embedded CRS, so the CRS hint (and the
-        auto-version hint) must survive the chain to reach write().
+        auto-version hint) must survive the chain to reach write(). An
+        operation that changes the CRS (reproject) passes ``crs=`` explicitly
+        so the hint never mislabels transformed coordinates.
         """
-        derived = Table(table, geometry_column=geometry_column, crs=self._crs_hint)
+        hint = self._crs_hint if crs is Table._KEEP_CRS_HINT else crs
+        derived = Table(table, geometry_column=geometry_column, crs=hint)
         derived._auto_version_hint = self._auto_version_hint
         return derived
 
@@ -1737,7 +1747,9 @@ class Table:
             geometry_column=self._geometry_column,
             assume_crs84=assume_crs84,
         )
-        return self._wrap(result, self._geometry_column)
+        # The source-CRS hint would mislabel transformed coordinates; the
+        # reprojected table's CRS is target_crs by construction.
+        return self._wrap(result, self._geometry_column, crs=target_crs)
 
     def partition_by_quadkey(
         self,
@@ -2029,7 +2041,7 @@ class Table:
         if n < 0:
             raise ValueError(f"n must be non-negative, got {n}")
         n = min(n, self.num_rows)
-        return Table(self._table.slice(0, n), self._geometry_column)
+        return self._wrap(self._table.slice(0, n), self._geometry_column)
 
     def tail(self, n: int = 10) -> Table:
         """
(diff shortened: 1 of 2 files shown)


cayetanobv:
Both findings were valid — fixed in the follow-up commit: `head()`/`tail()`/`add_bbox_metadata()` now go through `_wrap`, and `reproject()` sets the hint to `target_crs` via a new `crs=` override on `_wrap` (a stale source hint on transformed coordinates would have been worse than the original bug). Regression tests added for both.

cholmes:
## Review: no blocking findings ✅

Reviewed in an isolated worktree; everything below was verified by running code rather than by inspection.

**The fix looks correct and complete.**

- `_wrap` covers **every** chained operation that returns a derived `Table` — I enumerated them all. The three remaining bare `Table(...)` constructions are correct as-is: `read` and `read_partition` read GeoParquet (CRS is embedded in the Arrow table), and `extract_arcgis` attaches its own `geo` metadata at `core/arcgis.py:1370`, so `extract_crs_from_table` finds it. No silent droppers left.
- `reproject` explicitly passing `crs=target_crs` is the right call, and the comment explaining why is appreciated — that's exactly the case where a stale source hint would mislabel transformed coordinates.
- The precedence (embedded CRS wins, detected CRS is only a fallback) is applied consistently on the write path via `self.crs` at `api/table.py:1058`.
- The `_KEEP_CRS_HINT` sentinel is the right way to distinguish "don't change it" from "set it to None".

**Integration with the rest of the batch is clean.** I merged all four PRs onto `main` and checked the result: the merged `convert()` correctly retains both this PR's `crs=detected_crs` and #647's new parameters, with no conflict. Worth noting explicitly because #647 adds an *alternate* read path (`ST_Read(keep_wkb := true)`) that could plausibly have reintroduced this exact bug — it doesn't, because `detected_crs` is resolved from `crs_from_file` *before* the query runs, outside `_read_spatial_to_arrow`, so the fallback inherits it. I verified end-to-end that converting a curved GeoPackage still writes a real CRS rather than `null`.

**Test suite:** 3543 passed. One run showed a failure in `tests/test_pipe_integration.py::test_add_bbox_to_sort_hilbert`, which was alarming given this PR touches `add_bbox` and `sort_hilbert` — but it is **not reproducible** (3 clean reruns, including under `-n auto`). Load-related flake, not a regression. `tests/test_geojson_stream.py::test_streaming_handles_broken_pipe` fails on `main` too.

**One suggestion, non-blocking:** the property that makes this work — CRS detection sitting *outside* `_read_spatial_to_arrow` — isn't asserted anywhere. That's the kind of thing that regresses silently, which is what this PR is fixing in the first place. A test that converts a projected source through the curved/fallback path and asserts the written CRS would pin it. (Related: #647's fixture has `srs_id 0`, so there's currently zero coverage of curved + projected.)

Suggested merge order across the batch: **#645 → #646 → #647 → #644** (#645 first since it's independent and #647 widens its exposure; #647 must follow #646 as its base). #644 and #647 are mechanically interchangeable — putting #644 last just means its `_wrap` audit runs against the final `convert()` signature.

cayetanobv:
Thanks! The read-level half of your suggestion landed in #647 (a projected curved GPKG test pinning that CRS detection still fires on the fallback path). The write-level assert needs this PR's fix plus the linearize path, so per your merge order I'll add the end-to-end curved+projected write test here when #644 rebases onto the merged batch.

Issue #642, opened 2026-08-18
convert() segfaults (SIGSEGV) on a GeoPackage containing NULL-geometry rows


cayetanobv:
### Bug Description

Converting a real-world GeoPackage with the Table API segfaults the Python process (SIGSEGV, exit 139) — no exception, no message. The file is Junta de Extremadura open data ("terrenos acotados a la caza", 4,112 MultiPolygon features, 3 of them with NULL geometry). Removing the 3 NULL-geometry rows makes the identical file convert cleanly, and plain DuckDB spatial reads/COPYs the original fine, so the crash is in the gpio pipeline, not the read.

Because it is a hard crash rather than an error, portolan-cli (which converts through this API) loses its whole worker pool with no indication of which file was at fault.

### Steps to Reproduce

```python
import geoparquet_io as gpio
gpio.convert("terrenos_acotados.gpkg").write("out.parquet")  # SIGSEGV
```

Note: a minimal 2-feature file with a NULL geometry does NOT reproduce — the crash needs an interaction with something else in this file (it also carries an anomalous Real-typed `fid` attribute column). Subsets rebuilt via `ogr2ogr -sql` don't reproduce either, so I stopped reducing; the original file reproduces 100% of the time.

### Expected Behavior

Convert (GeoParquet permits null geometry), or raise a Python error naming the file.

### Actual Behavior

`Segmentation fault (core dumped)`, exit 139.

### Environment

Linux, Python 3.12, geoparquet-io @ git main (via portolan-cli 1.0.0b0), bundled DuckDB spatial.

### Sample Data (Optional)

The 53 MB crashing file is CC-BY 4.0 (CICTEX, Junta de Extremadura) — comment and I'll share a download link.

### Logs (Optional)

None produced — that's part of the problem.

### Would you like to submit a PR?

Happy to test a fix against the file.

cayetanobv:
Working on a fix for this.

cholmes:
**Recording the root cause, since it is now known.**

This was closed earlier, but #737 identified what was almost certainly behind it: DuckDB
<= 1.5.1 applies the selection vector twice in `TRY()` under conditional execution
(duckdb/duckdb-spatial#858), so `TRY(ST_GeomFromWKB(...))` over real WKB can read
uninitialised vector memory and SIGSEGV the process.

`convert()` reaches that path — `geoparquet_io/core/convert.py:1457` calls
[...]

PR #645, opened 2026-08-19
Fix SIGSEGV in geometry repair on tables with NULL geometry rows

cayetanobv:
## Description

`gpio.convert()` crashed the whole Python process (SIGSEGV, exit 139) on a real-world GeoPackage with 4,112 features, 3 of them NULL geometry. `faulthandler` pinned it to the invalid-geometry count in `repair_arrow_table_geometry`: filtering `IS NOT NULL` and evaluating `ST_IsValid`/`ST_GeomFromWKB` over raw rows in one WHERE clause misaligns DuckDB spatial's selection vector — the same bug family as the OR-form repair crash already documented in this module.

Fix: a layered count (`_layered_invalid_count_sql`) — project the parsed geometry, filter NULLs in a middle layer, run `ST_IsValid` outermost. Logically identical, empirically crash-free (variant-tested A–D against the crashing file), applied to both the Arrow and query repair paths.

## Technical Details

The crashing file converts cleanly now: all 4,112 rows survive including the 3 NULL geometries. The trigger needs the file's real polygon blobs (synthetic points, chunked or not, don't reproduce), so the regression tests pin the NULL-row contract of both repair paths rather than the crash itself; the 53 MB reproduction file (CC-BY 4.0, Junta de Extremadura) is available on request — it's the one from #642.

## Breaking Changes

## Related Issue(s)
- #642

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved geometry repair reliability for datasets containing NULL or invalid geometries.
  - Preserved NULL geometry rows and overall row counts during repair.
  - Improved query-based repair to accurately detect invalid geometries and repair affected data.

- **Tests**
  - Added coverage for NULL geometry handling across table and query workflows.
  - Added large-scale regression coverage for mixed NULL and invalid geometries.

### DIFF
--- a/geoparquet_io/core/geometry_repair.py
+++ b/geoparquet_io/core/geometry_repair.py
@@ -52,6 +52,25 @@ def repair_geometry_sql(geom_expr: str) -> str:
     )
 
 
+def _layered_invalid_count_sql(source_sql: str, parsed_expr: str) -> str:
+    """COUNT of invalid geometries, in a shape DuckDB executes without crashing.
+
+    Filtering ``IS NOT NULL`` and evaluating ``ST_IsValid`` over the raw rows in
+    one WHERE clause segfaults DuckDB 1.5.x's spatial extension when the column
+    contains NULLs (selection-vector misalignment under conditional execution;
+    the same bug family as the OR-form repair crash noted below). Verified on
+    issue #642's reproduction: projecting the parsed geometry first, filtering
+    NULLs in a middle layer, and running ``ST_IsValid`` outermost is logically
+    identical and crash-free.
+    """
+    return (
+        f"SELECT COUNT(*) FROM ("
+        f"SELECT __g FROM (SELECT {parsed_expr} AS __g FROM {source_sql}) "
+        f"WHERE __g IS NOT NULL"
+        f") WHERE NOT ST_IsValid(__g)"
+    )
+
+
 def repair_query_geometry(con, query: str, geometry_column: str, *, repair: bool = True) -> str:
     """Count and optionally repair invalid geometry in an arbitrary query.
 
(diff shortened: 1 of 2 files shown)


cholmes:
## Review: 2 non-blocking findings

Reviewed in an isolated worktree. Everything below was verified by running code.

**The fix itself is correct.** I worked through the semantic equivalence case by case — NULL geometry, empty geometry, invalid-but-parseable WKB, unparseable WKB, zero rows, all-NULL — and the layered count matches the old predicate on every one, for both the `GEOMETRY` and `BLOB` branches. It also holds up at scale: clean at 20k, 50k and 200k rows.

---

### 🟡 The new tests pass against the *unfixed* code

`test_repair_arrow_table_geometry_with_null_rows` and `test_repair_query_geometry_with_null_rows` build a 5-row fixture. I ran the **old** single-WHERE count against the PR's own fixture:

```
OLD form on the 5-row fixture -> (1,)   exit 0    # correct answer, no crash
```

So the tests document the intended behaviour but wouldn't catch a revert to the single-WHERE form.

The good news is that a genuine regression test is achievable — it just needs scale. Probing the old form with a synthetic table (NULL every 997 rows, invalid every 13):

```
n=100    exit 0
n=2049   exit 0
n=5000   exit 0
n=8000   exit 139  SIGSEGV
n=10000  exit 139  SIGSEGV
n=15000  exit 139  SIGSEGV
n=20000  nondeterministic
```

The layered form was clean at all of those. Suggest keeping the existing small tests for semantics and adding one ~10k-row NULL-bearing case (`@pytest.mark.slow` if runtime matters). Since it's memory corruption it's nondeterministic at the margin, so it'll catch a regression most of the time rather than always — still far better than today.

---

### 🟡 The fix depends on optimizer behaviour that nothing enforces

`EXPLAIN` of the generated SQL:

```
UNGROUPED_AGGREGATE
  FILTER        (NOT ST_IsValid(__g))          <- separate operator
    PROJECTION  __g
      SEQ_SCAN  Filters: (TRY(st_geomfromwkb(geometry)) IS NOT NULL)
```

The optimizer already pushes the NULL filter down into the scan while leaving `ST_IsValid` above the projection. The fix works precisely because those two *don't* get merged — but that's the optimizer's current choice, not something the SQL pins. A future DuckDB that combines adjacent filters more aggressively would bring the crash back.

I don't think this should block the merge: it's a real improvement over the status quo, the code comment documents the empirical finding well, and a regression would surface as a hard crash in CI rather than a silent wrong answer — *provided* there's a test at sufficient scale, which is the finding above. The two are worth fixing together.

Also worth doing: this now has a minimal reproduction, so it's a good candidate for an upstream issue against `duckdb/duckdb-spatial`. The real defect is theirs.

---

Merge order suggestion across the batch: this one **first**. It's independent of the other three, and #647's fallback path reliably hands `repair` the exact NULL-in-WKB shape that triggers #642 — so it widens the exposure this PR closes.

cayetanobv:
Added the 10k-row case with your exact recipe (a622143) — runs in CI (~2s), pins repair/NULL counts at scale. One honest note: on my machine the old single-WHERE form survives even at 20k through the real repair path (9/9 runs) — matches what we saw writing the fix, where only real files crashed. Environment-dependent, as you said; the test catches the revert where it manifests. Will draft the upstream duckdb-spatial issue with the recipe plus the real-file evidence.

cayetanobv:
Upstream issue filed: duckdb/duckdb-spatial#858.

Issue #643, opened 2026-08-18
Curved geometries (CurvePolygon/MultiSurface, WKB 8-12) fail with opaque 'Unsupported geometry type in WKB'


cayetanobv:
### Bug Description

Sources containing curved geometries (CircularString, CompoundCurve, CurvePolygon, MultiCurve, MultiSurface — ISO WKB codes 8–12) fail conversion with `Unsupported geometry type in WKB`. GDAL reads them fine; DuckDB spatial's WKB parser rejects them. The error names neither the offending rows/types nor a remedy.

Real-world impact: GeoPackage layers *declared* MultiPolygon can silently mix MULTISURFACE rows in (ArcGIS/QGIS exports do this). Example: Junta de Extremadura's CORINE Land Cover 2018 GPKG — `gpkg_geometry_columns` says MultiPolygon, rows include MULTISURFACE.

Since GeoParquet cannot represent curves (the spec's `geometry_types` vocabulary is the linear seven ± Z), the correct conversion semantic is linearize-on-ingest — a converter concern.

### Steps to Reproduce

```bash
printf 'wkt,id\n"CURVEPOLYGON(CIRCULARSTRING(0 0,1 1,2 0,1 -1,0 0))",1\n' > curve.csv
ogr2ogr -f GPKG curve.gpkg curve.csv -oo GEOM_POSSIBLE_NAMES=wkt -nlt CURVEPOLYGON
python -c "import geoparquet_io as gpio; gpio.convert('curve.gpkg').write('out.parquet')"
```

### Expected Behavior

Ideally: linearize and convert. At minimum: an error naming the curved type(s) found and suggesting `ogr2ogr -nlt CONVERT_TO_LINEAR`.

### Actual Behavior

`GeoParquetError: Reading failed: Invalid Input Error: Unsupported geometry type in WKB`

### Environment

Linux, Python 3.12, geoparquet-io @ git main (via portolan-cli 1.0.0b0), bundled DuckDB spatial.

### Sample Data (Optional)

Repro above is self-contained. Cheap GPKG pre-detection is possible with stdlib sqlite3+struct (scan geometry blob headers for WKB codes 8–12). For actual linearization, GDAL has the RFC 49 switch (`OGRSetNonLinearGeometriesEnabledFlag(FALSE)`) that makes layers deliver pre-stroked linear geometry — worth asking duckdb-spatial whether its GDAL binding can expose it; fallback is an ogr2ogr shell-out when present.

[...]

PR #646, opened 2026-08-19
Name the offending types when non-linear WKB geometries fail conversion

cayetanobv:
## Description

Converting a source with curved geometries died with DuckDB's bare `Unsupported geometry type in WKB` — no hint of which rows, types, or what to do. Real-world case: Junta de Extremadura's CORINE GeoPackage, declared MultiPolygon in `gpkg_geometry_columns` but containing MULTISURFACE rows.

Now the error names the types and the remedy:

```
curve.gpkg contains non-linear geometries (CURVEPOLYGON), which DuckDB cannot parse
and GeoParquet cannot represent. Linearize the source first, e.g.
`ogr2ogr -nlt CONVERT_TO_LINEAR out.gpkg in.gpkg`.
```

**Scope note (from review):** this PR improves the error at the Python API level (`read_spatial_to_arrow`). `gpio convert` goes through a separate SQL path and still shows the raw DuckDB error with only this PR — CLI parity lands in #647 (stacked on this one), which wires curve handling into `convert_to_geoparquet` and adds the CLI flags.

## Technical Details

New `core/curved_geometry.py`: a stdlib (sqlite3+struct) scan of GPKG geometry blob headers for the non-linear WKB families (curves 8–12, surfaces 15–17), plus the message builder; `read_spatial_to_arrow` routes the DuckDB error through it. The scan is capped at 100k rows per feature table and escapes table/column identifiers. Detection is diagnostic-only (never gates a successful read). Non-GPKG sources get a generic message with the same remedy.

Since GeoParquet's `geometry_types` vocabulary is the linear set and duckdb-spatial has stated there are no plans to support circular types (duckdb/duckdb-spatial#510), this boundary is permanent — actual linearize-on-ingest is implemented in #647.

Adds a 96 KB curved-geometry GPKG fixture.

## Breaking Changes

## Related Issue(s)
- #643

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages when converting GeoPackages containing unsupported curved or non-linear geometries.
  * Identifies detected geometry types when possible and provides guidance to linearize geometries before conversion.
  * Preserves clear fallback errors for unreadable files, malformed data, and other conversion failures.


cayetanobv:
Ready for review. Heads-up: #647 is stacked on this branch and adds the actual linearization (same repair_geometry-style policy, opt-out included) — reviewing this one first keeps both small. Story and verification numbers in https://github.com/geoparquet/geoparquet-io/issues/643#issuecomment-5353312401.

cholmes:
## Review: 1 blocking finding (shared with #647), 3 minor

Reviewed in an isolated worktree. Everything below was verified by running code.

**The module itself is solid.** I probed the robustness paths directly and they all behave: a truncated file, a non-GeoPackage file with a `.gpkg` suffix, and a missing path each return `[]` cleanly rather than raising, and the original DuckDB error is preserved in the fallback message. The bounds checks (`len(blob) < 8`, `len(blob) < off + 5`), the per-geometry endianness handling, the `% 1000` Z/M/ZM normalization, and the `(sqlite3.Error, struct.error, OSError)` catch-all are all the right calls for parsing a file that has *already* failed to parse. Read-only URI mode is a nice touch.

---

### 🔴 The new error never reaches the CLI

This is the blocking one, and it applies to #647 equally.

```
$ gpio convert geoparquet tests/data/curved_geometry_test.gpkg out.parquet
Converting tests/data/curved_geometry_test.gpkg...
Error: Conversion failed: Invalid Input Error: Unsupported geometry type in WKB
```

versus the Python API on the same file:

```
>>> gpio.convert('tests/data/curved_geometry_test.gpkg')
curved_geometry_test.gpkg contains non-linear geometries (CURVEPOLYGON),
which DuckDB cannot parse and GeoParquet cannot represent. ...
```

The hook at `core/convert.py:1236` is inside `read_spatial_to_arrow`, which has exactly one caller:

```
$ grep -rn "read_spatial_to_arrow(" geoparquet_io/
  core/convert.py:1192   # internal _read_spatial_to_arrow
  api/table.py:329       # <- the only external caller
```

`gpio convert` goes through `convert_to_geoparquet` (`core/convert.py:1526`) → `_convert_spatial_path` (`:982`, called at `:1641`), a separate SQL path that never touches the hook. So from the primary interface, #643 is still open.

Two reasonable ways forward: route the CLI through the same hook, or land this as API-only and amend the PR description with a follow-up issue for CLI parity. Either is fine — it just shouldn't merge as-is while the description implies the CLI is fixed.

---

### 🔵 Unbounded scan on the diagnostic path

`find_non_linear_gpkg_types` (`core/curved_geometry.py:45`) reads every geometry of every feature table with no `LIMIT` and no early exit. The cursor streams and only touches blob headers, so it's not memory-hungry — but on a multi-million-feature GeoPackage the user waits a long time for what is only an error message. A generous per-table cap (say 100k rows, noting the message is partial if hit), or breaking out once `found` is non-empty, would bound it. This gets slightly worse alongside #647, where the scan can now fire *after* a full linearize attempt.

### 🔵 Identifier quoting

Table and column names from `gpkg_contents` are interpolated with `"` quoting and no escaping (`:63-66`). The database is entirely attacker-controlled already and the connection is read-only, so the practical impact is a confusing error rather than a vulnerability — but a table name containing a `"` would break the scan.

### 🔵 Error-message double-wrap with #647

Once #647 lands, `:1236` here and `:1354` there both branch on the same string `"Unsupported geometry type in WKB"`. #647's own message doesn't contain that phrase, so this outer handler misses it and re-wraps — same underlying condition, two message shapes:

```
opt-out (linearize_curves=False):
  "curved.gpkg contains non-linear geometries (CURVEPOLYGON)... CONVERT_TO_LINEAR..."
unlinearizable geometry:
  "Reading failed: curved.gpkg contains non-linear geometries..."
```

A shared exception type that the outer handler recognizes and passes through unwrapped would fix it.

cholmes:
Above generated by claude - on:
>Two reasonable ways forward: route the CLI through the same hook, or land this as API-only and amend the PR description with a follow-up issue for CLI parity. Either is fine — it just shouldn't merge as-is while the description implies the CLI is fixed.

I prefer it not to be API-only, and I'd say we should do the CLI parity asap - either in this PR or a follow-on that lands soon.

cayetanobv:
Scan capped at 100k rows per table and identifiers quote-escaped (3f24514), with tests. Description amended with the scope note: this PR is API-level; CLI parity isn't a follow-on — it's already implemented in #647 (GPKG pre-scan strategy, linearized view in `convert_to_geoparquet`, CLI flags), which also fixes the double-wrap by re-raising `GeoParquetError` untouched.

PR #647, opened 2026-08-19
Linearize curved WKB geometries on read instead of failing

cayetanobv:
## Description

Builds on #646 (detection + actionable error): curved sources now **convert** instead of erroring. When the normal read hits DuckDB's `Unsupported geometry type in WKB`, the file is re-read with `ST_Read(keep_wkb := true)` — the escape hatch DuckDB documents for geometry subtypes it cannot represent — curved blobs are stroked into linear WKB in pure Python, and the result is validated through `ST_GeomFromWKB` so the output is indistinguishable from a normal read. `gpio convert` of the CURVEPOLYGON fixture now yields a Polygon whose area matches the true circle within 0.08%.

Why here and not upstream: duckdb-spatial has stated there are no plans to support circular geometry types (duckdb/duckdb-spatial#510), and GDAL's RFC 49 linearize flag cannot help because ST_Read's GPKG Arrow fast path streams raw blobs past the OGR geometry model. Linearize-at-the-WKB-boundary in the converter is the remaining correct place.

## Technical Details

New `core/linearize.py` (stdlib only, ~250 lines): ISO WKB transformer mapping CIRCULARSTRING/COMPOUNDCURVE→LINESTRING, CURVEPOLYGON→POLYGON, MULTICURVE/MULTISURFACE→MULTILINESTRING/MULTIPOLYGON, recursing through GEOMETRYCOLLECTION; arcs sampled at ≤4°/segment (GDAL's `OGR_ARC_STEPSIZE` default), exact endpoints, linear Z/M interpolation, collinear arcs degrade to segments. Fallback wiring in `_read_spatial_to_arrow`; a per-file warning reports how many geometries were stroked. EWKB flags and the surface family (POLYHEDRALSURFACE/TIN/TRIANGLE) raise `LinearizeError`, which surfaces as #646's actionable message.

## Breaking Changes
**Does this PR introduce breaking changes?** No — inputs that previously hard-errored now convert; everything else is untouched.

## Related Issue(s)
- #643

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Curved geometries are now automatically converted to linear geometries during GeoParquet conversion.
  * Added configurable angular sampling to control curve approximation accuracy.
  * Added CLI and Python API options to enable or disable curve linearization.
  * Linearization supports complex curved geometry types and preserves dimensional coordinates where applicable.
  * Added clearer errors and guidance for unsupported or malformed geometries.

* **Documentation**
  * Documented curved geometry conversion behavior, configuration options, defaults, and compatibility details.

### DIFF
--- a/docs/guide/convert.md
+++ b/docs/guide/convert.md
@@ -395,6 +395,52 @@ Only invalid geometry is touched: valid geometry is byte-identical, NULL is
 preserved, and bounding boxes stay correct (`ST_MakeValid` never expands an
 envelope).
 
+### Curved Geometries
+
+GeoParquet cannot represent curved geometry types (CircularString,
+CompoundCurve, CurvePolygon, MultiCurve, MultiSurface — common in GeoPackage
+and FileGDB exports from CAD or ArcGIS), so by default gpio strokes arcs into
+line segments on read, the same operation as `ogr2ogr -nlt CONVERT_TO_LINEAR`
+or PostGIS `ST_CurveToLine`. A warning reports how many features were
+linearized.
+
+Note this **alters the geometry**: arcs become chains of straight segments,
+sampled every 4° of arc by default (GDAL's default; ~0.08% area error on a
+full circle). Earlier gpio versions failed on curved input — opt out with
+`--no-linearize-curves` to keep that strict behavior.
+
+=== "CLI"
+
+    ```bash
+    # Default: linearizes and warns ("Linearized 66 curved geometries ...")
+    gpio convert curved.gpkg output.parquet
+
+    # Denser sampling: 1 degree per segment
+    gpio convert curved.gpkg output.parquet --max-angle-deg 1
+
+    # Opt out: curved input fails with an actionable error
+    gpio convert curved.gpkg output.parquet --no-linearize-curves
+    ```
+
+=== "Python"
+
+    ```python
+    import geoparquet_io as gpio
+
+    # Default: linearizes curved geometries
+    gpio.convert("curved.gpkg").write("output.parquet")
+
+    # Denser sampling: 1 degree per segment
+    gpio.convert("curved.gpkg", max_angle_deg=1.0).write("output.parquet")
+
+    # Opt out: raise on curved input instead
+    gpio.convert("curved.gpkg", linearize_curves=False)
+    ```
+
+Linear geometries pass through untouched, and NULL is preserved. The surface
+family (PolyhedralSurface, TIN, Triangle) is not linearized and still raises
+an error.
+
 ### Skip Hilbert Ordering
 
 For faster conversion when spatial ordering isn't critical:
--- a/docs/guide/convert.md
+++ b/docs/guide/convert.md
@@ -437,9 +437,15 @@ full circle). Earlier gpio versions failed on curved input — opt out with
     gpio.convert("curved.gpkg", linearize_curves=False)
     ```
 
-Linear geometries pass through untouched, and NULL is preserved. The surface
-family (PolyhedralSurface, TIN, Triangle) is not linearized and still raises
-an error.
+Linear geometries pass through untouched, NULL is preserved, and empty curves
+become their empty linear counterpart. The surface family (PolyhedralSurface,
+TIN, Triangle) is not linearized and still raises an error.
+
+Curved input is spotted either by a header scan of local GeoPackages or on the
+first pass that parses geometry. `--skip-hilbert` removes that pass, so a
+curved source which is not a local `.gpkg` (FileGDB, a GeoPackage on S3) still
+errors under `--skip-hilbert` — drop the flag, or linearize with `ogr2ogr`
+first.
 
 ### Skip Hilbert Ordering
 
--- a/geoparquet_io/api/table.py
+++ b/geoparquet_io/api/table.py
@@ -287,6 +287,8 @@ def convert(
     profile: str | None = None,
     layer: str | None = None,
     repair_geometry: bool = True,
+    linearize_curves: bool = True,
+    max_angle_deg: float | None = None,
 ) -> Table:
     """
     Convert a geospatial file to a Table.
@@ -306,6 +308,11 @@ def convert(
         profile: AWS profile name for S3 authentication (default: None)
         layer: Layer name for multi-layer formats (GeoPackage, FileGDB). If not specified,
                reads the first/default layer.
+        linearize_curves: Stroke curved geometries (CircularString..MultiSurface) into
+               their linear equivalents when DuckDB cannot parse them (default: True,
+               mirroring repair_geometry). False raises an actionable error instead.
+        max_angle_deg: Maximum angular step per stroked arc segment in degrees
+               (default: 4.0, GDAL's OGR_ARC_STEPSIZE default).
 
     Returns:
         Table for chaining operations
@@ -331,6 +338,8 @@ def convert(
         geometry_column=geometry_column,
         layer=layer,
         repair_geometry=repair_geometry,
+        linearize_curves=linearize_curves,
+        max_angle_deg=max_angle_deg,
     )
 
     return Table(arrow_table, geometry_column=geom_col)
--- a/geoparquet_io/cli/decorators.py
+++ b/geoparquet_io/cli/decorators.py
@@ -193,6 +193,30 @@ def repair_geometry_option(func):
     )(func)
 
 
+def linearize_curves_options(func):
+    """
+    Add --linearize-curves/--no-linearize-curves and --max-angle-deg options.
+
+    Curved geometries (CircularString through MultiSurface) cannot be
+    represented in GeoParquet, so they are stroked into line segments by
+    default. Users who want curved input to fail instead can opt out.
+    """
+    func = click.option(
+        "--max-angle-deg",
+        type=click.FloatRange(min=0, min_open=True),
+        default=None,
+        help="Maximum degrees per stroked arc segment when linearizing curves (default: 4).",
+    )(func)
+    return click.option(
+        "--linearize-curves/--no-linearize-curves",
+        default=True,
+        help=(
+            "Stroke curved geometries (CircularString..MultiSurface) into line "
+            "segments (default: on). Use --no-linearize-curves to fail on curved input."
+        ),
+    )(func)
+
+
 def write_memory_option(func):
     """
     Add --write-memory option to a command.
--- a/geoparquet_io/cli/main.py
+++ b/geoparquet_io/cli/main.py
@@ -21,6 +21,7 @@ from geoparquet_io.cli.decorators import (
     grid_aggregate_options,
     handle_directory_sub_partition,
     handle_geoparquet_errors,
+    linearize_curves_options,
     metric_nodata_option,
     output_format_options,
     overwrite_option,
@@ -1264,6 +1265,7 @@ def convert(ctx):
     help="Allow conversion to plain Parquet when no geometry column is detected (default: error)",
 )
 @repair_geometry_option
+@linearize_curves_options
 @geoparquet_version_option
 @verbose_option
 @output_format_options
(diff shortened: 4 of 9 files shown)


cayetanobv:
Second independent real-world validation: CAL FIRE's Historic Fire Perimeters FileGDB (https://data.ca.gov/dataset/california-fire-perimeters-all, 23,334 features) — declared MultiPolygon, 91 hidden curved geometries — converts in ~13 s on this branch; hard-fails on main. Also exercises the non-GPKG (FileGDB) fallback path.

cayetanobv:
Third real-world validation — this one is the exact file from duckdb/duckdb-spatial#365 (GHUB2018.gpkg, Zenodo 5168383, CC-BY): four of six layers declare MultiPolygon but contain MULTISURFACE rows (72 in UE_2018 alone). The UE_2018 layer converts in ~15 s on this branch; the file was reported unreadable upstream and hard-fails on main. With SITEX CORINE (GPKG) and CAL FIRE (FileGDB), that's three independent government/academic datasets covered, plus PAD-US Kansas as a negative control where the fallback never triggers.

cholmes:
## Review: 2 blocking findings, 2 important, several minor

Reviewed in an isolated worktree, including a differential comparison against GDAL (`ogr.Geometry.GetLinearGeometry(4.0)` over hand-built WKB, plus end-to-end vs `ogr2ogr -nlt CONVERT_TO_LINEAR`). Everything below was verified by running code.

**First, the good news: the stroking math is genuinely sound.** Across 18 differential cases the max coordinate deviation from GDAL was 2.3e-13. Specifically verified correct: arc direction (including near-coincident endpoints where the midpoint disambiguates), multi-arc chaining, Z/M/ZM handling (type codes 1002/2002/3002, extras interpolated along the arc parameter, *not* mis-parsed as XY), EWKB flag rejection, big-endian input producing byte-identical output, COMPOUNDCURVE joint de-duplication, ring closure and winding preservation (CCW → +3.139041, CW → −3.139041; a donut's interior ring matches GDAL exactly), and the `max_angle_deg` budget honoured with no off-by-one (90°/4° → 23 segments of 3.913°; 180° → 46 points, exactly GDAL's count). The 0.08% accuracy claim checks out: 3.1390413184856385 vs π, **0.0812%**, and the analytic value `0.5·90·sin(4°)` agrees.

There is one correctness hole, and it's a bad one.

---

### 🔴 A closed CIRCULARSTRING silently collapses to a zero-area line

`core/linearize.py:113`:

```python
d = 2.0 * (x0*(y1-y2) + x1*(y2-y0) + x2*(y0-y1))
if abs(d) < 1e-12:
    return [p1, p2]   # collinear
```

When `p0 == p2` — the standard compact full-circle encoding — substituting `x2=x0, y2=y0` makes `d` **identically zero**. It's an algebraic identity, not a precision artifact, so it holds at every scale (confirmed at unit magnitude and at UTM magnitude). The collinear guard therefore always fires on a closed arc:

```
CURVEPOLYGON( CIRCULARSTRING(0 0, 2 0, 0 0) )   # circle, centre (1,0), r=1
  -> POLYGON ((0 0, 2 0, 0 0))
  -> ST_Area = 0.0   (true area = π)
  -> changed = True  # reports success
```

A consequence worth flagging: the `ccw_end == 0.0` branch at `:139`, commented *"closed circle described by three points"*, is **provably unreachable**. The case was anticipated — the guard above just wins first.

**It gets worse on the default path.** `repair_geometry` masks it rather than catching it:

```
before repair: POLYGON ((0 0, 2 0, 0 0))   ST_IsValid = False
after  repair: LINESTRING (0 0, 2 0)       ST_IsValid = True,  area 0.0
```

So through `gpio.convert()` the circle becomes a 2-point line, the geometry *type* silently changes, and it **passes validity checks** — no error, no warning, nothing in the invalid count. Measured against GDAL on the same input: 0.000000 vs 3.139097.

It also isn't limited to standalone 3-point strings — any mid-chain sub-arc where `pts[i] == pts[i+2]` collapses the same way, e.g. `CIRCULARSTRING(1 0, -1 0, 1 0, 0 -1, -1 0)`.

CI is green because `tests/data/curved_geometry_test.gpkg` uses the **5-point** encoding, which works correctly (91 pts, 0.081% error — the figure in the PR description). The 3-point form is never exercised, and the suite has no closed-arc case at all. Both encodings are standard and both are emitted by GDAL/PostGIS and CAD exports.

**Verified fix:** special-case `p0 == p2` *ahead* of the collinear guard — centre = midpoint of `p0`–`p1`, radius = half the distance, sweep = 2π (what OGR and PostGIS do). Gives 91 pts, area 3.139041, max radius error 1.1e-16, and zero regressions across all 18 GDAL cases. The now-dead `ccw_end == 0.0` branch can go with it.

---

### 🔴 Linearization never reaches the CLI

Same finding as on #646, and it applies here too:

```
$ gpio convert geoparquet tests/data/curved_geometry_test.gpkg out.parquet
Error: Conversion failed: Invalid Input Error: Unsupported geometry type in WKB
```

`read_spatial_to_arrow` has exactly one caller — `api/table.py:329`. `gpio convert` goes through `convert_to_geoparquet` → `_convert_spatial_path`, a separate SQL path. Compounding it:

- **Phantom parameters.** `convert_to_geoparquet`'s docstring (`core/convert.py:1579-1585`) documents `linearize_curves` and `max_angle_deg`; the signature (`:1526-1546`) has neither. That's the docstring a user reads right before hitting the error above.
- **Orphaned sentence.** The insertion left `repair_geometry`'s trailing line reading as if it belongs to `max_angle_deg` at `:1585`: *"When False, invalid geometry is preserved and a warning reports the count."*
- **Parity claim.** The PR says this mirrors `repair_geometry` — but that one has `cli/decorators.py:179 repair_geometry_option` wired into `cli/main.py:1266,1361,1383`. `linearize_curves` has no decorator and no flag.
- No `docs/guide/` changes; per CLAUDE.md new options need CLI + Python examples.

Also worth stating explicitly in the docs: `linearize_curves=True` by default means conversion now silently alters geometry where it previously raised #646's error. Defensible — it mirrors `repair_geometry` and does emit a warning — but it's a behaviour change that deserves a line in the guide.

---

### 🟡 `max_angle_deg` is unvalidated; a negative value silently corrupts

`core/linearize.py:141`, measured on a unit circle (true area π):

```
max_angle_deg=0       -> ZeroDivisionError: float division by zero
max_angle_deg=-4      ->       5 pts, area=2.000000   NO WARNING   # a square
max_angle_deg=0.0001  -> 3600001 pts, area=3.141593   NO WARNING
```

The negative case is the same failure class as the P1 above: silently wrong geometry with `changed=True`. The zero case matters too, because `_read_spatial_linearized` (`core/convert.py:1400`) catches only `LinearizeError`, so a bare `ZeroDivisionError` bypasses the actionable-error path and surfaces as *"Reading failed: float division by zero"*. Suggest raising `ValueError` for `<= 0` at the entry points (and `click.FloatRange` if a flag is added).

---

### 🟡 Layer-name escaping is bypassed on the fallback

`core/convert.py:1381`:

```python
opts = f", layer := '{layer}'" if layer else ""
```

Every other `ST_Read` call routes through `_build_st_read_expr` → `_validate_layer_name` (`:50`), which **escapes** single quotes by doubling them rather than rejecting them. This path skips that. With `layer = "Côte d'Ivoire"`:

```
normal path    : ST_Read('f.gpkg', layer := 'Côte d''Ivoire')          # OK
linearize path : ST_Read('f.gpkg', keep_wkb := true, layer := 'Côte d'Ivoire')
                 -> ParserException: syntax error at or near "Ivoire"
```

Not an injection hole — the `--`/`;`/`/*` blocklist still runs earlier — but it's a functional break on a legitimate layer name, on the curved path only. `_build_st_read_expr`'s own docstring also warns that `ST_Read` may *segfault* on a bad layer name, which makes skipping the sanitizer more than cosmetic. Cleanest fix is a `keep_wkb` flag on `_build_st_read_expr` so one helper builds every `ST_Read`.

---

### 🔵 Minor

- **Degenerate arcs emit duplicate vertices.** When `p0 == p1` or `p1 == p2`, `_stroke_arc` returns `[p1, p2]` (`:117,125`) after `:193` already appended `pts[0]`, so `CIRCULARSTRING(0 0, 0 0, 2 0)` → `[(0,0), (0,0), (2,0)]`. Distinct from the COMPOUNDCURVE joint handling, which is correct.
- **`contains_curved_wkb` (`:294`) is production-dead** — only `tests/test_linearize.py` calls it. Detection is done by matching the DuckDB error string instead. Either use it to pre-screen or drop it.
- **Performance.** Measured on 50k curved features (22MB GPKG) against the `ogr2ogr`-linearized equivalent: 4.5s / 660MB peak RSS vs 0.7s / 411MB — ~7× slower, +60% RSS. Attribution: `keep_wkb` read 0.1s, `to_pylist` 0.0s, `linearize_wkb` **4.0s**. So the cost is the pure-Python stroking loop, not the double read — worth knowing if this is ever optimized. Stroking also inflates WKB 5MB → 74MB (15×), which then flows into #645's `ST_IsValid` pass.
- **Fixture has no CRS.** `curved_geometry_test.gpkg` has `srs_id 0`, so there's zero coverage of curved + projected. The CRS-survives-the-fallback property does hold (I verified it end-to-end with an EPSG:6933 fixture), but it holds because of *where* detection sits, not because anything asserts it — and #644 exists precisely because that class of property regressed silently before. Worth adding a projected fixture.
- **Design note, explicitly not a blocker.** `core/convert.py` now has two hooks branching on the same DuckDB error string, and they already collide (the double-wrap noted on #646). The alternative is half-built already: `curved_geometry.py` can detect curves *without* provoking an error. A `_choose_read_strategy(input_url, layer) -> normal | linearized | error` called at `:1192` would remove the wasted failing query, make strategy unit-testable, and give `convert_to_geoparquet` one place to opt in — which is what closes the CLI finding above. I wouldn't re-architect under a bugfix PR, but I'd avoid adding a third error-string hook.

---

**Verified clean:** read-path divergence (both paths produce identical column order, names, types and bytes); repair ordering (linearize → repair, and stroked arcs aren't flagged invalid even at 170°/segment); full pre-push hook set on the merged tree (28/28, including import-linter and xenon); all four PRs merge onto `main` without conflict.

cayetanobv:
Thanks for the thorough review — the closed-circle catch especially. All addressed in ff4a3d8:

- **Closed circles**: `p2 == p0` is handled ahead of the circumcenter determinant (centre = p0–p1 midpoint, full 2π sweep). Tests cover the 3-point form, mid-chain closed sub-arcs, and CURVEPOLYGON rings — area ≈ π now. The dead `ccw_end == 0.0` branch is gone.
- **CLI**: took your `_choose_read_strategy` suggestion. Local GPKGs are pre-scanned with the stdlib scanner (no more wasted failing query), and `gpio convert` reads curved input through a linearized view — bounds/bbox/Hilbert included. Added `--linearize-curves/--no-linearize-curves` and `--max-angle-deg`, fixed the docstring, and added a guide section with the behavior-change note. Formats without a cheap scan (FileGDB) still go through the error-triggered fallback.
- **max_angle_deg**: rejected at every entry point when not > 0 (ValueError / InvalidParameterError / FloatRange on the flag).
- **Layer escaping**: keep_wkb reads now build through `_build_st_read_expr`.
- Minors: no duplicate vertices from degenerate arcs; `contains_curved_wkb` now prescreens rows so linear rows keep their bytes and skip the Python loop (that was the 4s hotspot); projected-CRS detection pinned at the read level (the write-side assert lands with #644); #646's double-wrap fixed by re-raising `GeoParquetError` untouched.

Re-validated on the three real datasets from the issue thread (Extremadura CORINE GPKG, CalFire FileGDB, GHUB2018 11GB GPKG): row counts and total areas identical to the pre-review outputs, and the prescreen cuts read time 30–60% on them — they're mostly linear rows. The pre-scan on the 11GB GPKG is layer-scoped and takes 0.1s.

Issue #649, opened 2026-08-20
gpio convert fails on any source containing an empty geometry (ST_Hilbert)


cayetanobv:
## Description

`gpio convert` dies on any input that contains an empty geometry, because the Hilbert `ORDER BY` in `_build_conversion_query` calls `ST_Hilbert` on every row:

```
Error: Conversion failed: Invalid Input Error: ST_Hilbert(geom, bounds) does not support empty geometries
```

`gpio sort hilbert` already handles this case correctly — it partitions empty/null geometries out and appends them at the end (`core/hilbert_order.py:131-150`), warning `Found N empty/null geometries. These will be placed at the end of the sorted output.` The convert path never got the same treatment, so the two commands disagree about the same data.

NULL geometries are fine; only *empty* ones (`POLYGON EMPTY`, `LINESTRING EMPTY`, …) trigger it.

## Reproduction

```bash
python -c "
import duckdb
con = duckdb.connect(); con.execute('INSTALL spatial; LOAD spatial;')
con.execute(\"\"\"
COPY (SELECT 1 AS id, ST_GeomFromText('POINT(1 1)') AS geom
      UNION ALL SELECT 2, ST_GeomFromText('POLYGON EMPTY'))
TO 'empties.gpkg' WITH (FORMAT GDAL, DRIVER 'GPKG')
\"\"\")"

gpio convert empties.gpkg out.parquet
# Error: Conversion failed: Invalid Input Error: ST_Hilbert(geom, bounds) does not support empty geometries

gpio convert empties.gpkg out.parquet --skip-hilbert   # works
gpio sort hilbert out.parquet sorted.parquet           # works, warns about the empty row
```

Empty geometries are common in real GeoPackage/Shapefile/FileGDB exports, so this makes `gpio convert` unusable on those files unless the user knows to pass `--skip-hilbert` (which costs the spatial ordering). It also surfaces on curved sources now that #647 linearizes `CIRCULARSTRING EMPTY` into `LINESTRING EMPTY`.

## Proposed fix

Mirror what `hilbert_order.py` already does, in `_build_conversion_query` (`core/convert.py:920-931`): order the non-empty rows and append the empty/null ones, rather than ordering the whole set.

```sql
WITH ordered AS (
    <base_select> WHERE geom IS NOT NULL AND NOT ST_IsEmpty(geom)
    ORDER BY ST_Hilbert(geom, bounds)
),
unordered AS (
[...]

PR #650, opened 2026-08-20
Stream the linearized curved-geometry read instead of materializing the file

cayetanobv:
## Description

Follow-up to the review on #647, closing the two findings that were left out of scope there.

**1. The linearized read held the whole file in Python.** `_linearize_to_arrow` materialized the entire `keep_wkb` result, then built a Python list of every blob, then rebuilt the column — peak memory of roughly 2–3× the dataset's WKB, and it turned the otherwise-streaming `gpio convert` into an in-memory operation for curved input only. `pa.array(values, type=pa.binary())` also capped the geometry column at 2 GB of blobs regardless of what Arrow type the source actually used, so a large curved dataset raised a capacity error rather than converting.

`_LinearizedRead` now strokes one record batch at a time and keeps the geometry column's source type (`large_binary` stays `large_binary`). The query-based path (`_register_linearized_view`) inserts each batch into a temp relation as it arrives, so DuckDB owns the rows — and can spill them to disk — instead of Python holding the file. The API path still returns one Arrow table, as its contract requires, but without the extra full-column Python copy.

One wrinkle worth flagging for review: the read runs on its own `con.cursor()`. A DuckDB streaming result is invalidated by the next statement on its connection, and the consumer issues plenty (`DROP TABLE`, `register`, `CREATE TEMP TABLE`, `INSERT`) while pulling batches. Only the raw `keep_wkb` read happens on that cursor, so the connection-level spatial settings a cursor does not inherit (axis order) cannot affect it.

**2. The warning over-claimed.** A curved *type* containing no arcs — an empty `CIRCULARSTRING`, a `MULTISURFACE` of plain polygons — is rewritten to a linear type but strokes nothing, while the old message reported it as `Linearized N curved geometries (arcs stroked at <= 4.0 degrees per segment)`. `linearize_wkb_stats` now returns the arc count alongside `changed`, and the message reports both:

```
Linearized 1 curved geometry (2 arcs stroked at <= 4.0 degrees per segment); GeoParquet cannot represent curves.
```

`linearize_wkb` keeps its two-value contract for existing callers.

## Technical Details

- `core/linearize.py`: `_Linearizer.arcs` counts arcs actually sampled; new `linearize_wkb_stats()` returns `(wkb, changed, arcs)`.
- `core/convert.py`: `_linearize_to_arrow` → `_LinearizedRead` (batch generator, source Arrow type preserved, own cursor); `_register_linearized_view` streams batches into `_gpio_linearized`, including the no-batch and empty-batch cases; `_read_spatial_linearized` builds its table from the same batches.
- Tests: arc-count stats (arcs present, chained arcs, curved-type-without-arcs), batch generation, Arrow type preservation for `binary`/`large_binary`, a multi-batch source landing wholly in one relation, an empty source still registering a usable relation, and the warning wording.
- Also adds the end-to-end **curved + projected write** test review asked for on #644: a source taking the linearize path must write its real CRS (EPSG:25830), not fall back to OGC:CRS84. The read-level half already landed in #647.

Local: 818 passed / 7 skipped on the convert/crs/repair/linearize/curved/hilbert/geopackage selection; pre-commit, xenon, and import-linter clean.

## Breaking Changes
**Does this PR introduce breaking changes?** No — same outputs, less memory, an extra number in one warning.

## Related Issue(s)
- #643, follow-up to #647
- Unrelated but adjacent: #649 (`ST_Hilbert` rejects empty geometries, so `gpio convert` still fails on any source containing one)

## Checklist

## Summary by CodeRabbit

* **Improvements**
  * Curved-geometry conversion now processes large datasets in smaller batches, reducing memory usage.
  * Preserves binary geometry formats and coordinate reference systems during conversion.
  * Provides more detailed conversion statistics, including the number of linearized arcs.
  * Improved support for multi-batch, empty, and zero-record data sources.
* **Bug Fixes**
  * Improved reliability, validation, and error reporting for curved-geometry conversion.

### DIFF
--- a/geoparquet_io/core/convert.py
+++ b/geoparquet_io/core/convert.py
@@ -1502,6 +1502,17 @@ def _read_spatial_to_arrow(
         return _read_spatial_linearized(con, input_url, layer, geom_column, max_angle_deg)
 
 
+#: Rows per batch for the linearized read. DuckDB's ``.arrow()`` defaults to
+#: 1,000,000, which hands back most files as a single batch and defeats the
+#: point of streaming. Stroking multiplies a geometry's size (an arc becomes
+#: ~90 vertices), so the batch has to be small: on a 50k-row curved GeoPackage,
+#: peak RSS was 503 MB at the default and 100k rows, 395 MB at 20k, 370 MB at
+#: 5k, with no measurable difference in wall time. The repo's generic
+#: ``DEFAULT_BATCH_SIZE`` (100k, write_strategies/arrow_streaming.py) is too
+#: coarse here for that reason.
+_LINEARIZE_BATCH_ROWS = 20_000
+
+
 class _LinearizedRead:
     """A ``keep_wkb`` read whose curved WKB is stroked batch by batch.
 
@@ -1535,12 +1546,16 @@ class _LinearizedRead:
         # connection, and the caller runs plenty (the batches are inserted into
         # a temp relation as they arrive), so the read gets its own cursor. The
         # cursor is kept alive on self: dropping it would close the reader.
-        # Only the raw keep_wkb read runs here, so the connection-level spatial
-        # settings a cursor does not inherit (axis order) cannot affect it.
+        #
+        # GLOBAL settings survive the cursor — including arrow_large_buffer_size,
+        # which is what makes DuckDB hand back large_binary blobs and lets this
+        # path exceed the 2 GB a pa.binary() column can address. Connection-level
+        # spatial settings (axis order) would not carry over, but the raw
+        # keep_wkb read does not consult them.
         self._cursor = self.con.cursor()
         return self._cursor.execute(
             f"SELECT * FROM {_build_st_read_expr(self.input_url, self.layer, keep_wkb=True)}"
-        ).arrow()
+        ).arrow(rows_per_batch=_LINEARIZE_BATCH_ROWS)
 
     def batches(self):
         """Yield the source's record batches with curved WKB stroked in place."""
--- a/geoparquet_io/core/convert.py
+++ b/geoparquet_io/core/convert.py
@@ -1502,74 +1502,106 @@ def _read_spatial_to_arrow(
         return _read_spatial_linearized(con, input_url, layer, geom_column, max_angle_deg)
 
 
-def _linearize_to_arrow(con, input_url, layer, geom_column, max_angle_deg=None):
-    """Read a spatial file whose WKB contains curved types, stroking them.
+class _LinearizedRead:
+    """A ``keep_wkb`` read whose curved WKB is stroked batch by batch.
 
     ``ST_Read(keep_wkb := true)`` hands back the raw WKB blobs untouched (the
-    escape hatch DuckDB documents for geometry subtypes it cannot represent);
-    each curved blob is stroked into its linear equivalent in Python. Returns
-    ``(raw_table, wkb_col)`` with the geometry column holding linear WKB
-    under its original name.
+    escape hatch DuckDB documents for geometry subtypes it cannot represent),
+    and each curved blob is stroked into its linear equivalent in Python.
+    Working per record batch keeps one chunk of geometry live at a time instead
+    of the whole file (previously the read was materialized, then copied into a
+    Python list of every blob, then rebuilt), and the geometry column keeps its
+    source Arrow type — ``pa.binary()`` caps a column at 2 GB of blobs, which a
+    large curved dataset can exceed.
     """
-    from geoparquet_io.core.curved_geometry import unsupported_wkb_error_message
-    from geoparquet_io.core.linearize import (
-        DEFAULT_MAX_ANGLE_DEG,
-        LinearizeError,
-        contains_curved_wkb,
-        linearize_wkb,
-    )
 
-    if max_angle_deg is None:
-        max_angle_deg = DEFAULT_MAX_ANGLE_DEG
+    def __init__(self, con, input_url, layer, geom_column, max_angle_deg=None):
+        from geoparquet_io.core.linearize import DEFAULT_MAX_ANGLE_DEG
 
-    raw = (
-        con.execute(f"SELECT * FROM {_build_st_read_expr(input_url, layer, keep_wkb=True)}")
-        .arrow()
-        .read_all()
-    )
+        self.con = con
+        self.input_url = input_url
+        self.layer = layer
+        self.max_angle_deg = DEFAULT_MAX_ANGLE_DEG if max_angle_deg is None else max_angle_deg
+        self._reader = self._open_reader()
+        self.schema = self._reader.schema
+        self.wkb_col = geom_column if geom_column in self.schema.names else "wkb_geometry"
+        if self.wkb_col not in self.schema.names:
+            raise GeoParquetError(f"keep_wkb read of {input_url} exposes no '{geom_column}' column")
+        self.linearized = 0
+        self.arcs = 0
 
-    wkb_col = geom_column if geom_column in raw.column_names else "wkb_geometry"
-    if wkb_col not in raw.column_names:
-        raise GeoParquetError(f"keep_wkb read of {input_url} exposes no '{geom_column}' column")
+    def _open_reader(self):
+        # A DuckDB streaming result is invalidated by the next statement on its
+        # connection, and the caller runs plenty (the batches are inserted into
+        # a temp relation as they arrive), so the read gets its own cursor. The
+        # cursor is kept alive on self: dropping it would close the reader.
+        # Only the raw keep_wkb read runs here, so the connection-level spatial
+        # settings a cursor does not inherit (axis order) cannot affect it.
+        self._cursor = self.con.cursor()
+        return self._cursor.execute(
+            f"SELECT * FROM {_build_st_read_expr(self.input_url, self.layer, keep_wkb=True)}"
+        ).arrow()
 
-    import pyarrow as pa
+    def batches(self):
+        """Yield the source's record batches with curved WKB stroked in place."""
+        import pyarrow as pa
 
-    changed = 0
-    values = []
-    for value in raw.column(wkb_col).to_pylist():
-        if value is None:
-            values.append(None)
-            continue
-        if not contains_curved_wkb(value):
-            values.append(value)  # linear at the top level: keep the bytes as-is
-            continue
+        idx = self.schema.get_field_index(self.wkb_col)
+        wkb_type = self.schema.field(idx).type
+        for batch in self._reader:
+            columns = list(batch.columns)
+            columns[idx] = pa.array(
+                [self._linearize(value) for value in batch.column(idx).to_pylist()],
+                type=wkb_type,
+            )
+            yield pa.RecordBatch.from_arrays(columns, schema=batch.schema)
+        self._report()
+
+    def _linearize(self, value):
+        from geoparquet_io.core.curved_geometry import unsupported_wkb_error_message
+        from geoparquet_io.core.linearize import (
+            LinearizeError,
+            contains_curved_wkb,
+            linearize_wkb_stats,
+        )
+
+        if value is None or not contains_curved_wkb(value):
+            return value  # NULL, or linear at the top level: keep the bytes as-is
         try:
-            linear, did_change = linearize_wkb(value, max_angle_deg)
+            linear, changed, arcs = linearize_wkb_stats(value, self.max_angle_deg)
         except LinearizeError as e:
             # e.g. the surface family (POLYHEDRALSURFACE/TIN/TRIANGLE) or
             # malformed blobs: fall back to the actionable error (#643).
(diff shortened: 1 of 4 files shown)


cayetanobv (commented):
**Recommendation: changes requested** (posting as a comment review -- GitHub blocks REQUEST_CHANGES on your own PR).

## Measured: the streaming doesn't stream

`DuckDBPyConnection.arrow()` defaults to `rows_per_batch=1_000_000`, so for any source under 1M rows the reader emits the whole result as **one batch**. On the real 12,001-row curved GPKG: `batches: 1, sizes: [12001]`.

Peak RSS through `_register_linearized_view`, 200,001-row curved GeoPackage:

| | peak RSS |
|---|---|
| `origin/main` | 1158 MB |
| this branch as written | **1174 MB** |
| this branch + `rows_per_batch=100_000` | 1142 MB |
| this branch + `rows_per_batch=20_000` | **661 MB** |

So the headline claim holds only above 1M rows, and even there the working set is 1M rows x (Arrow batch + `to_pylist()` Python bytes + rebuilt array). Fix is one argument -- see the inline comment on `_open_reader`.

## What checks out

- **`large_binary` preservation is a real fix.** `get_duckdb_connection` sets `arrow_large_buffer_size = true`, DuckDB emits `large_binary`, and the old `pa.array(..., type=pa.binary())` would have raised past 2 GB.
- **Cursor isolation works.** I streamed from `con.cursor()` while the parent connection ran interleaved `CREATE TEMP TABLE` / `INSERT`: all 12,001 rows landed, all POLYGON.
- **Arc counting is correct** across nesting: COMPOUNDCURVE -> 1, MULTICURVE of two -> 2, plain LINESTRING -> 0.
- **The no-batch fallback is genuinely needed.** DuckDB yields *zero* batches for a 0-row result (verified), and a real 0-row GPKG through the `not created` branch returns `id INTEGER, geom GEOMETRY` with 0 rows. Correct -- just covered by the wrong path, see inline.
- Merges cleanly with #651 in either order; merged tree passes 887 / 7 skipped on the convert/crs/hilbert/sort/csv/geopackage/curved/linearize/repair selection.

## Recommendation

Hold for the `rows_per_batch` fix. Without it this adds a class, a cursor, and a streaming protocol while leaving memory where it was; with it, 200k rows goes 1174 MB -> 661 MB. The `large_binary` fix and the honest arc count are worth landing either way.

geoparquet_io/core/convert.py
```diff
@@ -1502,74 +1502,106 @@ def _read_spatial_to_arrow(
...
+        # A DuckDB streaming result is invalidated by the next statement on its
+        # connection, and the caller runs plenty (the batches are inserted into
+        # a temp relation as they arrive), so the read gets its own cursor. The
+        # cursor is kept alive on self: dropping it would close the reader.
+        # Only the raw keep_wkb read runs here, so the connection-level spatial
+        # settings a cursor does not inherit (axis order) cannot affect it.
+        self._cursor = self.con.cursor()
+        return self._cursor.execute(
+            f"SELECT * FROM {_build_st_read_expr(self.input_url, self.layer, keep_wkb=True)}"
+        ).arrow()
```
cayetanobv: This is where the memory claim is lost. `.arrow()` with no argument means `rows_per_batch=1_000_000`:

```
arrow(self, rows_per_batch: int = 1000000) -> pyarrow.lib.RecordBatchReader
```

So for any source under 1M rows this yields exactly one batch containing the whole file, and `batches()` does the same full-column `to_pylist()` + rebuild that the old `_linearize_to_arrow` did. Measured on a 200,001-row curved GPKG: main 1158 MB, this branch 1174 MB, `rows_per_batch=20_000` 661 MB.

The repo already has the convention for this -- `write_strategies/arrow_streaming.py:27` defines `DEFAULT_BATCH_SIZE = 100_000` and passes `rows_per_batch=batch_size` at lines 160 and 774. Worth noting 100k barely moved the needle in my measurement (1142 MB); stroked geometry expands a lot, so something nearer 20k looks right for this path specifically.

geoparquet_io/core/convert.py
```diff
@@ -1502,74 +1502,106 @@ def _read_spatial_to_arrow(
...
+        self.linearized = 0
+        self.arcs = 0
+
+    def _open_reader(self):
+        # A DuckDB streaming result is invalidated by the next statement on its
+        # connection, and the caller runs plenty (the batches are inserted into
+        # a temp relation as they arrive), so the read gets its own cursor. The
+        # cursor is kept alive on self: dropping it would close the reader.
+        # Only the raw keep_wkb read runs here, so the connection-level spatial
+        # settings a cursor does not inherit (axis order) cannot affect it.
```
cayetanobv: One correction for this comment, and it matters more than the axis-order point it makes: `arrow_large_buffer_size` is **GLOBAL** scope in DuckDB 1.5.1, so the cursor *does* inherit it.

```
SELECT name, scope FROM duckdb_settings() WHERE name='arrow_large_buffer_size'
-> ('arrow_large_buffer_size', 'GLOBAL', 'true')
```

That's load-bearing for this PR: it's exactly why the reader hands back `large_binary` and the `pa.binary()` 2 GB cap goes away. Since the comment is already justifying what the cursor can and can't see, it's worth stating that one too -- a future reader who assumes the cursor inherits nothing would conclude the `large_binary` fix can't work.

tests/test_linearize.py
```diff
@@ -618,3 +639,133 @@ def test_cli_error_names_the_remedy_when_it_cannot_linearize(self, tmp_path):
...
+        read.layer = None
+        read.max_angle_deg = 4.0
+        read.wkb_col = "geom"
+        read.schema = batches[0].schema if batches else pa.schema([("id", pa.int64())])
+        read.linearized = 0
+        read.arcs = 0
+        read._reader = iter(batches)
+        return read
+
+    def test_batches_are_generated_not_materialized(self):
```
cayetanobv: This assertion can't fail. `inspect.isgenerator()` on the result of a generator function is true regardless of what `self._reader` does -- it passes today *while the real reader hands back the whole file in one batch*. It reads as coverage for the PR's central claim and isn't.

What would have caught the `rows_per_batch` issue: read a real source and assert more than one batch, or assert the batch size the reader was configured with. With the fixture below at ~2 rows either is cheap; `_curved_batch(rows=...)` plus a bounded `rows_per_batch` gives you a deterministic batch count.

tests/test_linearize.py
```diff
@@ -618,3 +639,133 @@ def test_cli_error_names_the_remedy_when_it_cannot_linearize(self, tmp_path):
...
+            con, "fake.gdb", None, "geom", read=self._read(con, batches)
+        )
+
+        rows, kinds = con.execute(
+            f"SELECT count(*), list_distinct(list(ST_GeometryType(geom))) FROM {name}"
+        ).fetchone()
+        assert rows == 6
+        assert kinds == ["POLYGON"]
+
+    def test_empty_source_still_registers_a_relation(self):
```
cayetanobv: This exercises the `created` path, not the branch it's named for. `self._read(con, [empty])` supplies one 0-row batch, so the loop runs once and `CREATE TEMP TABLE` fires inside it -- `if not created:` is never reached.

The branch is real, though: DuckDB yields **zero** batches for a 0-row result, so a genuine 0-row curved source does hit it. I confirmed it behaves correctly (relation comes back `id INTEGER, geom GEOMETRY`, 0 rows). To cover it as written, pass `[]` rather than `[empty]` -- which also needs `read.schema` set, since `self._read` falls back to an `id`-only schema when `batches` is empty and `wkb_col` would then be missing.

geoparquet_io/core/linearize.py
```diff
@@ -327,7 +335,21 @@ def linearize_wkb(
         lin._write_geometry(_Reader(bytes(wkb)), out)
     except (struct.error, IndexError) as e:
         raise LinearizeError(f"Malformed WKB: {e}") from e
-    return bytes(out), lin.changed
+    return bytes(out), lin.changed, lin.arcs
+
+
+def linearize_wkb(
```
cayetanobv: After this PR `linearize_wkb` has no production callers -- `grep` finds it only in `tests/test_linearize.py`. The docstring's "for existing callers" framing doesn't match the tree any more.

Either drop it and move the tests onto `linearize_wkb_stats`, or keep it and say it's a deliberate public surface for API consumers. Worth deciding now rather than leaving an ambiguous wrapper; `vulture` won't flag it while the tests keep it live.

cayetanobv:
All five findings verified and fixed in 8c35a5b.

**`rows_per_batch` — confirmed and fixed.** Reproduced the single-batch behavior directly: `.arrow()` on a 12,001-row result gives `batches=1`, and only a bound below the row count splits it. Independently measured on a 50,001-row curved GeoPackage through `_register_linearized_view`:

| batch size | peak RSS | wall |
|---|---|---|
| default (1M) | 503 MB | 4.34 s |
| 100k | 503 MB | — |
| 20k | 395 MB | 4.19 s |
| 5k | 370 MB | 4.20 s |

Same shape as your 200k numbers, and no throughput cost, so `_LINEARIZE_BATCH_ROWS = 20_000`. The real path now measures **424 MB where it measured 503 MB**. The constant carries the measurement and says why the repo's generic 100k is too coarse here.

**Cursor comment — confirmed backwards, rewritten.** `duckdb_settings()` reports `arrow_large_buffer_size` as `GLOBAL`, and a cursor's blob column does come back `large_binary`. The comment now leads with that (it is what lifts the 2 GB ceiling) and keeps the narrower point that the raw `keep_wkb` read does not consult connection-scoped spatial settings.

**Vacuous assertion — replaced.** `test_real_read_is_split_into_batches` reads the real 42-row GeoPackage with `_LINEARIZE_BATCH_ROWS` monkeypatched to 10 and asserts more than one batch, `max(sizes) <= 10`, and all 42 rows surviving. It fails against the unbounded reader, which the old test could not.

**Empty-source test — confirmed mis-covered.** Now passes `[]` (what DuckDB actually yields for 0 rows) with an explicit schema, and asserts the relation comes back `GEOMETRY`-typed; the one-empty-batch case kept its own test, since it exercises the other branch.

**`linearize_wkb` — confirmed callerless, removed.** No production callers and no docs references, so the wrapper is gone; the tests go through `linearize_wkb_stats` via a local shim.

769 passed / 7 skipped on the convert/crs/curved/linearize/repair/geopackage selection; pre-commit and xenon clean.

cayetanobv (commented):
## Re-review: all five findings addressed, and I re-measured the one that mattered

**Recommendation: approve.** (Comment review -- GitHub blocks APPROVE on your own PR.)

### The memory fix is real this time

Same 200,001-row curved GeoPackage through `_register_linearized_view` as last round:

| | peak RSS |
|---|---|
| `origin/main` | 1158.6 MB |
| previous revision of this PR | 1174.5 MB |
| **this revision** | **639.7 MB** |

45% under `main`, where the previous revision was level with it.

### The new test actually guards it

I mutated `_open_reader` back to a bare `.arrow()` and re-ran the class:

```
FAILED tests/test_linearize.py::TestStreamingLinearizedRead::test_real_read_is_split_into_batches
E   AssertionError: reader was not batched: [42]
```

That is exactly the regression that slipped through last time, now caught. Reading a real source with `_LINEARIZE_BATCH_ROWS` monkeypatched down is the right shape for this -- it pins the constant is *used* rather than that it exists.

### The documented batch-size rationale holds up

Spot-checked the numbers in the `_LINEARIZE_BATCH_ROWS` comment on a 50k-row curved fixture:

| `rows_per_batch` | peak RSS (mine) | comment says |
|---|---|---|
| default (1M) | 510.9 MB | 503 MB |
| 100_000 | 580.4 MB | 503 MB |
| 20_000 | 401.1 MB | 395 MB |
| 5_000 | 360.2 MB | 370 MB |

Close enough that the reasoning is sound, and the conclusion -- that the generic `DEFAULT_BATCH_SIZE` of 100k is too coarse for a path where stroking multiplies geometry size -- is borne out. My 100k reading came in higher than yours rather than equal to the default; allocator noise, doesn't change the choice of 20k.

### Also confirmed

- `arrow_large_buffer_size` comment now says the right thing (GLOBAL, survives the cursor) -- and that's the load-bearing fact for the `large_binary` path, so good that it's stated.
- `linearize_wkb` deleted; the `_stroke` helper keeps the two-value call sites readable without a production wrapper, and the `contains_curved_wkb` docstring cross-reference was updated with it. Nice catch on that last one, I'd missed it.
- `test_source_yielding_no_batches_still_registers_a_relation` now passes `[]` and asserts the column comes back `GEOMETRY`, so it covers the branch it names; the one-empty-batch case kept as its own test.
- 867 passed / 7 skipped on the convert/csv/hilbert/crs/curved/linearize/sort/geopackage selection (run on #656, which stacks this).

The only thing I raised that isn't addressed is the cursor not being closed if `batches()` is abandoned mid-iteration on a `GeoParquetError`. I flagged it as minor and it still is -- GC handles it. Fine to leave.

PR #651, opened 2026-08-20
Order empty geometries last in convert instead of failing on ST_Hilbert

### LINKED ISSUE #649: gpio convert fails on any source containing an empty geometry (ST_Hilbert)

cayetanobv:
## Description

`ST_Hilbert` rejects empty geometries, and `gpio convert` keyed every row with it, so a single `POLYGON EMPTY` anywhere in the source failed the entire conversion:

```
Error: Conversion failed: Invalid Input Error: ST_Hilbert(geom, bounds) does not support empty geometries
```

`gpio sort hilbert` never had the problem — it orders the non-empty rows and appends the rest (`core/hilbert_order.py`) — so the two commands disagreed about the same data, and `--skip-hilbert` (giving up the spatial ordering) was the only way through. Empty geometries are common in GeoPackage/Shapefile/FileGDB exports, and #647's linearization produces them too (`CIRCULARSTRING EMPTY` becomes `LINESTRING EMPTY`).

Both conversion queries — the spatial/parquet path and the CSV/WKT path — now sort on an "unorderable" flag first, which lands empty geometry exactly where NULL geometry already landed under DuckDB's `NULLS LAST` default, and the geometry handed to `ST_Hilbert` is substituted with a corner of the dataset envelope for those rows so it is never asked to key an empty one.

## Technical Details

- New `_orderable_geom()` builds the substitution; `_build_conversion_query` and `_build_csv_conversion_query` order by `(unorderable), ST_Hilbert(...)`.
- **Deliberately not the CTE + `UNION ALL` shape** the issue sketched, for two reasons: it would duplicate the generated base select and scan the source twice (expensive for remote files and for #647's linearized temp relation), and ordering inside a branch of a `UNION ALL` is not guaranteed to survive it. A single `ORDER BY` has unambiguous semantics. Verified both shapes produce identical row order on a mixed empty/NULL/valid fixture.
- Only the *substitution* is conditional; `ST_Hilbert` is still evaluated for every row, so no spatial function runs under conditional execution — the shape behind the segfaults in #642/#645.
- The native-encoding branch keys on centroid expressions, which are NULL (not an error) for a geometry with no coordinates; it gets the same flag so placement is explicit rather than relying on the NULLS LAST default.
- A source whose geometry is *entirely* empty or NULL has no envelope to order by and used to fail with the unhelpful "Could not calculate dataset bounds". `_calculate_bounds(required=False)` now returns None there and the conversion proceeds unordered with a warning. Note this also means a **zero-row source converts instead of erroring** — a deliberate behavior change, and the more useful outcome, but worth a reviewer's eye.
- No warning for the ordinary mixed case: the rows are reordered, not dropped, and NULL geometry has always been placed last silently. Emitting one would also cost an extra count pass over the source, which `sort hilbert` can afford (in-memory table) and `convert` cannot.

**Row order for sources without empty geometry is unchanged** — verified by converting `tests/data/buildings_test.gpkg` on this branch and on `main` in a separate worktree and comparing the full centroid sequence (identical).

## Tests

New `tests/test_convert_empty_geometries.py`: one empty geometry no longer fails; empty *and* NULL sort last while the non-empty rows keep Hilbert order; all-empty converts unordered; native GeoArrow encoding; `--skip-hilbert` still fine; CSV with an empty WKT value; and a curved source carrying an empty curve (the #647 interaction). All written before the fix and failing without it.

Local: 856 passed / 7 skipped on the convert/crs/hilbert/sort/csv/geopackage/curved/linearize selection; pre-commit, xenon, and import-linter clean.

## Breaking Changes
**Does this PR introduce breaking changes?** No, unless "a zero-row input now converts rather than erroring" counts.

## Related Issue(s)
- Closes #649
- Touches the same file as #650 (streaming linearized read) but different functions; they should merge in either order.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Conversions now succeed when datasets contain empty or missing geometries.
  - Hilbert ordering safely places unorderable geometries last and skips ordering when bounds are unavailable.
  - CSV, native GeoArrow, and curved geometry conversions now handle empty geometries while preserving row counts.
  - Existing options to skip Hilbert ordering continue to work.

### DIFF
--- a/geoparquet_io/core/convert.py
+++ b/geoparquet_io/core/convert.py
@@ -233,20 +233,13 @@ def detect_all_geometry_columns(input_file: str, verbose: bool = False) -> dict:
 
 
 def _calculate_bounds(
-    con,
-    input_file,
-    geom_column,
-    verbose,
-    is_parquet=False,
-    encoding="WKB",
-    table_expr=None,
-    required=True,
+    con, input_file, geom_column, verbose, is_parquet=False, encoding="WKB", table_expr=None
 ):
     """Calculate dataset bounds from input file (or an explicit table_expr).
 
-    With ``required=False`` an unmeasurable dataset — no rows, or every
-    geometry empty or NULL — returns None instead of raising, letting the
-    caller convert without spatial ordering (issue #649).
+    Returns None for a dataset with nothing to measure — no rows, or every
+    geometry empty or NULL — leaving the caller to convert without spatial
+    ordering rather than fail (issue #649).
     """
     if verbose:
         debug("Calculating dataset bounds...")
@@ -289,9 +282,7 @@ def _calculate_bounds(
     bounds_result = con.execute(bounds_query).fetchone()
 
     if not bounds_result or any(v is None for v in bounds_result):
-        if not required:
-            return None
-        raise GeoParquetError("Could not calculate dataset bounds")
+        return None
 
     if verbose:
         xmin, ymin, xmax, ymax = bounds_result
@@ -771,7 +762,12 @@ def _get_geom_expr_and_where(geom_info, skip_invalid):
 
 
 def _calculate_csv_bounds(con, geom_info, skip_invalid, verbose):
-    """Calculate dataset bounds from CSV geometry."""
+    """Calculate dataset bounds from CSV geometry.
+
+    Returns None when there is nothing to measure, mirroring
+    :func:`_calculate_bounds`: an all-empty or row-less CSV converts unordered
+    instead of failing (issue #649).
+    """
     if verbose:
         debug("Calculating dataset bounds from CSV...")
 
@@ -799,7 +795,7 @@ def _calculate_csv_bounds(con, geom_info, skip_invalid, verbose):
         raise GeoParquetError(msg) from e
 
     if not bounds_result or any(v is None for v in bounds_result):
-        raise GeoParquetError("Could not calculate dataset bounds from CSV")
+        return None  # nothing to measure: caller writes unordered (#649)
 
     if verbose:
         xmin, ymin, xmax, ymax = bounds_result
@@ -829,20 +825,13 @@ def _build_plain_select_query(input_file, is_parquet=False, is_csv=False, delimi
     return f"SELECT * FROM ST_Read('{input_file}')"
 
 
-def _bounds_with_curve_fallback_checked(con, *args, **kwargs):
-    """``_bounds_with_curve_fallback`` plus the unmeasurable-dataset case.
-
-    Every geometry empty or NULL (or no rows at all) leaves nothing to build a
-    Hilbert envelope from, so the conversion proceeds unordered with a warning
-    instead of failing on bounds (issue #649).
-    """
-    bounds, table_expr = _bounds_with_curve_fallback(con, *args, **kwargs)
-    if bounds is None:
-        warn(
-            "No measurable geometry bounds (every geometry is empty or NULL); "
-            "writing without Hilbert ordering."
-        )
-    return bounds, table_expr
+#: Warned when Hilbert ordering is skipped for want of an envelope (#649). Both
+#: causes land here — a source with no rows at all, and one where every geometry
+#: is empty or NULL — so the wording has to cover both.
+_NO_BOUNDS_WARNING = (
+    "No geometry to measure (no rows, or every geometry empty or NULL); "
+    "writing without Hilbert ordering."
+)
 
 
 def _orderable_geom(geom_expr, xmin, ymin):
@@ -856,9 +845,12 @@ def _orderable_geom(geom_expr, xmin, ymin):
     which pins those rows last exactly where NULL geometry already landed
     under DuckDB's NULLS LAST default.
 
-    Only the substitution is conditional — ``ST_Hilbert`` itself is evaluated
-    for every row, keeping spatial functions out of conditional execution
-    (the shape behind the segfaults in #642).
+    ``ST_Hilbert`` itself is evaluated for every row rather than from inside a
+    branch. The substitution does put ``ST_Point`` in a THEN branch, but with
+    constant arguments and over an already-decoded GEOMETRY, not the raw-blob
+    ``IS NULL`` shape that misaligned the selection vector in #642 (see
+    ``core/geometry_repair.py``); a 200k-row source with NULL and empty rows
+    goes through both the ST_Read and parquet paths without incident.
     """
     return (
         f"CASE WHEN {geom_expr} IS NULL OR ST_IsEmpty({geom_expr}) "
@@ -1056,6 +1048,10 @@ def _convert_csv_path(
                 msg = "Reading CSV, creating geometries, and applying Hilbert ordering..."
         debug(msg)
 
+    if not effective_skip_hilbert and bounds is None:
+        warn(_NO_BOUNDS_WARNING)
+        effective_skip_hilbert = True
+
     query = _build_csv_conversion_query(
         geom_info, effective_skip_hilbert, bounds, skip_invalid, skip_bbox=skip_bbox
     )
@@ -1097,9 +1093,7 @@ def _bounds_with_curve_fallback(
         tuple: (bounds, table_expr) — table_expr is the linearized view when
         the fallback fired, otherwise the caller's value unchanged.
     """
-    # required=False: an all-empty or empty dataset has no envelope to order
-    # by, which the caller turns into an unordered write rather than an error.
-    kwargs = {"is_parquet": is_parquet, "encoding": encoding, "required": False}
+    kwargs = {"is_parquet": is_parquet, "encoding": encoding}
     try:
         bounds = _calculate_bounds(
             con, input_file, geom_column, verbose, table_expr=table_expr, **kwargs
@@ -1205,7 +1199,7 @@ def _convert_spatial_path(
     geom_encoding = geom_info["metadata"].get(geom_column, {}).get("encoding", "WKB")
     bounds = None
     if not skip_hilbert:
-        bounds, table_expr = _bounds_with_curve_fallback_checked(
+        bounds, table_expr = _bounds_with_curve_fallback(
             con,
             input_file,
             geom_column,
@@ -1218,6 +1212,8 @@ def _convert_spatial_path(
             max_angle_deg=max_angle_deg,
         )
         skip_hilbert = bounds is None
+        if skip_hilbert:
+            warn(_NO_BOUNDS_WARNING)
 
     if verbose:
         if secondary_columns:
--- a/geoparquet_io/core/convert.py
+++ b/geoparquet_io/core/convert.py
@@ -233,9 +233,21 @@ def detect_all_geometry_columns(input_file: str, verbose: bool = False) -> dict:
 
 
 def _calculate_bounds(
-    con, input_file, geom_column, verbose, is_parquet=False, encoding="WKB", table_expr=None
+    con,
+    input_file,
+    geom_column,
+    verbose,
+    is_parquet=False,
+    encoding="WKB",
+    table_expr=None,
+    required=True,
 ):
-    """Calculate dataset bounds from input file (or an explicit table_expr)."""
+    """Calculate dataset bounds from input file (or an explicit table_expr).
+
+    With ``required=False`` an unmeasurable dataset — no rows, or every
+    geometry empty or NULL — returns None instead of raising, letting the
+    caller convert without spatial ordering (issue #649).
+    """
     if verbose:
         debug("Calculating dataset bounds...")
 
@@ -277,6 +289,8 @@ def _calculate_bounds(
     bounds_result = con.execute(bounds_query).fetchone()
 
     if not bounds_result or any(v is None for v in bounds_result):
+        if not required:
+            return None
         raise GeoParquetError("Could not calculate dataset bounds")
 
     if verbose:
@@ -721,16 +735,17 @@ def _build_csv_conversion_query(geom_info, skip_hilbert, bounds, skip_invalid, s
 
     # With Hilbert ordering - use subquery
     xmin, ymin, xmax, ymax = bounds
+    bounds_box = f"ST_Extent(ST_MakeEnvelope({xmin}, {ymin}, {xmax}, {ymax}))"
+    # A WKT column can hold empty geometry, which ST_Hilbert rejects (#649).
+    unorderable = f"{geom_expr} IS NULL OR ST_IsEmpty({geom_expr})"
     return f"""
         SELECT
             * EXCLUDE ({exclude_cols}),
             {geom_expr} AS geometry{bbox_expr(geom_expr)}
         FROM {csv_read}
         {where_clause}
-        ORDER BY ST_Hilbert(
-            {geom_expr},
-            ST_Extent(ST_MakeEnvelope({xmin}, {ymin}, {xmax}, {ymax}))
-        )
+        ORDER BY ({unorderable}),
+            ST_Hilbert({_orderable_geom(geom_expr, xmin, ymin)}, {bounds_box})
     """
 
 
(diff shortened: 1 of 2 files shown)


geoparquet_io/core/convert.py
```diff
@@ -814,6 +829,43 @@ def _build_plain_select_query(input_file, is_parquet=False, is_csv=False, delimi
...
+    ``POLYGON EMPTY`` used to fail an entire conversion — while ``gpio sort
+    hilbert`` has always ordered the non-empty rows and appended the rest. The
+    substitute is a corner of the dataset envelope, so it is always in range,
+    and its key never matters: callers sort on an "unorderable" flag first,
+    which pins those rows last exactly where NULL geometry already landed
+    under DuckDB's NULLS LAST default.
+
+    Only the substitution is conditional — ``ST_Hilbert`` itself is evaluated
+    for every row, keeping spatial functions out of conditional execution
+    (the shape behind the segfaults in #642).
```
cayetanobv: This comment claims the opposite of what the SQL below does. The generated expression is:

```sql
CASE WHEN geom IS NULL OR ST_IsEmpty(geom) THEN ST_Point(xmin, ymin) ELSE geom END
```

`ST_Point` *is* a spatial function in the THEN branch, and the condition is the OR-with-`IS NULL` shape that `core/geometry_repair.py:176-186` documents as segfaulting DuckDB 1.5.1 -- "the raw-blob `IS NULL` term mis-aligns the selection vector fed to ST_MakeValid". The claim is true of `ST_Hilbert` only, which is the narrower point the next sentence already makes.

I could not reproduce a crash: 200,001-row GPKG, 201 NULL + 399 `POLYGON EMPTY`, both the ST_Read and parquet paths, all 200,000 rows out in the right order. Most likely `ST_Point(literal, literal)` constant-folds and the operand here is a decoded GEOMETRY rather than a raw blob. So not a bug -- but please reword. As written it reads as clearance for a shape this repo has already been burned by, and #642 only reproduced above ~8k rows, well past anything the new tests reach.

geoparquet_io/core/convert.py
```diff
@@ -233,9 +233,21 @@ def detect_all_geometry_columns(input_file: str, verbose: bool = False) -> dict:
...
 def _calculate_bounds(
-    con, input_file, geom_column, verbose, is_parquet=False, encoding="WKB", table_expr=None
+    con,
+    input_file,
+    geom_column,
+    verbose,
+    is_parquet=False,
+    encoding="WKB",
+    table_expr=None,
+    required=True,
```
cayetanobv: `required` only ever receives `False`. Both `_calculate_bounds` call sites are inside `_bounds_with_curve_fallback`, which hard-codes `required=False` into its shared `kwargs` -- so this default and the `raise GeoParquetError("Could not calculate dataset bounds")` branch below are now unreachable in `convert.py`.

Either drop the parameter and return `None` unconditionally, or keep the raise and let the single caller decide -- but a flag with one value at every call site is indirection without a payer.

geoparquet_io/core/convert.py
```diff
@@ -814,6 +829,43 @@ def _build_plain_select_query(input_file, is_parquet=False, is_csv=False, delimi
     return f"SELECT * FROM ST_Read('{input_file}')"
 
 
+def _bounds_with_curve_fallback_checked(con, *args, **kwargs):
```
cayetanobv: One-line wrapper with a single call site, and the thing it wraps is already a helper. The `warn` would read better next to `skip_hilbert = bounds is None` in `_convert_spatial_path` (line 1220), which is where the decision actually gets made -- right now the warning and the consequence sit in different functions and you have to jump between them to see that "unordered" is what actually happens.

cayetanobv (commented):
**Recommendation: approve after the CSV gap and the `_orderable_geom` comment.** (Posting as a comment review -- GitHub blocks APPROVE on your own PR.) The core mechanism is sound and the ordering contract now matches `sort hilbert`.

## The all-empty rescue only covers half the paths

`_calculate_csv_bounds` didn't get the `required=False` treatment, so the CSV/WKT path still dies where GPKG and parquet now succeed:

```
$ gpio convert allempty.csv out.parquet --wkt-column geom    # POLYGON EMPTY, LINESTRING EMPTY
Error: Conversion failed: Could not calculate dataset bounds from CSV
$ gpio convert zero.csv out.parquet --wkt-column geom        # header only
Error: Conversion failed: Could not calculate dataset bounds from CSV
```

Same command, same data shape, outcome depends on input format -- the class of inconsistency this PR set out to remove between `convert` and `sort hilbert`. Details inline on `TestEmptyGeometryFromCsv`.

## Claims I verified

- **Row order unchanged** for sources without empties: `buildings_test.gpkg` converted on `origin/main` and on this branch gives an identical 42-row centroid sequence.
- **No perf cost** from the duplicated `ST_GeomFromText` in the CSV `ORDER BY` -- DuckDB's CSE absorbs it. 300k-row WKT CSV: 4.54 / 4.67 s on main vs 4.58 / 4.59 s here, identical peak RSS. I went looking for a regression here; there isn't one.
- **No segfault at scale.** 200,001-row GPKG with 201 NULL + 399 `POLYGON EMPTY`, through both the ST_Read and the parquet path: 200,000 rows out, all 600 unorderable rows last, Hilbert order intact. (But see the inline note on `_orderable_geom` -- the comment claims something the SQL doesn't do.)
- **Degenerate envelope** (one point + one empty, so `xmin == xmax`) works -- covered incidentally by your own 2-row test.
- **Zero-row output is valid GeoParquet**: `geometry_types: []`, no column-level bbox, CRS present. The flagged behavior change is safe.
- Merges cleanly with #650 in either order; merged tree passes 887 / 7 skipped on the convert/crs/hilbert/sort/csv/geopackage/curved/linearize/repair selection.

geoparquet_io/core/convert.py
```diff
@@ -814,6 +829,43 @@ def _build_plain_select_query(input_file, is_parquet=False, is_csv=False, delimi
...
+    """``_bounds_with_curve_fallback`` plus the unmeasurable-dataset case.
+
+    Every geometry empty or NULL (or no rows at all) leaves nothing to build a
+    Hilbert envelope from, so the conversion proceeds unordered with a warning
+    instead of failing on bounds (issue #649).
+    """
+    bounds, table_expr = _bounds_with_curve_fallback(con, *args, **kwargs)
+    if bounds is None:
+        warn(
+            "No measurable geometry bounds (every geometry is empty or NULL); "
```
cayetanobv: This also fires for a **zero-row** source, where "every geometry is empty or NULL" is misleading -- there are no geometries at all. Observed on a 0-row GPKG:

```
Converting zero.gpkg...
No measurable geometry bounds (every geometry is empty or NULL); writing without Hilbert ordering.
```

Something like "no geometry to measure (no rows, or every geometry empty or NULL)" covers both. Minor, but it's the one user-visible string this PR adds, and the zero-row case is the behavior change you flagged for review.

tests/test_convert_empty_geometries.py
```diff
@@ -0,0 +1,189 @@
...
+        _write_gpkg(source, [(1, "POINT(1 1)"), (2, "POLYGON EMPTY")])
+        output = tmp_path / "out.parquet"
+
+        result = _convert(source, output, "--skip-hilbert")
+
+        assert result.exit_code == 0, result.output
+        assert len(_rows(output)) == 2
+
+
+class TestEmptyGeometryFromCsv:
```
cayetanobv, line 140: This class covers one empty among valid rows, which passes -- but it leaves the CSV analogue of `test_all_geometries_empty_converts_without_ordering` untested, and that case still fails:

```
$ gpio convert allempty.csv out.parquet --wkt-column geom    # POLYGON EMPTY, LINESTRING EMPTY
Error: Conversion failed: Could not calculate dataset bounds from CSV
$ gpio convert zero.csv out.parquet --wkt-column geom        # header only
Error: Conversion failed: Could not calculate dataset bounds from CSV
```

The CSV path computes bounds through `_calculate_csv_bounds`, which is a separate function that didn't get the `required=False` treatment -- it still raises unconditionally at `Could not calculate dataset bounds from CSV`. So the same all-empty source converts as a GPKG or parquet and errors as a CSV.

Either extend the fix (return `None` there too and let `_convert_csv_path` skip ordering the same way `_convert_spatial_path` now does) or say in the PR body that CSV is deliberately out of scope. Right now the asymmetry is silent, and adding the two cases above to this class is what would surface it.

cayetanobv:
Both blocking items and the three notes fixed in 5e610e8; every finding reproduced first.

**CSV gap — confirmed and fixed.** Both commands failed exactly as you showed. `_calculate_csv_bounds` now returns None for an unmeasurable dataset, mirroring `_calculate_bounds`, and the CSV path warns and writes unordered. `allempty.csv` and a header-only CSV both convert now, with tests for each.

**`_orderable_geom` docstring — confirmed wrong, reworded.** You are right that `ST_Point` sits in a THEN branch, so "keeps spatial functions out of conditional execution" was not what the SQL does. It now says what is true: `ST_Hilbert` is unconditional, and the substitution's `ST_Point` takes constant arguments over an already-decoded GEOMETRY rather than the raw-blob `IS NULL` shape from #642 — with your 200k-row result as the evidence rather than an unearned clearance.

**`required` — confirmed single-valued, dropped.** Both call sites went through the hard-coded `kwargs`, so the default and its raise were unreachable. `_calculate_bounds` now returns None unconditionally and the caller decides.

**Wrapper — removed.** The warning moved next to `skip_hilbert = bounds is None` in `_convert_spatial_path`, and is shared with the CSV path as `_NO_BOUNDS_WARNING`.

**Zero-row wording — confirmed, reworded** to "No geometry to measure (no rows, or every geometry empty or NULL); writing without Hilbert ordering." Verified against a 0-row GeoPackage.

**One finding deliberately not fixed here:** CodeRabbit's point that `where_clause` removes NULL WKT rows. It does — 3 rows in, 2 out — but that predates this PR (`main` gives the same 2 rows), and fixing it changes row counts for every CSV/WKT conversion, with `--skip-invalid` semantics to settle. Filed as #655 with a proposed fix rather than smuggled in here.

859 passed / 7 skipped on the convert/crs/hilbert/sort/csv/geopackage/curved/linearize selection; pre-commit and xenon clean.

tests/test_convert_empty_geometries.py
```diff
@@ -0,0 +1,210 @@
...
+        con.commit()
+        con.close()
+        output = tmp_path / "out.parquet"
+
+        result = _convert(source, output)
+
+        assert result.exit_code == 0, result.output
+        assert len(_rows(output)) == 2
+
+    def test_all_empty_csv_converts_without_ordering(self, tmp_path):
```
cayetanobv, line 191: These two new CSV tests landed in `TestLinearizedEmptyCurve`, whose docstring is about #647 turning `CIRCULARSTRING EMPTY` into `LINESTRING EMPTY`. Neither touches curves or linearization -- they belong in `TestEmptyGeometryFromCsv` just above, which is exactly the CSV bounds behavior they exercise.

Purely organizational, and they pass where they are. Mentioning it because `TestEmptyGeometryFromCsv` is where someone will look for CSV coverage, and finding only the mixed case there is what let this gap hide the first time round.

cayetanobv (commented):
## Re-review: all four findings addressed

**Recommendation: approve.** (Comment review -- GitHub blocks APPROVE on your own PR.)

Verified each fix:

- **The CSV gap is closed.** `_calculate_csv_bounds` returns `None` and `_convert_csv_path` warns and skips ordering, so the two formats now agree:
  ```
  CSV all-empty WKT   exit=0 rows=2  "No geometry to measure (no rows, or every geometry empty or NULL); writing without Hilbert ordering."
  CSV header-only     exit=0 rows=0  same warning
  ```
  Both were hard errors last round.
- **`required` is gone**, and `_calculate_bounds` just returns `None` -- no flag with one value, no unreachable raise. Confirmed nothing else depended on the old `"Could not calculate dataset bounds"` string from this module: the two other occurrences (`core/stac.py:156`, `core/add/country_codes.py:233`) raise their own.
- **`_bounds_with_curve_fallback_checked` is gone**, and the warning now sits at the decision point in both paths, next to the `skip_hilbert` assignment it explains. `_NO_BOUNDS_WARNING` as a shared constant is the right call given two callers.
- **The `_orderable_geom` comment is honest now** -- it says `ST_Point` *is* in a THEN branch and explains why that isn't #642's shape, rather than claiming no spatial function is conditional. That's the version a future maintainer can act on.

Also re-confirmed row order is untouched for sources without empties: `buildings_test.gpkg` gives an identical 42-row centroid sequence on `main`, this branch, and #656.

867 passed / 7 skipped on the convert/csv/hilbert/crs/curved/linearize/sort/geopackage selection (run on #656, which stacks this).

## One thing that is *not* a regression here, but is worth knowing

An all-NULL lat/lon CSV dies with a raw Python error:

```
$ gpio convert all_null_latlon.csv out.parquet
Error: Conversion failed: '<' not supported between instances of 'NoneType' and 'int'
```

`convert.py:527`, `_validate_latlon_ranges` -- `if min_lat < -90` with `min_lat = None`. I initially suspected your `_calculate_bounds` change, but that function runs its own MIN/MAX query and is untouched by this PR; it fails identically on `origin/main`. So: pre-existing, not yours.

Flagging it because it's the lat/lon analogue of the case this PR just rescued for WKT, so the formats are inconsistent again one level down -- and because #656 makes it more reachable by retaining NULL-coordinate rows. Might be worth folding into #656 or a small follow-up, since a `None` guard there is a two-line change.

PR #653, opened 2026-08-20
Reproject before repairing geometry in convert geojson

### LINKED ISSUE #652: convert geojson: geometry is repaired before reprojection, so ST_ReducePrecision raises TopologyException
# Geometry is repaired before reprojection, so ST_ReducePrecision throws TopologyException

`convert geojson` repairs invalid geometry in the **source** CRS and reprojects
afterwards. Reprojection is not topology-preserving, so a polygon that
`ST_MakeValid` just barely made valid in the source CRS can come out of
`ST_Transform` invalid in WGS84 — and `ST_ReducePrecision`, a GEOS topology
operation, then raises on it.

`core/geojson_stream.py: _build_geojson_query` builds:

```
ST_ReducePrecision( ST_Transform( ST_MakeValid(geom), src, 4326 ), 10^-precision )
```

## Symptom

```
$ gpio convert geojson mancomunidades.parquet --precision 6
_duckdb.InvalidInputException: Invalid Input Error: TopologyException:
side location conflict at -6.3052900000000003 39.878601000000003.
```

Also seen as `TopologyException: unable to assign free hole to a shell`.

Downstream this kills the whole PMTiles pipeline: `create_pmtiles` pipes
`convert geojson` into tippecanoe, so the failure surfaces as
`tippecanoe failed with exit code 110`, and in portolan-cli one bad feature
aborts an entire 28-collection catalog build.

## gpio already does this correctly on its other path

`core/pmtiles.py: _build_gpio_commands` has two routes:

- with an explicit `src_crs`, it runs `gpio convert reproject` **first** and
  pipes the result into `gpio convert geojson` — the repair then happens on
  already-reprojected data, in the CRS the output is expressed in. Correct.
- with `src_crs=None`, `convert_to_geojson` auto-detects the CRS from the
  parquet metadata and reprojects **inside** the query, after the repair.

Same file, same target CRS, opposite outcome:

```
gpio convert geojson mancomunidades.parquet --precision 6                -> exit 1
gpio convert reproject … --src-crs EPSG:25830 | gpio convert geojson -   -> exit 0
```

## Reproduction

Real data: three collections of the SITEX / Junta de Extremadura open data
(EPSG:25830), converted with `gpio.convert(...).write(...)`.

Every source geometry is valid (`ST_IsValid` reports 0 invalid rows). After
`ST_MakeValid` + `ST_Transform` to EPSG:4326: 8 invalid features in
mancomunidades (31 rows), 11 in CORINE 2012 (14,545 rows), 3 in CORINE 2018
(14,779 rows). The offending mancomunidades feature keeps all 5,486 vertices
and its exact area through the transform; only `ST_IsValid` flips true -> false.

It throws at every precision from 6 to 12, and whether `ST_MakeValid` runs
before or after `ST_ReducePrecision` — so `--no-repair-geometry` and lowering
`--precision` are both non-workarounds.

## Fix

Reproject first, repair in the CRS the output is expressed in:

```
ST_ReducePrecision( ST_MakeValid( ST_Transform(geom, src, 4326) ), 10^-precision )
```

Verified on all three files: 31 / 14,545 / 14,779 features out, no exception,
nothing dropped and no precision given up.

## Note on why this was not seen before

It only became reachable via #644. Before that fix the Table API wrote
`crs: null`, so pipelines that reprojected upstream produced WGS84 parquet, the
GeoJSON stream inserted no `ST_Transform` at all, and the repair happened in the
same CRS the precision reduction consumed. #644 put a reprojection into that
query for the first time and exposed the latent ordering bug.

cayetanobv:
Fixes #652.

`_build_feature_query` repaired geometry in the **source** CRS and reprojected
afterwards:

```
ST_ReducePrecision( ST_Transform( ST_MakeValid(geom), src, 4326 ), 10^-precision )
```

Reprojection is not topology-preserving, so a polygon `ST_MakeValid` had just
made valid in the source CRS could come out of `ST_Transform` **invalid** in
WGS84 — and `ST_ReducePrecision`, a GEOS topology operation, then raised on it.
One such feature aborts the whole conversion, and through `create_pmtiles` it
takes the tippecanoe pipeline with it (`tippecanoe failed with exit code 110`).

gpio already got this right on its other path: with an explicit `src_crs`,
`core/pmtiles.py` runs `gpio convert reproject` in a separate process **first**
and pipes into `convert geojson`, so the repair lands on already-reprojected
data. Only the auto-detected CRS path had the two in the wrong order.

## The change

Reproject in a subquery, then repair, so the repair runs in the CRS the output
is expressed in. The transform goes in a subquery rather than inline because
`repair_geometry_sql` repeats its argument — inlining would evaluate
`ST_Transform` four times per row, twenty with `--write-bbox`. It is now
evaluated exactly once, which a test asserts.

## Tests

Four regression tests, all failing on `main` and passing here:

- the repair consumes the reprojected column, not the raw one
- `ST_Transform` appears exactly once even with `--write-bbox`
- a real-world sliver polygon survives precision reduction
- the NULL-geometry filter survives the subquery rewrite

The fixture is a 7-vertex sliver reduced from the failing feature in the SITEX
Extremadura *mancomunidades* layer (EPSG:25830): valid in the source CRS — two
of its vertices differ only in the 13th decimal — and invalid once transformed.

## Real-world verification

Found rebuilding a 28-collection catalog from Junta de Extremadura open data in
its native CRS. Three collections failed on `main`; all three convert here with
nothing dropped and no precision given up:

| collection | features |
|---|---|
| mancomunidades-integrales | 31 |
| CORINE 2012 | 14,545 |
| CORINE 2018 | 14,779 |

Every source geometry is valid (`ST_IsValid` reports 0 invalid rows); after
`ST_MakeValid` + `ST_Transform` to WGS84, 8 / 11 / 3 features respectively come
out invalid. The full catalog now builds end to end: 28/28 GeoParquet, PMTiles,
thumbnails and styles.

This only became reachable via #644 — before it the Table API wrote `crs: null`,
so pipelines that reprojected upstream produced WGS84 parquet and this query
inserted no `ST_Transform` at all.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved GeoJSON export when coordinate reference system reprojection and geometry repair are applied together.
  * Preserved valid small geometries during precision reduction.
  * Continued filtering null geometries correctly.

* **Tests**
  * Added regression coverage for reprojection, geometry repair, precision handling, and null geometry filtering.

### DIFF
--- a/geoparquet_io/core/geojson_stream.py
+++ b/geoparquet_io/core/geojson_stream.py
@@ -159,18 +159,30 @@ def _build_feature_query(
     """
     quoted_geom = quote_identifier(geometry_column)
 
-    # Repair invalid geometry (issue #506) before reproject/precision/GeoJSON so
-    # downstream consumers (e.g. tippecanoe in the PMTiles pipeline) never see a
+    # Reproject first, in a subquery, so everything below sees the geometry in
+    # the CRS it will be written in. Reprojection is not topology-preserving: a
+    # polygon ST_MakeValid has just made valid in the source CRS can come out of
+    # ST_Transform invalid in WGS84, and ST_ReducePrecision - a GEOS topology
+    # operation - then raises TopologyException on it. Repairing after the
+    # transform is also what the explicit --src-crs route already does, since
+    # core/pmtiles.py reprojects in a separate process before piping into this
+    # query; only the auto-detected CRS path had the two in the wrong order.
+    #
+    # The transform lives in a subquery rather than inline because
+    # repair_geometry_sql repeats its argument, and ST_Transform is far too
+    # expensive to evaluate once per repetition per row.
+    if source_crs:
+        source_ref = (
+            f"(SELECT * REPLACE ("
+            f"ST_Transform({quoted_geom}, '{source_crs}', '{WGS84_CRS}') AS {quoted_geom}"
+            f") FROM {source_ref} WHERE {quoted_geom} IS NOT NULL)"
+        )
+
+    # Repair invalid geometry (issue #506) before precision/GeoJSON so downstream
+    # consumers (e.g. tippecanoe in the PMTiles pipeline) never see a
     # self-intersection. The NULL filter in the WHERE clause stays on the raw
     # column. ST_AsGeoJSON requires native GEOMETRY, which this already is.
-    geom_base = repair_geometry_sql(quoted_geom) if repair_geometry else quoted_geom
-
-    # Apply reprojection if needed
-    if source_crs:
-        # Transform to WGS84 before converting to GeoJSON
-        geom_for_output = f"ST_Transform({geom_base}, '{source_crs}', '{WGS84_CRS}')"
-    else:
-        geom_for_output = geom_base
+    geom_for_output = repair_geometry_sql(quoted_geom) if repair_geometry else quoted_geom
 
     # Apply coordinate precision using ST_ReducePrecision before GeoJSON conversion.
     # The grid size is 10^-precision (e.g., precision=7 -> grid=0.0000001).
--- a/tests/test_geojson_stream.py
+++ b/tests/test_geojson_stream.py
@@ -541,3 +541,80 @@ class TestWgs84Constant:
     def test_wgs84_crs_value(self):
         """Test WGS84_CRS is correctly defined."""
         assert WGS84_CRS == "EPSG:4326"
+
+
+class TestRepairAfterReprojection:
+    """Geometry must be repaired in the CRS it is written in, not the source CRS.
+
+    Reprojection is not topology-preserving, so repairing first leaves
+    ST_ReducePrecision to raise TopologyException on geometry the transform has
+    just invalidated.
+    """
+
+    # A sliver from the SITEX Extremadura "mancomunidades" layer: valid in
+    # EPSG:25830 (two of its vertices differ only in the 13th decimal), invalid
+    # once transformed to WGS84. Repairing before the transform makes
+    # ST_ReducePrecision raise "TopologyException: unable to assign free hole to
+    # a shell" on it.
+    SLIVER_25830 = (
+        "POLYGON (("
+        "217334.43251087563 4419513.990336159, "
+        "217334.2614384655 4419513.996670783, "
+        "217334.2527995768 4419513.996990671, "
+        "217334.5221580651 4419514.098189879, "
+        "217334.60769426968 4419514.09502257, "
+        "217334.60769426887 4419514.09502257, "
+        "217334.43251087563 4419513.990336159))"
+    )
+
+    def test_repair_wraps_the_reprojected_geometry(self):
+        """The repair must be applied after the transform, not before it."""
+        query = _build_feature_query("t", "geometry", ["name"], source_crs="EPSG:25830")
+        # The transform happens in the subquery that feeds the repair, so the
+        # repaired expression is a plain column reference.
+        assert "ST_MakeValid(ST_Transform(" not in query.replace(" ", "")
+        transform_pos = query.index("ST_Transform")
+        repair_pos = query.index("ST_MakeValid")
+        assert transform_pos > repair_pos, "repair must consume the reprojected column"
+
+    def test_transform_is_evaluated_once_per_row(self):
+        """repair_geometry_sql repeats its argument; ST_Transform must not be."""
+        query = _build_feature_query(
+            "t", "geometry", ["name"], source_crs="EPSG:25830", write_bbox=True
+        )
+        assert query.count("ST_Transform") == 1
+
+    def test_reprojected_sliver_survives_precision_reduction(self):
+        """The real-world regression: this raised before the reorder."""
+        import duckdb
+
+        con = duckdb.connect()
+        con.execute("INSTALL spatial; LOAD spatial;")
+        con.execute(
+            f"CREATE TABLE t AS SELECT ST_GeomFromText('{self.SLIVER_25830}') "
+            "AS geometry, 'sliver' AS name"
+        )
+        query = _build_feature_query(
+            "t", "geometry", ["name"], precision=6, source_crs="EPSG:25830"
+        )
+        rows = con.execute(query).fetchall()
+        assert len(rows) == 1
+        assert json.loads(rows[0][0])["type"] == "Feature"
+
+    def test_null_geometry_still_filtered_with_reprojection(self):
+        """The NULL filter must survive the subquery rewrite."""
+        import duckdb
+
+        con = duckdb.connect()
+        con.execute("INSTALL spatial; LOAD spatial;")
+        con.execute(
+            "CREATE TABLE t AS SELECT * FROM (VALUES "
+            f"(ST_GeomFromText('{self.SLIVER_25830}'), 'kept'), "
+            "(NULL::GEOMETRY, 'dropped')) AS v(geometry, name)"
+        )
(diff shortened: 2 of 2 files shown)


tests/test_geojson_stream.py
```diff
@@ -541,3 +541,80 @@ class TestWgs84Constant:
...
+
+    def test_repair_wraps_the_reprojected_geometry(self):
+        """The repair must be applied after the transform, not before it."""
+        query = _build_feature_query("t", "geometry", ["name"], source_crs="EPSG:25830")
+        # The transform happens in the subquery that feeds the repair, so the
+        # repaired expression is a plain column reference.
+        assert "ST_MakeValid(ST_Transform(" not in query.replace(" ", "")
+        transform_pos = query.index("ST_Transform")
+        repair_pos = query.index("ST_MakeValid")
+        assert transform_pos > repair_pos, "repair must consume the reprojected column"
```
cayetanobv, line 578: This assertion does regress-guard (it fails on `main`, I checked), but via textual position, which is a fragile proxy for the property you actually care about.

`transform_pos > repair_pos` holds because the subquery lives in the `FROM` clause and `FROM` comes after the `SELECT` list. It's a fact about query layout, not about which expression consumes which -- a future refactor that moved the transform into a CTE above the `SELECT` would flip it while being perfectly correct, and conversely a query that repaired the raw column and transformed in a *later* wrapper would satisfy it while being the exact bug this PR fixes.

The assertion above it is weaker still: `"ST_MakeValid(ST_Transform("` was already absent on `main`, because `main` generated `ST_Transform(CASE WHEN ... ST_MakeValid("geometry") ... END, ...)` -- the nesting was the other way round, so that substring never appeared. It's vacuous in both directions.

The property worth pinning is behavioral, and you already have the machinery for it in `test_reprojected_sliver_survives_precision_reduction`: the repair sees WGS84 coordinates. Asserting on the *output* -- that the emitted coordinates are in degrees, not metres -- would survive any layout change.

tests/test_geojson_stream.py
```diff
@@ -541,3 +541,80 @@ class TestWgs84Constant:
...
+
+    def test_transform_is_evaluated_once_per_row(self):
+        """repair_geometry_sql repeats its argument; ST_Transform must not be."""
+        query = _build_feature_query(
+            "t", "geometry", ["name"], source_crs="EPSG:25830", write_bbox=True
+        )
+        assert query.count("ST_Transform") == 1
```
cayetanobv, line 585: The name promises per-row evaluation; the body counts a substring in the generated SQL. Those come apart precisely where it matters: the whole reason for the subquery is that you're relying on DuckDB's optimizer not to duplicate the expression when it flattens the projection, and a string count can't see that. This would pass unchanged if the optimizer inlined `ST_Transform` into all 20 references.

Good news -- it does hold. `EXPLAIN (FORMAT JSON)` on the generated query has exactly one `st_transform` node, with and without `write_bbox`. So the assertion you want is available:

```python
plan = con.execute("EXPLAIN (FORMAT JSON) " + query).fetchone()[1]
assert len(re.findall(r"st_transform", plan, re.I)) == 1
```

That tests the claim in the comment on `geojson_stream.py:171-173` rather than a proxy for it, and it would catch a future DuckDB whose optimizer stops being so obliging.

geoparquet_io/core/geojson_stream.py
```diff
@@ -159,18 +159,30 @@ def _build_feature_query(
...
+    # core/pmtiles.py reprojects in a separate process before piping into this
+    # query; only the auto-detected CRS path had the two in the wrong order.
+    #
+    # The transform lives in a subquery rather than inline because
+    # repair_geometry_sql repeats its argument, and ST_Transform is far too
+    # expensive to evaluate once per repetition per row.
+    if source_crs:
+        source_ref = (
+            f"(SELECT * REPLACE ("
+            f"ST_Transform({quoted_geom}, '{source_crs}', '{WGS84_CRS}') AS {quoted_geom}"
```
cayetanobv, line 177: Unrelated to this fix and pre-existing on `main`, so entirely your call whether it's worth a follow-up: `source_crs` is interpolated straight into SQL with no escaping, so a CRS string containing a single quote produces broken (or injected) SQL. The move into a `FROM`-clause subquery widens the blast radius slightly -- a malformed value now breaks the relation rather than one projection expression.

Everything else in this module goes through `quote_identifier`. If `source_crs` only ever arrives from CRS detection or `--src-crs`, the practical risk is about nil; mentioning it because this line is the one that moved.

cayetanobv (commented):
**Recommendation: approve.** (Posting as a comment review -- GitHub blocks APPROVE on your own PR.) The diagnosis is right, the fix is minimal and in the right place, and I reproduced the whole thing end to end.

## Verified independently

- **The bug reproduces on `main`** with the exact reported failure, and is gone here. Same sliver, same query, `precision=6`, `EPSG:25830`:
  ```
  origin/main:  InvalidInputException: TopologyException: unable to assign free hole
                to a shell at -6.305288 39.878601000000003
  this branch:  {"type":"Feature","geometry":{...},"properties":{"name":"sliver"}}
  ```
- **All four new tests fail on `main`**, as claimed -- I copied the new test file onto `main` and ran it: 4 failed.
- **The output really is reprojected.** Worth stating because the ASCII plan is misleading here: `EXPLAIN` renders the flattened transform as a bare `geometry` in the projections and shows `ST_Transform` only inside the pushed-down `FILTER`, which reads at a glance like the repair is consuming the raw column. It isn't -- the emitted coordinates are `-6.305287, 39.878602`, matching a direct `ST_Transform` of the source.
- **"Evaluated exactly once" holds in the optimized plan.** `EXPLAIN (FORMAT JSON)` contains exactly one `st_transform` node, both with and without `--write-bbox` -- DuckDB merges the inner and outer `IS NOT NULL` filters rather than transforming twice. The subquery is doing what you wanted it to. (The *test* doesn't check this -- see inline.)
- **`ST_Transform` never returns NULL for non-NULL input**, so moving the filter next to the transform can't silently drop rows: empty geometries pass through as `POINT EMPTY` / `POLYGON EMPTY`, out-of-domain coordinates come back as `POINT (inf inf)`. I went looking for a row-count regression here; there isn't one.
- 251 passed on the `geojson or pmtiles or repair or reproject` selection.

## One thing to document

The rescued features change geometry type. For the fixture sliver:

| | |
|---|---|
| source type, EPSG:25830 | `POLYGON`, `ST_IsValid` = true |
| after `ST_Transform` to WGS84 | `ST_IsValid` = **false** |
| after `ST_MakeValid` | **`GEOMETRYCOLLECTION`** |

So the emitted GeoJSON is `"geometry":{"type":"GeometryCollection","geometries":[{"type":"Polygon",...}]}` rather than a Polygon. That's strictly better than the hard failure on `main`, and it's inherent to `ST_MakeValid` rather than anything this PR introduced -- but it's not visible in the "nothing dropped and no precision given up" summary, and by your own numbers it's exactly the 8 / 11 / 3 features per collection that this PR rescues. Given the stated downstream consumer is tippecanoe via the PMTiles pipeline, and that mixed geometry types are the kind of thing tile pipelines and GeoParquet `geometry_types` metadata care about, it's worth a line in the PR body at minimum.

`ST_CollectionExtract(ST_MakeValid(g), 3)` returns `MULTIPOLYGON` for this fixture if you'd rather keep the dimension stable -- but that's a `repair_geometry_sql` design question affecting every caller, so out of scope here. Flagging it, not asking for it.

PR #656, opened 2026-08-20
Keep CSV rows whose geometry is NULL

### LINKED ISSUE #655: CSV/WKT conversion silently drops rows with NULL geometry
## Description

`gpio convert` on a CSV with a WKT column drops any row whose geometry is NULL, without a warning. Three rows in, two out:

```bash
printf 'id,geom\n1,POINT(1 1)\n2,\n3,POINT(5 5)\n' > nullwkt.csv
gpio convert nullwkt.csv out.parquet --wkt-column geom
python -c "import duckdb; print(duckdb.connect().execute(\"SELECT count(*) FROM 'out.parquet'\").fetchall())"
# [(2,)]
```

`_get_geom_expr_and_where()` builds `WHERE <wkt_col> IS NOT NULL` (plus `AND <parsed> IS NOT NULL` under `--skip-invalid`), and that clause is applied to the conversion query itself, not just to the bounds pass — so the rows never reach the output.

Every other input path keeps them: a GeoPackage or parquet source with NULL geometry converts row-for-row, and after #651 NULL and empty geometries sort last rather than being dropped or failing the conversion. Attribute data attached to a NULL geometry is still data, and losing it silently is worse than either keeping it or refusing.

Not introduced by #651 — `main` behaves identically. Filed separately because the fix changes row counts for every CSV/WKT conversion, which deserves its own review rather than riding along.

## Proposed fix

Use the NULL filter only where it belongs — in `_calculate_csv_bounds()` — and let `_build_csv_conversion_query()` emit all rows, with `TRY(ST_GeomFromText(...))` yielding NULL geometry for unparseable or missing WKT. #651's unorderable-first `ORDER BY` already places those rows last, so the ordering contract needs no change.

Points to settle in review:

- `--skip-invalid` currently means "drop rows whose WKT will not parse". If NULL rows are kept, does `--skip-invalid` still drop unparseable ones, or does it become "keep them with NULL geometry"? The flag's help text says skip, so probably the former — meaning two different filters with two different scopes.
- The lat/lon CSV path has the same shape (`WHERE lat IS NOT NULL AND lon IS NOT NULL`) and should move with it.
- Worth a warning reporting how many rows carry NULL geometry, mirroring what `sort hilbert` reports for empty/null geometry.
- Tests: NULL WKT among valid rows, all-NULL WKT, NULL lat/lon, and each with and without `--skip-invalid`.

cayetanobv:
## Description

Stacked on #651 (it shares `_build_csv_conversion_query`, and the retained rows rely on its unorderable-first ordering to have a defined place). Closes #655, the finding CodeRabbit raised on #651 that I deliberately left out of that PR because it changes row counts.

`WHERE <wkt_col> IS NOT NULL` was applied to the conversion query, not just to the bounds pass, so a CSV row with an empty geometry column disappeared — three rows in, two out, no warning, attributes gone. The lat/lon path did the same with `WHERE lat IS NOT NULL AND lon IS NOT NULL`. Both are pre-existing (`main` behaves identically); every other input format keeps such rows.

NULL already propagates through both geometry expressions, so the filters were the only thing dropping the rows:

```
ST_GeomFromText(NULL)              -> NULL
ST_Point(NULL, NULL)               -> NULL
ST_XMin(ST_GeomFromText(NULL))     -> NULL   (so bounds aggregates skip them)
```

The bounds pass keeps its filters — a row with nothing to measure must not enter the envelope — and #651's `ORDER BY (unorderable), ST_Hilbert(...)` places the NULL rows last.

## The `--skip-invalid` question

The issue flagged this as needing a decision, and this PR takes the narrow reading: **`--skip-invalid` skips geometry that cannot be *parsed*.** A row with no geometry at all is not invalid, so it is kept with NULL geometry. That matches the flag's own help text ("skip rows with invalid geometries") and keeps the two concerns separate — malformed input is a data-quality problem, missing geometry is a legitimate state.

Concretely, on `tests/data/points_invalid_wkt.csv` (2 valid POINTs, 1 unparsable, 1 empty) with `--skip-invalid`: was 2 rows out, now 3 — the unparsable row still dropped, the empty one retained with NULL geometry.

Two existing tests in `test_convert.py` asserted the old contract and now assert the new one; the CLI help and `docs/guide/convert.md` spell out the distinction.

## Tests

New `tests/test_convert_csv_null_geometry.py`: NULL WKT retained; NULL geometry sorts last while the real rows keep Hilbert order; all-NULL WKT converts; unparsable WKT still errors without `--skip-invalid`; `--skip-invalid` drops the unparsable row and keeps the empty one; NULL lat/lon and a partially missing coordinate retained; and bounds still measured from real geometry only. All eight failed before the change.

799 passed / 7 skipped on the csv/convert/hilbert/crs/curved/linearize selection; pre-commit and xenon clean.

## Breaking Changes
**Does this PR introduce breaking changes?** Behaviorally yes, deliberately: CSV/TSV conversions now emit rows they previously dropped, and `--skip-invalid` drops strictly less. No API change.

## Related Issue(s)
- Closes #655
- Stacked on #651

## Checklist

## Summary by CodeRabbit

- **New Features**
  - CSV and TSV conversion now retains rows with missing geometry, storing them with `NULL` geometry and preserving their attributes.
  - Rows with invalid WKT continue to fail by default or can be skipped with `--skip-invalid`.
  - Missing geometries are sorted after valid geometries and excluded from bounds calculations.

- **Documentation**
  - Clarified handling of missing and invalid geometries, coordinate validation, and `--skip-invalid` behavior.


cayetanobv (commented):
**Recommendation: approve after the stale lat/lon warning.** (Comment review -- GitHub blocks APPROVE on your own PR.) The reasoning is right, the `--skip-invalid` reading is the defensible one, and I could not find a case where a row is wrongly kept or wrongly dropped. One user-facing message now contradicts the new behavior, and there is a lat/lon coverage gap that hides a pre-existing crash.

## Verified

- **Rows are retained, and the retained ones sort last.** Partial-NULL lat/lon: `main` writes 2 rows, this branch writes 3 with 1 NULL geometry. WKT NULL rows land after the ordered rows, and the real rows keep Hilbert order.
- **Empty CSV fields are NULL in every spelling I tried.** I went looking for a hole between "absent" and "unparsable empty string", since the whole fix hinges on `<wkt_col> IS NULL`:

  | field | result |
  |---|---|
  | `2,` (unquoted empty) | kept, NULL geometry |
  | `2,""` (quoted empty) | kept, NULL geometry |
  | `2,""` + `--skip-invalid` | kept, NULL geometry |
  | `2,"  "` (whitespace) | errors -- unparsable, correctly *not* treated as absent |

  DuckDB's CSV reader maps quoted-empty to NULL too, so the sentinel catches both. Whitespace erroring is the right side of your own distinction.
- **Bounds are still measured from real geometry only** -- the retained rows do not drag the envelope, and `_get_geom_expr_and_where`'s new docstring makes the split explicit, which is the bit that would otherwise read as an inconsistency.
- **NULL-geometry output is valid GeoParquet.** `gpio check all` passes clean on a mixed file with the bbox covering intact -- worth stating, since NULL geometry with a NULL bbox struct is a shape the CSV path had not been emitting before.
- **The two amended `test_convert.py` tests assert the new contract properly**, including *which* row is NULL rather than just the count. That is the assertion that would catch a future regression folding the two concerns back together.
- 867 passed / 7 skipped on the convert/csv/hilbert/crs/curved/linearize/sort/geopackage selection.

## On the behavior change

Worth being explicit that this is the good kind of breaking: previously three rows in, two out, silently, attributes gone. Every other input format keeps them. The `--skip-invalid` narrowing (unparsable is not the same as absent) matches the flag's own help text, and you updated both the help and the guide. No objection to any of it.

geoparquet_io/core/convert.py
```diff
@@ -708,8 +714,9 @@ def bbox_expr(geom):
         geom_expr = f"ST_Point(CAST({lon_col} AS DOUBLE), CAST({lat_col} AS DOUBLE))"
         exclude_cols = f"{lat_col}, {lon_col}"
 
-        # Skip rows with NULL lat/lon
-        where_clause = f"WHERE {lat_col} IS NOT NULL AND {lon_col} IS NOT NULL"
+        # A missing coordinate means no geometry, not no row: ST_Point yields
+        # NULL and the row keeps its attributes (issue #655).
+        where_clause = ""
```
cayetanobv, line 735: Dropping this filter makes `_validate_latlon_ranges` lie. It still warns about the very rows this PR now keeps:

```
$ gpio convert partial_null_latlon.csv out.parquet     # rows: 1,(10 10)  2,(,)  3,(20 20)
!  Warning: 1 rows have NULL lat/lon values and will be skipped
```

On `main` that was accurate -- 2 rows out. Here it is 3 rows out with 1 NULL geometry, so the warning describes behavior this PR deliberately removed. It is the same claim you already corrected in the CLI help and `docs/guide/convert.md`, just in the other geometry path (`convert.py:519-520`).

Either reword it to say the rows are kept with NULL geometry, or drop it -- the NULL count is arguably still useful information, just not as a "will be skipped" claim.

geoparquet_io/core/convert.py
```diff
@@ -682,24 +682,30 @@ def bbox_expr(geom):
...
+                        {wkt_col} IS NULL AS __gpio_wkt_missing
                     FROM {csv_read}
                 )
                 SELECT
-                    * EXCLUDE (geometry),
+                    * EXCLUDE (geometry, __gpio_wkt_missing),
                     geometry{bbox_expr("geometry")}
                 FROM parsed_geoms
-                WHERE geometry IS NOT NULL
+                WHERE __gpio_wkt_missing OR geometry IS NOT NULL
```
cayetanobv: Two small things about the sentinel column.

**It can collide.** A CSV with a column literally named `__gpio_wkt_missing` fails with a baffling error rather than a clear one:

```
Error: Conversion failed: Conversion Error: Could not convert string 'x' to BOOL
when casting from source column ...
```

Vanishingly unlikely, and it fails loudly rather than silently, so this alone is not worth changing.

**But you may not need the column at all.** The only reason for the sentinel is that the CTE does `* EXCLUDE ({wkt_col})`, so the outer `WHERE` cannot see the raw WKT. Keep the column in the CTE and exclude it in the outer `SELECT` instead, and the predicate becomes direct:

```sql
WITH parsed_geoms AS (
    SELECT *, TRY(ST_GeomFromText(geom)) AS geometry FROM ...
)
SELECT * EXCLUDE (geom, geometry), geometry, <bbox>
FROM parsed_geoms
WHERE geom IS NULL OR geometry IS NOT NULL
```

Same semantics, one less synthesized column, and the collision goes away with it. Your call whether it is worth the churn.

tests/test_convert_csv_null_geometry.py
```diff
@@ -0,0 +1,141 @@
...
+        assert result.exit_code == 0, result.output
+        con = duckdb.connect()
+        con.execute("INSTALL spatial; LOAD spatial;")
+        try:
+            xmin, ymax = con.execute(
+                f"SELECT MIN(ST_XMin(geometry)), MAX(ST_YMax(geometry)) FROM '{output}'"
+            ).fetchall()[0]
+        finally:
+            con.close()
+        assert (xmin, ymax) == pytest.approx((0.0, 10.0))
```
cayetanobv, line 141: The lat/lon cases here cover NULL and partially-missing coordinates, but not *all* rows NULL -- the lat/lon analogue of `test_all_null_wkt_converts` above. That asymmetry hides a real failure:

```
$ gpio convert all_null_latlon.csv out.parquet          # id,lat,lon / 1,, / 2,,
Error: Conversion failed: '<' not supported between instances of 'NoneType' and 'int'
```

`convert.py:527` in `_validate_latlon_ranges` -- `if min_lat < -90` with `min_lat = None`.

To be clear about ownership: this is **pre-existing**. It fails identically on `origin/main`, and neither this PR nor #651 caused it -- `_validate_latlon_ranges` runs its own MIN/MAX query, untouched by either. I checked all three branches to be sure.

Flagging it here because this PR is where the two CSV geometry paths get squared up on retaining no-geometry rows, so it is the natural place to notice that WKT-all-NULL now converts cleanly while lat/lon-all-NULL exits with a raw Python `TypeError`. A `None` guard plus the missing test would finish the symmetry; equally fine as a follow-up issue if you would rather keep this PR's scope tight.

cayetanobv:
All three verified and fixed in the latest commit.

**Stale warning — confirmed, and it was in both paths.** Reproduced your lat/lon case (warning claimed 1 row skipped, 3 rows written), and `_check_null_wkt_rows` made the same promise for the WKT path. Both now say the rows are "written with NULL geometry", matching the CLI help and the guide. Two tests assert the message no longer contains "skipped" while the row count stays at 3.

**Sentinel — dropped, your restructure was better.** Keeping the WKT column inside the CTE and excluding it in the outer `SELECT` gives the same predicate with no synthesized column, so the collision goes away rather than merely failing loudly. Verified with a CSV carrying a literal `__gpio_wkt_missing` column: converts cleanly now, and there is a test pinning it.

**All-NULL lat/lon — confirmed pre-existing, fixed here.** `Error: Conversion failed: '<' not supported between instances of 'NoneType' and 'int'` on this branch and identically on `main`. Since this PR is where the two CSV geometry paths get squared up, leaving WKT-all-NULL converting while lat/lon-all-NULL exits with a raw `TypeError` would have been the wrong asymmetry to ship. `_validate_latlon_ranges` returns early when the aggregates are NULL — there is no range to check — and `test_all_null_lat_lon_converts` sits next to its WKT counterpart.

Thanks for the empty-field spelling table; that was the hole I was most worried about, and quoted-empty mapping to NULL is what makes the `IS NULL` sentinel sound.

853 passed / 7 skipped on the csv/convert/hilbert/crs/curved/linearize/sort selection; pre-commit and xenon clean.

geoparquet_io/core/convert.py
```diff
@@ -522,7 +522,15 @@ def _validate_latlon_ranges(con, csv_read, lat_col, lon_col, verbose):
...
         if null_count > 0:
-            warn(f"⚠️  Warning: {null_count} rows have NULL lat/lon values and will be skipped")
+            warn(
+                f"⚠️  Warning: {null_count} rows have NULL lat/lon values and will be "
+                "written with NULL geometry"
+            )
+
+        # Every coordinate NULL: the aggregates are NULL, so there is no range
+        # to check — and comparing None would raise a bare TypeError (#655).
+        if min_lat is None or min_lon is None:
```
cayetanobv: This guard is wider than the crash it fixes, and it costs a validation that `main` performed.

`MIN`/`MAX` ignore NULLs, so `min_lon` is None only when *every* longitude is NULL. But the `return` then skips the **latitude** check too -- and latitude may have had perfectly measurable, out-of-range data:

```
# id,lat,lon
# 1,200,
# 2,300,
origin/main:    Error: Invalid parameter 'lat_column': invalid latitude values (range: 200.000000 to 300.000000)
this revision:  exit 0, 2 rows, 2 NULL geometries, silent about lat=200/300
```

The rows produce NULL geometry either way, so nothing bad is written -- but a garbage latitude column paired with a missing longitude column now converts silently where it used to be a clear error.

Guarding each axis on its own keeps the fix and restores the check:

```python
if min_lat is not None and (min_lat < -90 or max_lat > 90):
    raise InvalidParameterError(...)

if min_lon is not None and (min_lon < -180 or max_lon > 180):
    raise InvalidParameterError(...)
```

The `verbose` debug block below formats all four values with `:.6f`, so it needs the same treatment -- `if verbose and min_lat is not None and min_lon is not None:` -- since the early `return` is what currently protects it.

I patched this into your branch and re-ran my matrix:

| case | this revision | per-axis guard |
|---|---|---|
| all-NULL lat/lon | converts (the fix) | converts (fix preserved) |
| bad lat, lon all NULL | silent | **caught** |
| bad lon, lat all NULL | silent | **caught** (`main` crashed here) |
| ordinary out-of-range lat / lon | caught | caught |

`260 passed, 1 skipped` on the `csv or latlon or convert` selection either way -- which also means no existing test covers the middle two rows, so the coverage loss is currently invisible. Worth a test alongside the change.

cayetanobv (commented):
## Re-review: all three addressed, one new finding

**Recommendation: approve after the lat/lon guard is narrowed** -- a one-line change, details inline. (Comment review; GitHub blocks APPROVE on your own PR.)

### Verified

- **Both warnings reworded**, and you found a second one I had missed -- `_check_null_wkt_rows` had the same "will be skipped" claim on the WKT side. Good catch; I only checked the lat/lon path.
- **The sentinel is gone**, restructured exactly as suggested: `*` in the CTE, `EXCLUDE ({exclude_cols}, geometry)` outside, and a direct `WHERE {wkt_col} IS NULL OR geometry IS NOT NULL`. A CSV with a column literally named `__gpio_wkt_missing` now converts cleanly (2 rows, 1 NULL geometry) with and without `--skip-invalid`.
- **All-NULL lat/lon converts** instead of raising `TypeError`.
- **The new tests genuinely guard the fixes.** I mutated each one back:

  | mutation | result |
  |---|---|
  | restore `"will be skipped"` wording | both `TestWarningsMatchBehavior` tests FAIL |
  | remove the `None` guard | `TestAllNullLatLon` FAILS |

- **871 passed / 7 skipped** on the convert/csv/hilbert/crs/curved/linearize/sort/geopackage selection.

### The new finding, in short

`if min_lat is None or min_lon is None: return` is wider than the crash it fixes. `MIN`/`MAX` ignore NULLs, so `min_lon` is None only when *every* longitude is NULL -- but that also suppresses the **latitude** check, which still had data:

```
# id,lat,lon / 1,200, / 2,300,     (garbage latitudes, longitude column entirely empty)
origin/main:    Error: Invalid parameter 'lat_column': invalid latitude values
this revision:  exit 0, 2 rows, 2 NULL geometries, no mention of lat=200/300
```

A per-axis guard keeps both properties. I prototyped it and ran the numbers -- see the inline comment.

### One pre-existing thing, for the record

A CSV carrying a column literally named `geometry` alongside `--wkt-column geom` fails to convert:

```
Error: Invalid geometry data: Binder Error: No function matches the given name ...
```

Identical on `origin/main`, so nothing to do with this PR -- the CTE has always ended up with two `geometry` columns. Noting it only because I exercised it while testing the restructured query, in case it's worth its own issue.

Issue #665, opened 2026-08-21
Reduce test suite runtime and CI cost


cholmes:
The fast suite (3,771 tests) runs in about 97 seconds locally but costs about 85 CI job-minutes per push, and the default local invocation currently fails its own coverage gate. Profiling shows the local wall time is machine-bound (~700s of aggregate CPU across 12 workers), so the practical wins are in CI configuration, a handful of real time sinks, and the developer loop — not in fixture scoping or dist-mode tuning, which measured fine.

This issue tracks the measured optimizations. Each row below comes from a timed profile (two full runs, per-file serial timing, and CI log analysis), not estimates.

## Goals

- Cut CI cost by roughly 20 job-minutes per push and shorten the Windows critical path.
- Remove the fast suite's only flaky tests.
- Make plain local `pytest` runs fast and green.

## Deliverables

- [ ] Drop `--cov` from the 9 CI test jobs that never report coverage (only ubuntu-3.11 feeds Codecov and diff-cover). Apply the same change to the slow-suite jobs.
- [ ] Zero out retry backoff in the six `test_wfs.py` tests that each spend 6.15s in real `time.sleep` calls, and investigate why a mocked path retries at all.
- [ ] Move meta/tooling tests (codespell, commitizen, doc-sync, mutmut, validate-claude-md, mypy, security tool checks) out of the fast suite into a `meta` lane that runs in nightly/slow.
- [ ] Remove `--cov` from `addopts` in `pyproject.toml`; have CI pass coverage flags explicitly (it already does).
- [ ] Diagnose the local coverage-combine failure: the same suite reports 41% locally (macOS, `-n 12`) but ≥67% in CI, so every plain local run exits 1.
- [ ] Optional: validate `--dist=worksteal` plus targeted `xdist_group` marks on Windows instead of blanket `--dist=loadfile`.

## Non-goals

Measured negative findings — don't spend time here:

- `--import-mode=importlib`: collection is already 0.85s.
- Fixture scope promotion: no expensive function-scoped fixtures exist.
- Micro-test merging for speed: per-test overhead is 6–45ms.
- Local dist-mode tuning: two full runs were identical to 0.03s at machine saturation.

<details>
<summary><strong>Detailed plan and measurements</strong></summary>

### Baseline profile

| Measurement | Value |
|---|---|
| Fast suite, local, `-n auto` (12 workers) | 97.2s wall, ~700s user CPU, ~8× effective parallelism |
[...]

PR #683, opened 2026-08-21
ci: replace blanket loadfile with targeted xdist_group on Windows (experiment)

cholmes:
**Experiment — do not merge without a green Windows matrix and a timing comparison** (issue #665, item 6, the optional one).

## What the rationale research found

`--dist=loadfile` for the Windows fast-test job arrived in #167 (`358e588`) with the comment *"Use loadfile distribution on Windows to reduce DuckDB extension conflicts"*, and was copied to the slow-test job in `0cb159c`. The conflict it targeted is [duckdb#12589](https://github.com/duckdb/duckdb/issues/12589) — concurrent `INSTALL`s racing on the shared extension directory. The same PR also wrapped every `INSTALL` in `try/except` (now `core/duckdb_utils.py:474-500`).

So the intent was: keep a whole test file on one worker so its extension installs do not collide with another worker's.

**Measured, the premise does not hold.** I patched `duckdb.DuckDBPyConnection.execute` and recorded every `INSTALL ... FROM community` executed by the fast suite, together with the xdist worker that ran it:

| dist mode | community `INSTALL` calls | files involved | workers installing `a5` | `geography` | `h3` |
|---|---|---|---|---|---|
| `--dist=loadfile` (today) | 96 | 13 | **3** (gw1, gw2, gw3) | **2** | **3** |
| `--dist=loadgroup` + marks (this PR) | 96 | 13 | **1** (gw0) | **1** | **1** |

The installs are spread over 13 test files, so pinning each *file* to a worker never serialized them — with `-n 4` the files simply land on different workers and race anyway. The protection has been nominal since it was added.

Note also that CI's "Pre-install DuckDB extensions" step covers only `spatial`, `httpfs` and `h3`; `a5` and `geography` are genuinely downloaded during the test run, so they are the live race.

## What this PR changes

1. The 13 files that execute community-extension installs get a module-level `pytestmark = pytest.mark.xdist_group("duckdb-community-extensions")`.
2. The Windows fast-test job runs `--dist=loadgroup` instead of `--dist=loadfile`.

Under `loadgroup`, a test *without* an `xdist_group` mark has its full nodeid as its scope, i.e. it is its own work unit and schedules individually — while every marked test shares one work unit and therefore one worker. The result is strictly stronger isolation than `loadfile` (one installer instead of three) plus per-test load balancing for the other ~3,600 tests, which no longer queue behind 200-test files such as `test_wfs.py` (204 tests, 41.5s serial) or the corpus file.

### Why not `worksteal`, as the issue suggested

`xdist_group` is honoured **only** by the `loadgroup` scheduler. In pytest-xdist 3.8.0, `xdist/remote.py:243` applies group names only when `config.getvalue("loadgroup")` is true, and `WorkStealingScheduling` has no scope concept at all. "worksteal + targeted `xdist_group` marks" is not implementable — under `worksteal` the marks are inert, making it an all-or-nothing removal of the extension-race protection. `loadgroup` is the scheduler that expresses the intent, and it delivers the same fine-grained balancing for everything outside the group.

## Local measurements

macOS, `-n 4`, no coverage, same machine (shared with other jobs, so treat as indicative):

| dist mode | wall | result |
|---|---|---|
| `--dist=loadfile` | 150.9s | 3959 passed, 40 skipped |
| `--dist=loadgroup` + marks | 93.2s | 3959 passed, 40 skipped |

Grouping does not create a tail: the 13 grouped files are 416 tests / ~36s serial, against ~410s of total serial suite CPU.

Expected saving on the 12m18s Windows step: 1–3 minutes. **The Windows matrix in this PR is the actual validation** — if it is red or not faster, the recommendation is to close this and keep `loadfile`.

## Scope and merge order

The edit is deliberately confined to the Windows `--dist` flag plus the `xdist_group` marks. The Windows **slow**-test job keeps `--dist=loadfile` (it ignores the new marks, so it behaves exactly as before).

Two other open PRs touch `.github/workflows/tests.yml`:
- #677 (coverage flags on the test jobs)
- the meta-lane branch (moves repo-tooling tests out of the fast selection)

Both edit the same `Run fast tests with pytest (parallel)` block. This PR should merge **after** them and be rebased onto the result, so the `--dist` change lands on top of their final pytest invocation rather than conflicting with it.

https://claude.ai/code/session_015dJzA17TYiysJ19Y35E1Cx

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -157,8 +157,17 @@ jobs:
         shell: bash
         run: |
           if [ "$RUNNER_OS" == "Windows" ]; then
-            # Use loadfile distribution on Windows to reduce DuckDB extension conflicts
-            uv run pytest -n 4 --dist=loadfile -m "not slow and not network" -v --tb=short --durations=25 --durations-min=0.5 --cov=geoparquet_io --cov-report=term-missing --cov-report=xml
+            # EXPERIMENT (#665 item 6): loadgroup replaces the blanket loadfile.
+            # loadfile pinned every file to one worker to dodge concurrent DuckDB
+            # community-extension INSTALLs (duckdb#12589) -- but those INSTALLs are
+            # spread over 13 files, so loadfile never actually serialized them.
+            # loadgroup schedules per test, except for tests carrying an
+            # xdist_group mark, which share a worker. The 13 extension-touching
+            # files carry xdist_group("duckdb-community-extensions"), so exactly
+            # one worker ever INSTALLs -- stricter than loadfile -- while the
+            # remaining ~3.6k tests load-balance instead of queueing behind
+            # 200-test files.
+            uv run pytest -n 4 --dist=loadgroup -m "not slow and not network" -v --tb=short --durations=25 --durations-min=0.5 --cov=geoparquet_io --cov-report=term-missing --cov-report=xml
           elif [ "$RUNNER_OS" == "macOS" ]; then
             uv run pytest -n 3 -m "not slow and not network" -v --tb=short --durations=25 --durations-min=0.5 --cov=geoparquet_io --cov-report=term-missing --cov-report=xml
           else
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -196,6 +196,17 @@ This is the first beta release of geoparquet-io 1.0, featuring major new spatial
 
 ### Changed
 
+- **Windows CI distributes tests per test instead of per file (#665).** The
+  Windows fast-test job used `--dist=loadfile` to avoid concurrent DuckDB
+  community-extension `INSTALL`s racing on the shared extension directory
+  (duckdb#12589). Instrumenting the suite showed the extension installs are
+  spread over 13 test files, so `loadfile` never actually serialized them —
+  `a5` and `h3` were each installed by three workers concurrently. Those 13
+  files now carry `xdist_group("duckdb-community-extensions")` and the job runs
+  `--dist=loadgroup`, which pins exactly that group to one worker while
+  scheduling the other ~3.6k tests individually rather than queueing them
+  behind 200-test files.
+
 - **Coordinate/CRS mismatch heuristic downgraded to WARNING.** The
   `gpio check spec` heuristic that flags geographic-looking coordinates (values
   within ±180/±90) under a projected CRS now reports a WARNING instead of
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -196,6 +196,17 @@ This is the first beta release of geoparquet-io 1.0, featuring major new spatial
 
 ### Changed
 
+- **Windows CI distributes tests per test instead of per file (#665).** The
+  Windows fast-test job used `--dist=loadfile` to avoid concurrent DuckDB
+  community-extension `INSTALL`s racing on the shared extension directory
+  (duckdb#12589). Instrumenting the suite showed the extension installs are
+  spread over 13 test files, so `loadfile` never actually serialized them —
+  `a5` and `h3` were each installed by three workers concurrently. Those 13
+  files now carry `xdist_group("duckdb-community-extensions")` and the job runs
+  `--dist=loadgroup`, which pins exactly that group to one worker while
+  scheduling the other ~3.6k tests individually rather than queueing them
+  behind 200-test files.
+
 - **Coordinate/CRS mismatch heuristic downgraded to WARNING.** The
   `gpio check spec` heuristic that flags geographic-looking coordinates (values
   within ±180/±90) under a projected CRS now reports a WARNING instead of
--- a/tests/test_add.py
+++ b/tests/test_add.py
@@ -10,6 +10,15 @@ from click.testing import CliRunner
 
 from geoparquet_io.cli.main import add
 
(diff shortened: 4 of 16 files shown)


cholmes:
## Experiment result: green, but **no speedup** — hypothesis falsified

Windows matrix on `4d4c28d` ([run 32535508774](https://github.com/geoparquet/geoparquet-io/actions/runs/32535508774)), "Run fast tests with pytest (parallel)" step, against the average of two recent `main` runs (32526179913, 32524624049):

| Job | `main` baseline (avg) | This PR | Delta | Result |
|---|---|---|---|---|
| windows / 3.11 | 11m46s | 12m06s | +20s slower | pass |
| windows / 3.12 | 10m27s | 11m54s | **+1m27s slower** | pass |
| windows / 3.13 | 11m51s | 11m49s | -2s | pass |

**10/10 test jobs pass, 17/17 checks pass** — so the change is safe, and per-test scheduling
does not break Windows file locking. But the mean delta is **+35s slower**, against an
estimated 1–3 minute *saving*.

### Why

Windows CI runs `-n 4` on a 4-core runner. All four workers saturate the CPU, so wall time
is `total CPU work / 4` almost regardless of how tests are assigned — scheduling only buys
wall time when workers would otherwise idle. This is the same negative finding #665 already
recorded for local tuning ("two full runs were identical to 0.03s at machine saturation").

My local measurement (150.9s `loadfile` → 93.2s `loadgroup`) did **not** predict CI: that
machine has 12 cores, so at `-n 4` there is idle headroom for an unlucky file-to-worker
assignment to cost real wall time. It was the wrong benchmark for a saturated 4-core runner.

### Recommendation: keep `--dist=loadfile`, close #665 item 6

Not worth pursuing in any form. `worksteal` in particular cannot honour `xdist_group` at all
(see the PR description), so it would be an unconditional removal of the protection — and
there is now no speed prize to justify it. The real Windows savings are item 1 (drop `--cov`
from the nine non-reporting jobs) and item 3 (meta lane).

### The one finding worth keeping

`--dist=loadfile` **does not actually serialize DuckDB community-extension installs.** The 96
`INSTALL ... FROM community` calls are spread over 13 test files, so with `-n 4` those files
land on different workers: `a5` and `h3` are each installed by **three workers concurrently**
today, `geography` by two. The inline comment on that line promises isolation the flag does
not deliver.

In practice the race appears to be covered by the `try/except` in `core/duckdb_utils.py` and
the "Pre-install DuckDB extensions" step, and it has not been observed failing. So my
suggestion is to **close this PR** and instead fix the misleading comment, rather than pay
~35s per Windows job to make the isolation real.

Leaving this as a draft for the maintainer's call.

cholmes:
Closing without merging — this is a **negative result**, which is the outcome the experiment existed to determine.

## Outcome: green, but slower

All 17 checks passed (10/10 test jobs), so the change is safe. It is simply not faster. Windows fast-test pytest step, `4d4c28d` vs the average of two recent `main` runs (32526179913, 32524624049):

| Job | `main` baseline (avg) | This PR | Delta |
|---|---|---|---|
| windows / 3.11 | 11m46s | 12m06s | +20s slower |
| windows / 3.12 | 10m27s | 11m54s | **+1m27s slower** |
| windows / 3.13 | 11m51s | 11m49s | -2s |

Mean **+35s slower**, against an estimated 1–3 minute saving.

**Why:** Windows CI runs `-n 4` on a 4-core runner, so all four workers saturate the CPU and wall time is `total CPU work / 4` almost regardless of how tests are assigned. Scheduling only buys wall time when workers would otherwise idle — Windows is CPU-bound here, not scheduling-bound. (My local measurement showed -38%, but that machine has 12 cores and therefore idle headroom at `-n 4`; it was the wrong benchmark for a saturated runner.)

## Two findings worth keeping

1. **`--dist=worksteal` cannot honour `xdist_group` marks.** In pytest-xdist 3.8.0, `xdist/remote.py:241` applies group names only when `config.getvalue("loadgroup")` is true, and `WorkStealingScheduling` has no scope concept at all. "worksteal plus targeted `xdist_group` marks" — the shape #665 item 6 proposed — is not implementable; under worksteal the marks are inert, making it an unconditional removal of the protection.

2. **`--dist=loadfile` never actually serialized the DuckDB community-extension installs.** Instrumenting `DuckDBPyConnection.execute` across the fast suite recorded **96 `INSTALL ... FROM community` calls spread over 13 test files**. With `-n 4` those files land on different workers, so today `a5` and `h3` are each installed by **three workers concurrently** and `geography` by two — exactly the race `loadfile` was adopted to prevent. Whatever protection exists in practice comes from the `try/except` in `core/duckdb_utils.py:474-500` and the "Pre-install DuckDB extensions" CI step, not from the distribution mode.

## Recommendation

Keep `--dist=loadfile`. The real Windows savings are #665 item 1 (drop `--cov` from the nine non-reporting jobs) and item 3 (the meta lane).

Optional follow-up: the inline comment on the `loadfile` line in `tests.yml` ("Use loadfile distribution on Windows to reduce DuckDB extension conflicts") promises a serialization that finding 2 shows never happened, and could be corrected whenever that file is next touched.

cholmes:
Closing as a negative result — see the closeout comment above: 17/17 checks green but Windows averaged +35s slower, so we keep --dist=loadfile.

PR #747, opened 2026-08-28
fix(geometry): combine Arrow chunks before ST_IsValid to stop a segfault

### LINKED ISSUE #737: Segfault in repair_arrow_table_geometry: ST_IsValid over chunked Arrow table crashes DuckDB spatial
## Summary

`repair_arrow_table_geometry()` segfaults (SIGSEGV, exit 139) when the
Arrow table's geometry column is chunked. The crash happens in the
**count** query, before any repair runs, and it takes down the whole
process — the `except Exception` guard at the end of the function cannot
catch it.

Calling `.combine_chunks()` before registering the table fixes it.

## Impact

Any ArcGIS extraction large enough to produce a multi-chunk table dies
after downloading every feature. On a 559,077-feature layer that is about
6 minutes of paging discarded at the final step, with no output file and
no error message. `pq.read_table()` returns 280 chunks here, so the
failure is reached by ordinary use rather than an edge case.

The `finally` block deletes the temp parquet, so the downloaded data is
gone too. The caller sees only a silent non-zero exit.

## Reproduce

```python
import pyarrow.parquet as pq
from geoparquet_io.core.geometry_repair import repair_arrow_table_geometry

t = pq.read_table("land_use.parquet")   # 559,077 rows, 280 chunks
print(t.column("geometry").num_chunks)  # 280
repair_arrow_table_geometry(t, "geometry", repair=True)   # SIGSEGV
```

Reduced to the count query the function runs internally:

```python
con.register("_s", t)
p = 'TRY(ST_GeomFromWKB("geometry"))'
con.execute(
    f'SELECT COUNT(*) FROM _s '
    f'WHERE "geometry" IS NOT NULL AND {p} IS NOT NULL AND NOT ST_IsValid({p})'
).fetchone()          # SIGSEGV
```

A geometry-only table reproduces it, so no other column is involved.

## Evidence

Same table, same query, 3 runs each:

| Table | Runs | Result |
|---|---|---|
| 280 chunks (as `read_table` returns) | 3 | 3x exit 139 |
| `.combine_chunks()` -> 1 chunk | 3 | 3x exit 0, `invalid = 29` |

The crash is **non-deterministic at the margin**. Truncating to the first
54 chunks crashed on 2 of 3 runs and passed on the third, which points at
a memory-safety fault rather than one bad geometry. No individual chunk
crashes in isolation: chunk 139 alone returns cleanly, and chunks
`[140:280)` return `invalid = 15`. Crash probability rises with the number
of chunks in a single query.

Synthetic WKB (bowtie polygons) across 280 chunks does **not** crash, so
reproducing it needs real geometry.

## Suggested fix

Combine chunks before registering the table:

```python
 con = get_duckdb_connection(load_spatial=True, load_httpfs=False)
-con.register("_gpio_repair_src", table)
+con.register("_gpio_repair_src", table.combine_chunks())
```

This is a one-line change at the two `con.register` sites in
`repair_arrow_table_geometry`, and it costs one copy of the geometry
column.

Worth noting: the comment above `repair_expr` already documents an
earlier segfault in this extension and calls the current AND-form
"crash-free." That holds for the repair expression, but the count query
on the preceding line has the same class of bug. The comment reads as
broader protection than the code has.

## Environment

- geoparquet-io 1.3.0
- duckdb 1.5.1, spatial extension
- pyarrow 24.0.0
- shapely 2.1.2, GEOS 3.13.1
- Python 3.12.2, Linux 6.8.0-56-generic

## Source data

Philadelphia Land Use, City of Philadelphia ArcGIS FeatureServer,
559,077 polygons, EPSG:3857, 29 invalid geometries.
`https://services.arcgis.com/fLeGjb7u4uXqeF9q/arcgis/rest/services/Land_Use/FeatureServer/0`

cholmes:
Fixes #737.

## The bug

`ST_IsValid` over a **chunked** registered Arrow table crashes DuckDB 1.5.1's spatial extension. The crash lands in the invalid-geometry **count** — before any repair runs — so the `except Exception` guard at the end of `repair_arrow_table_geometry` cannot catch it and the process dies with SIGSEGV. The `finally` block then deletes the temp parquet, so an ArcGIS extraction loses every downloaded feature (6 minutes of paging for the reporter's 559k-feature layer) and the caller sees only a non-zero exit and no error message.

`pq.read_table()` returns hundreds of chunks for any sizable file — 280 in the report — so this is reached by ordinary use, not an edge case.

## Scope: two call sites, not one

The report names `repair_arrow_table_geometry`. Auditing the other Arrow-table registrations found one more path that runs the *same* `ST_IsValid` count over a registered table: `extract_table()` registers the caller's table and hands the query to `repair_query_geometry`, so the crash is also reachable from `gpio.read(...).extract(...)` and the streaming extract. Both are fixed; the remaining registration sites do not run `ST_IsValid` over their registered table.

## The fix

Both now register a single-chunk copy through a new `register_arrow_table_for_spatial()` helper in `core/duckdb_utils.py`. The helper exists rather than two copies of the workaround so the reason lives in one place and future registrations don't have to rediscover it. Combining costs one copy of the data; in both paths the caller already materializes the full table, so it fits the existing memory profile, and `repair_arrow_table_geometry` still hands back the *original* table when there's nothing to repair, so the copy is released on return.

## Tests

Memory corruption can't be provoked reliably in a test — it's non-deterministic at the margin (54 chunks crashed 2 of 3 runs), needs real geometry, and synthetic WKB across 280 chunks does not reproduce it. Following the precedent set for the #642 segfault in this same module, the regression guards assert the **invariant that avoids it**: what gets registered is a single chunk per column. Both guards were sabotage-checked — reverting either call site makes its guard fail.

Alongside them: chunked input is counted (32), repaired (0 invalid remaining), and keeps its schema metadata.

Fast suite green (3961 passed; the two `test_commitizen` failures are the known parallel-run flakes and pass serially). Pre-commit hooks pass.

https://claude.ai/code/session_0131ZaoMzLmembS7nUkz5sB3

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes when validating geometries in chunked Arrow tables.
  * Improved reliability of WFS, ArcGIS, BigQuery, Carto, Python API, and streaming extractions.
  * Ensured downloaded features are preserved when geometry validation encounters invalid data.
  * Preserved repaired table schemas and row counts for chunked inputs.

* **Tests**
  * Added coverage for chunked geometry repair, extraction, invalid-geometry counts, and metadata preservation.

* **Documentation**
  * Documented the segmentation fault fix in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -233,6 +233,21 @@ This is the first beta release of geoparquet-io 1.0, featuring major new spatial
 
 ### Fixed
 
+- **Segfault on chunked Arrow input to geometry repair (#737).** `ST_IsValid`
+  over a *chunked* registered Arrow table crashes DuckDB 1.5.1's spatial
+  extension. The crash lands in the invalid-geometry **count** — before any
+  repair runs — so the `except Exception` guard downstream cannot catch it and
+  the process dies with SIGSEGV; the `finally` block then deletes the temp
+  parquet, so an ArcGIS extraction loses every downloaded feature and the caller
+  sees only a non-zero exit. `pyarrow.parquet.read_table` returns hundreds of
+  chunks for any sizable file (280 for the reporter's 559k-row layer), so this
+  was reached by ordinary use rather than an edge case. Both paths that run
+  `ST_IsValid` over a registered Arrow table — `repair_arrow_table_geometry`
+  (WFS/ArcGIS/BigQuery/Carto extraction) and `extract_table` (the Python API and
+  streaming extract) — now register a single-chunk copy through the new
+  `register_arrow_table_for_spatial()` helper, which carries the reason so
+  future registrations do not have to rediscover it.
+
 - **Six internal DuckDB connections now route through the shared connection
   factory.** `benchmark_duckdb`, `get_file_info`, `wkb_to_wkt_preview`,
   `get_column_statistics`, `add country-codes`'s connection setup, and the
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -233,6 +233,21 @@ This is the first beta release of geoparquet-io 1.0, featuring major new spatial
 
 ### Fixed
 
+- **Segfault on chunked Arrow input to geometry repair (#737).** `ST_IsValid`
+  over a *chunked* registered Arrow table crashes DuckDB 1.5.1's spatial
+  extension. The crash lands in the invalid-geometry **count** — before any
+  repair runs — so the `except Exception` guard downstream cannot catch it and
+  the process dies with SIGSEGV; the `finally` block then deletes the temp
+  parquet, so an ArcGIS extraction loses every downloaded feature and the caller
+  sees only a non-zero exit. `pyarrow.parquet.read_table` returns hundreds of
+  chunks for any sizable file (280 for the reporter's 559k-row layer), so this
+  was reached by ordinary use rather than an edge case. Both paths that run
+  `ST_IsValid` over a registered Arrow table — `repair_arrow_table_geometry`
+  (WFS/ArcGIS/BigQuery/Carto extraction) and `extract_table` (the Python API and
+  streaming extract) — now register a single-chunk copy through the new
+  `register_arrow_table_for_spatial()` helper, which carries the reason so
+  future registrations do not have to rediscover it.
+
 - **Six internal DuckDB connections now route through the shared connection
   factory.** `benchmark_duckdb`, `get_file_info`, `wkb_to_wkt_preview`,
   `get_column_statistics`, `add country-codes`'s connection setup, and the
--- a/geoparquet_io/core/duckdb_utils.py
+++ b/geoparquet_io/core/duckdb_utils.py
@@ -637,6 +637,30 @@ class _DuckDBFieldWrapper:
         self.name = col_info.get("name", "")
 
 
+def register_arrow_table_for_spatial(con, name: str, table):
+    """Register an Arrow table under ``name`` for querying with spatial functions.
+
+    ``ST_IsValid`` over a *chunked* registered Arrow table segfaults DuckDB
+    1.5.1's spatial extension (issue #737). It is a memory-safety fault, not one
+    bad geometry: crash probability rises with the chunk count, and no single
+    chunk reproduces it in isolation. A SIGSEGV takes the process down, so no
+    ``except`` guard downstream can contain it — and `pyarrow.parquet.read_table`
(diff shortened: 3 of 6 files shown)


cholmes:
Closing this: adversarial review plus a direct reproduction says `combine_chunks()` does not prevent the crash it targets, and the real fix is the DuckDB pin.

## What reproduces

200k rows of WKB polygons with NULLs interleaved, running the issue's reduced count query, three runs per configuration:

| DuckDB | chunks | exit codes |
|---|---|---|
| 1.5.1 | 300 | **139, 139, 139** (SIGSEGV) |
| 1.5.1 | **1** — i.e. exactly what this PR's `combine_chunks()` produces | **139, 139, 139** |
| 1.5.5 | 300 | 0, 0, 0 |

A single chunk crashes just as reliably as 300. Chunking is not the variable; the DuckDB version is.

Independently, a review agent could not reproduce a chunk-count effect at all — across 223k real ag-field polygons and 106k features pulled from the reporter's own Philadelphia Land Use FeatureServer, at 3 / 281 / 1000 / 2002 chunks, with and without injected NULL geometry, against both the reduced query and the layered query `repair_arrow_table_geometry` actually runs. It isolated the trigger instead: `TRY()` over a blob under conditional execution, with the same expression minus `TRY()` running clean at 281 chunks.

That is the bug `pyproject.toml` already documents — duckdb/duckdb-spatial#858, dropped in DuckDB 1.5.2 — and the reason `core/geometry_repair.py` counts invalid geometry in layers (#645).

## Cost, for the record

Measured peak RSS on a 223,375-row / 396.5 MB Arrow table (geometry only 52.4 MB of it):

| path | before | with this PR |
|---|---|---|
| `repair_arrow_table_geometry` | 0.0 MB | 313.4 MB |
| `extract_table` | 378.1 MB | 588.5 MB (+56%) |

87% of the copied bytes are non-geometry columns the `ST_IsValid` count never reads. `combine_chunks()` also silently does not deliver its invariant above 2 GiB: on a `binary` column it returns a still-multi-chunk table rather than raising, so the largest inputs pay the copy and stay chunked.

## Where the fix goes

DuckDB 1.5.2+ fixes it. gpio pins `<1.5.2` only because the `geography` community extension — needed by `gpio add s2` and `gpio partition s2` — is published for 1.5.0 and 1.5.1 and returns HTTP 404 for 1.5.2 through 1.5.5 on every platform. I confirmed that both by probing `community-extensions.duckdb.org` and by installing DuckDB 1.5.5 and running `INSTALL geography FROM community`.

So the path forward is to unblock that extension, not to copy tables. Details and the reproduction script go on #737, which stays open.

Issue #721, opened 2026-08-23
Windows: pyarrow-written native GEOMETRY columns carry all-zero geospatial statistics


cholmes:
## What

On `windows-latest` (Python 3.11, 3.12 and 3.13 alike), a Parquet file gpio writes with a **native GEOMETRY logical type via the pyarrow writer** comes back with geospatial statistics of `[0.00, 0.00, 0.00, 0.00]`, while the geometries themselves are elsewhere. `gpio check spec` then correctly reports:

```
native_geo_stats_contains_data_geometry: 2 of 2 geometries fall outside geospatial statistics
```

Note `native_geo_stats_geometry` **passes** — the statistics are present, they are just all zeros.

## Evidence

Two independent PRs added tests that validate native-geo output and both fail on Windows only, on the same check:

- #702 `test_pgo_validates_against_target_version_oracle` (parquet-geo-only via `write_geoparquet_table`) — [job](https://github.com/geoparquet/geoparquet-io/actions/runs/32624053670/job/97156583777)
- #707 `TestSecondaryGeometryColumns::test_20_multi_geometry_validates_clean` and `..._with_geoarrow_imported` (GeoParquet 2.0 via the streaming strategy) — [job](https://github.com/geoparquet/geoparquet-io/actions/runs/32624053670/job/97156128507); both geometry columns are affected, so it is not specific to the primary.

The same write path on macOS produces correct statistics:

```
pyarrow 24.0.0 | duckdb 1.5.1
geo_bbox: ('geometry', {'xmin': -122.4, 'xmax': 151.2, 'ymin': -33.9, 'ymax': 37.8, ...})
```

Versions are pinned in `uv.lock`, so Windows CI runs the same pyarrow and DuckDB — this is a platform/wheel difference, not a version skew.

## Scope

Only the **pyarrow** writer path is implicated. #714 also writes native 2.0 output and is green on Windows; it goes through the DuckDB `COPY` strategy. That makes the likely culprit pyarrow's geospatial-statistics support in the Windows wheel rather than anything in gpio, but it is worth confirming from which side the zeros come (pyarrow writing them, or DuckDB's `parquet_metadata()` reading them) before deciding where a fix or a workaround belongs.

## Impact

Files gpio writes on Windows with native geometry types ship statistics that do not describe their data. Any reader using them for predicate pushdown would skip row groups that actually match — silently wrong query results, not an error.

## Interim handling

[...]

PR #748, opened 2026-08-28
test(geo-stats): record that Windows geo-stat zeros are DuckDB's read, not pyarrow's write

cholmes:
Refs #721. **This is the investigation step the issue asks for, not a fix** — read on for why that is the right first move here.

## The question the issue poses

> it is worth confirming from which side the zeros come (pyarrow writing them, or DuckDB's `parquet_metadata()` reading them) before deciding where a fix or a workaround belongs

That cannot be answered on a machine where the bug does not reproduce — on macOS both readers agree exactly:

```
pyarrow  is_geo_stats_set: True   xmin -122.4  ymin -33.9  xmax 151.2  ymax 48.85
duckdb   geo_bbox:               xmin -122.4  ymin -33.9  xmax 151.2  ymax 48.85
```

But CI already runs `windows-latest`. So this PR asks the question there.

## What it adds

`tests/test_native_geo_statistics.py` writes a native-GEOMETRY file at coordinates far from the origin in every direction (so all-zero statistics cannot contain the data by accident, and even a single zeroed bound is caught), then reads the **same file** through both readers and asserts each one **separately**, across both implicated pyarrow entry points:

- `write_geoparquet_table` → `pq.write_table` (the #702 path)
- the arrow-streaming write strategy → `pq.ParquetWriter` (the #707 path)

On Windows the assertions are non-strict `xfail`, so CI stays green while recording the answer:

| Windows result | Meaning | Where the fix belongs |
|---|---|---|
| only `via_duckdb` xfails | pyarrow wrote correct stats, DuckDB misreads them | gpio's stats-reading path |
| both xfail | pyarrow wrote zeros | upstream, or route Windows native-geo writes through DuckDB COPY |
| both xpass | the platform bug is gone | delete the markers |

A third test asserts the two readers *agree* — whichever side is wrong, disagreement is itself the signal. A fourth pins the containment predicate against `[0, 0, 0, 0]`, so the guard cannot rot into one that passes on the very shape it exists to catch.

## Why not just fix it

Both candidate workarounds are wrong if the diagnosis is wrong. Routing Windows writes through DuckDB COPY does nothing if the files were always fine and DuckDB misreads them. Verifying stats after each write via DuckDB would false-alarm on every Windows write in exactly that case. Guessing here buys a change that is either unnecessary or actively misleading, so this lands the discriminator first and the fix follows once CI names the culprit.

## Verification

9 tests pass on macOS (both write paths × both readers, plus the premise and predicate guards). Fast suite green — remaining failures are the known `test_commitizen`/`test_mutmut` parallel-run flakes, which pass serially. Pre-commit hooks pass.

Test-only, so no CHANGELOG entry.

https://claude.ai/code/session_0131ZaoMzLmembS7nUkz5sB3

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected documentation describing Windows geospatial statistics, clarifying that reported zero values come from metadata reading rather than file creation.
  - Documented that generated files contain correct native geometry statistics.

- **Tests**
  - Added comprehensive coverage for native geospatial statistics across supported readers and writing workflows.
  - Added validation for geometry metadata, geographic bounds, cross-reader consistency, and Windows-specific behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -244,9 +244,10 @@ This is the first beta release of geoparquet-io 1.0, featuring major new spatial
   xfailed: auto-version resolution splitting four ways on a native-geo input
   (#600), and the validator's `crs_valid` check flipping with geoarrow
   registration (the #603 family). The oracle also excuses
-  `native_geo_stats_contains_data` on Windows only, where pyarrow's wheel
-  writes all-zero geospatial statistics for native GEOMETRY columns (#721);
-  every other oracle check stays enforced on every platform.
+  `native_geo_stats_contains_data` on Windows only, where DuckDB's
+  `parquet_metadata()` misreads native GEOMETRY geospatial statistics as all
+  zeros (#721 — the file on disk is correct; see below); every other oracle
+  check stays enforced on every platform.
 
 - **`gpio pmtiles pyramid` (#570)**: bake an aggregate and its overview levels
   into a single zoom-banded PMTiles archive. Each level is tiled once with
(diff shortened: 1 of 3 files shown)


cholmes:
## The discriminator answered — and it inverts the hypothesis

Windows CI result, identical across Python 3.11/3.12/3.13 and both pyarrow write paths:

- `..._via_pyarrow[table-writer]` → **XPASS**
- `..._via_pyarrow[streaming-strategy]` → **XPASS**
- `..._via_duckdb[table-writer]` → **XFAIL**
- `..._via_duckdb[streaming-strategy]` → **XFAIL**
- `both_readers_agree_on_the_statistics` (both) → **XFAIL**

Per the table in the PR description, that is the "only `via_duckdb` xfails" row: **pyarrow writes correct geospatial statistics on Windows; DuckDB's `parquet_metadata()` misreads them from the same file.**

So the files gpio writes on Windows were never wrong — gpio's *reading* of them is. `get_aggregated_native_geo_stats` and `get_per_row_group_native_geo_stats` both go through DuckDB `parquet_metadata()`, which is why `gpio check spec` misreports good files there.

Details and the suggested fix are on the issue: https://github.com/geoparquet/geoparquet-io/issues/721#issuecomment-5455935497

This PR is still worth merging as-is — it is the regression guard, and the xfail markers now document a *known reader* bug rather than an open question. They stay non-strict, so they flip to XPASS on their own once the read path moves to pyarrow (or DuckDB fixes it), which is the signal to delete them.

PR #766, opened 2026-08-29
feat(sort): add Sort-Tile-Recursive ordering

oakhill87:
## Summary

- add Sort-Tile-Recursive (STR) as `gpio sort str`, `ops.sort_str()`, and `Table.sort_str()`
- pack geometry-envelope centers into row-group-sized X strips and alternating Y runs, with deterministic input-order tie breaking
- preserve GeoParquet metadata, streaming/file modes, bbox generation, output-version handling, and empty/null geometry behavior
- document when to use STR versus Hilbert and keep the CLI/API parity contracts and generated command references in sync

The implementation follows the algorithm in the linked spatial-sort presentation/benchmark: compute the number of row-group tiles, create `ceil(sqrt(tile_count))` X strips, then order each strip on Y. The Y direction alternates between strips to keep adjacent runs close.

## Real AIS verification

I converted one 518 MB partition from `ais_databricks` (5 July 2025): 22,774,317 positions, written as 227 row groups with a 100,000-row target. Both outputs passed all 24 `gpio check spec` validations, including a scan of every WKB value.

| Metric | Hilbert | STR | Change |
|---|---:|---:|---:|
| Output size | 508.4 MB | 448.6 MB | 11.8% smaller |
| Summed row-group bbox area | 31,655.7 | 15,086.6 | 52.3% lower |
| Pairwise bbox overlap area | 3,587.7 | 162.7 | 95.5% lower |
| Overlap / summed area | 11.3% | 1.1% | 10.5x lower |
| Oslofjord candidate groups | 18 | 15 | 16.7% fewer |
| Lofoten candidate groups | 40 | 36 | 10.0% fewer |
| Sort/write time on this machine | 85.9 s | 231.5 s | STR is 2.7x slower |

The tradeoff on this distribution is clear: STR costs more to build, but produces substantially tighter and less-overlapping row-group bounds.

### Hilbert — Oslofjord

18 / 227 candidate row groups, 1.81M candidate rows.

Hilbert row-group bounding boxes over Oslofjord

### Sort-Tile-Recursive — Oslofjord

15 / 227 candidate row groups, 1.51M candidate rows.

Sort-Tile-Recursive row-group bounding boxes over Oslofjord

[Open both full-resolution screenshots on Imgur](https://imgur.com/a/Hh9SEYH).

## Tests

- `pre-commit run --all-files`
- `pytest -q -n auto -m 'not slow and not network and not meta'`
  - 4,694 passed, 64 skipped, 6 expected xfailed
- focused CLI/API/STR regression surface
  - 315 passed
- executable `docs/guide/sort.md` examples
  - 9 passed, 5 skipped
- synthetic shuffled 460,800-point grid
  - STR produced zero pairwise row-group bbox overlap versus a 30.8% overlap ratio for Hilbert

## Summary by CodeRabbit

- **New Features**
  - Added Sort-Tile-Recursive (STR) spatial ordering through `gpio sort str`, `ops.sort_str`, and `Table.sort_str`.
  - Supports configurable strip sizing, geometry columns, row groups, compression, bounding boxes, and streaming workflows.
  - Improves spatial locality and places empty or null geometries last.

- **Documentation**
  - Added CLI, Python API, guide, reference, and changelog documentation.

- **Tests**
  - Added coverage for ordering quality, streaming, validation, metadata, and CLI/API consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ long run of write-path, metadata and CRS correctness fixes.
 
 ### Added
 
+- **Sort-Tile-Recursive spatial ordering.** `gpio sort str`, `ops.sort_str`, and `Table.sort_str` pack geometry-envelope centers into row-group-sized X/Y tiles as an alternative to Hilbert ordering.
 - **The guides' examples now run as tests.** Every fenced bash/python block in `docs/guide/*.md` runs verbatim as a pytest item, which found and fixed real bugs in the docs. ([#667](https://github.com/geoparquet/geoparquet-io/issues/667), [#732](https://github.com/geoparquet/geoparquet-io/issues/732))
 - **Honest notebook CI signal.** The `notebooks` job no longer reports green for notebooks that execute no real `gpio` call. ([#667](https://github.com/geoparquet/geoparquet-io/issues/667), [#720](https://github.com/geoparquet/geoparquet-io/issues/720))
 - **Canonical sample dataset.** `tests/data/canonical/` holds one small, spec-clean dataset (~405 KB) shared by the docs examples, e2e journeys and notebooks. ([#667](https://github.com/geoparquet/geoparquet-io/issues/667), [#719](https://github.com/geoparquet/geoparquet-io/issues/719))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ long run of write-path, metadata and CRS correctness fixes.
 
 ### Added
 
-- **Sort-Tile-Recursive spatial ordering.** `gpio sort str`, `ops.sort_str`, and `Table.sort_str` pack geometry-envelope centers into row-group-sized X/Y tiles as an alternative to Hilbert ordering.
+- **Sort-Tile-Recursive spatial ordering.** `gpio sort str`, `ops.sort_str` and `Table.sort_str` sort geometry-envelope centers into X strips and sort each strip on Y, alternating direction between strips, as an alternative to Hilbert ordering. ([#766](https://github.com/geoparquet/geoparquet-io/issues/766))
 - **The guides' examples now run as tests.** Every fenced bash/python block in `docs/guide/*.md` runs verbatim as a pytest item, which found and fixed real bugs in the docs. ([#667](https://github.com/geoparquet/geoparquet-io/issues/667), [#732](https://github.com/geoparquet/geoparquet-io/issues/732))
 - **Honest notebook CI signal.** The `notebooks` job no longer reports green for notebooks that execute no real `gpio` call. ([#667](https://github.com/geoparquet/geoparquet-io/issues/667), [#720](https://github.com/geoparquet/geoparquet-io/issues/720))
 - **Canonical sample dataset.** `tests/data/canonical/` holds one small, spec-clean dataset (~405 KB) shared by the docs examples, e2e journeys and notebooks. ([#667](https://github.com/geoparquet/geoparquet-io/issues/667), [#719](https://github.com/geoparquet/geoparquet-io/issues/719))
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -66,7 +66,7 @@ geoparquet_io/
 | `gpio process` | aggregate, overview | Transform or reduce GeoParquet data (aggregate, overview, |
 | `gpio publish` | stac, upload | Commands for publishing GeoParquet data (STAC metadata, cloud uploads) |
 | `gpio skills` |  | List and access LLM skills for gpio |
-| `gpio sort` | column, hilbert, quadkey | Commands for sorting GeoParquet files |
+| `gpio sort` | column, hilbert, quadkey, str | Commands for sorting GeoParquet files |
 <!-- END GENERATED: cli-commands -->
 
 <!-- BEGIN GENERATED: core-modules -->
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -24,6 +24,7 @@ long run of write-path, metadata and CRS correctness fixes.
 
 ### Added
 
+- **Sort-Tile-Recursive spatial ordering.** `gpio sort str`, `ops.sort_str`, and `Table.sort_str` pack geometry-envelope centers into row-group-sized X/Y tiles as an alternative to Hilbert ordering.
 - **The guides' examples now run as tests.** Every fenced bash/python block in `docs/guide/*.md` runs verbatim as a pytest item, which found and fixed real bugs in the docs. ([#667](https://github.com/geoparquet/geoparquet-io/issues/667), [#732](https://github.com/geoparquet/geoparquet-io/issues/732))
 - **Honest notebook CI signal.** The `notebooks` job no longer reports green for notebooks that execute no real `gpio` call. ([#667](https://github.com/geoparquet/geoparquet-io/issues/667), [#720](https://github.com/geoparquet/geoparquet-io/issues/720))
 - **Canonical sample dataset.** `tests/data/canonical/` holds one small, spec-clean dataset (~405 KB) shared by the docs examples, e2e journeys and notebooks. ([#667](https://github.com/geoparquet/geoparquet-io/issues/667), [#719](https://github.com/geoparquet/geoparquet-io/issues/719))
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -24,7 +24,7 @@ long run of write-path, metadata and CRS correctness fixes.
 
 ### Added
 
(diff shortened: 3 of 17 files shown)


cholmes:
Thanks for the contribution! I'm pushing towards a release today so I'm going to do a review but then just fix things instead of going back and forth, so we can try to get this in for 1.4.0

cholmes:
Thanks for this — the AIS verification with the before/after row-group bbox screenshots is a genuinely
useful way to argue for a spatial sort, and the tradeoff table makes the cost of STR explicit rather
than burying it.

Heads-up on process: we're running this through a full review now, and rather than send you a long
list, **we plan to push the fixes directly to `feat/sort-tile-recursive`** (the PR has
"allow edits from maintainers" enabled). That covers the failing `lint` job — ruff's isort wants
`import io` before `import json` in `tests/test_str_order.py` — plus whatever the review turns up,
including the two streaming-path issues the automated reviewer already flagged:

- `--add-bbox` isn't threaded through the streaming dispatch, so `gpio sort str - out.parquet --add-bbox`
  may silently omit the bbox column;
- the geometry column may stay hardcoded `"geometry"` in streaming mode instead of being resolved
  from the stream's `geo` metadata, which would break a file whose primary column is named something
  else.

Nothing is pushed yet, and we'll summarise every change we make so you can see exactly what moved and
why. If you'd rather make the changes yourself, say so and we'll send the findings as review comments
instead — just let us know before we push.

cholmes:
Pushed to `feat/sort-tile-recursive` as `3701219`. Here is everything that changed and why — all
of it verified by running it, and two of the numbers below came out differently from the review
that prompted them, so I've quoted my own measurements rather than the reviewer's.

The algorithm itself came through review clean: permutation-checked across 13 shapes (n=0,
all-identical coordinates, NaN/inf, mixed null/EMPTY) with no rows dropped or duplicated,
deterministic across runs, byte-identical through the file write path even though `duckdb-kv`
sets `preserve_insertion_order = false`, and the boustrophedon flip verified per-strip. The
quoting is correct at every one of the 20+ interpolation sites — hostile column names
(`geo"m`, `geo'm`, `g"),1 AS pwn,(1`) and an `o'brien` output path all round-trip through both
the CLI and the Python API with no injection and no double-escape. The uuid-suffixed internal
columns are better than what the rest of the codebase does for the same problem.

## CI (both were red)

- **ruff** — `import io` now precedes `import json` in `tests/test_str_order.py`.
- **mypy** — deleted the trailing `if __name__ == "__main__": str_order()`. It fails
  `[call-arg]` and also `tests/test_mypy.py` in the meta lane. It looks safe because
  `hilbert_order.py` has the same block, but that module is exempt via
  `[[tool.mypy.overrides]]` and `str_order` isn't. The block is dead anyway — calling it raises
  `TypeError`. I did **not** add a mypy override.

## The docs described behaviour the code does not have

This is the biggest change, and it is documentation, not code. The guide, `docs/cli/sort.md` and
the CLI help each claimed `--row-group-size` controls "both the spatial tile capacity and the
output row-group target", that "both default to 100,000 rows", and that STR "aligns spatial tiles
to Parquet row groups". Measured:

- with no flag the writer emits DuckDB's **122,880**-row groups while `DEFAULT_STR_TILE_SIZE` is
  100,000 — the default invocation never aligns;
- the writer rounds up to a multiple of 2048, so `--row-group-size 100000` writes 100,352;
- `tile_size` only reaches the ordering through `strip_count = ceil(sqrt(ceil(N / tile_size)))`.
  There is no per-tile packing — within a strip, rows are simply sorted on Y. On 20,000 points,
  `--row-group-size 800` and `1000` produce a **byte-identical** ordering, as do 1,500 and 2,000.
  And on 2M rows a "perfectly aligned" `--row-group-size 122880` produced byte-identically the
  same row-group bboxes as the misaligned default (`sum_area=68709.6` both).

So the docs now say what `--row-group-size` actually does, and keep the benefit claim to what I
could measure: on 2M uniform points with `--row-group-size 100000`, STR's mean row-group bbox
area was **3,846 deg² against Hilbert's 4,426 — about 13% tighter**. The guide gains a
"What --row-group-size does here" and a "How much does it help?" section.

## Behaviour fixes

- **`strip_size` now rounds to a whole number of tiles** (`tile_size * ceil(tile_count /
  strip_count)`). Previously a fractional tile at each strip boundary meant the row group
  spanning it covered two strips and its bbox stretched across the entire X extent — I
  reproduced one at `x=179.83` on a 360°-wide domain. With `--row-group-size` a multiple of
  2048: straddling groups eliminated, mean bbox area 6–11% smaller (200k rows @10240: 3595.6 →
  3239.2, 4 straddling → 0). With non-multiples it is neutral, ±3% noise, because written groups
  never line up with tiles regardless.
- **`--verbose`'s only new output was dead.** `str_order()` never called `configure_verbose()`,
  so both `debug("STR layout: …")` lines never emitted. `hilbert` gets it by accident via
  `get_dataset_bounds()`; `sort_by_column` does it explicitly. Now it does too.
- **The passthrough path claimed it sorted when it didn't** — on all-empty input it printed both
  "Writing file without STR ordering." and "Successfully reordered data using STR". Now emits
  "Wrote data without STR ordering to: …", matching `hilbert`.
- **`--row-group-size 0`** reported `Invalid parameter 'tile_size'`, naming an internal that is
  not a CLI option. Validated against the flag the user typed.
- **Streaming silently ignored `--add-bbox`.** `sort hilbert` has the identical gap, so rather
  than implement bbox in streaming here it now warns. (I wrote "no bbox column will be **added**"
  rather than "written" — a live pipe check showed the output *does* carry a `bbox` column
  inherited from the input, so "written" would have been a second false statement.)
- **Streaming ignored the stream's own `geo.primary_column`**, so a file whose primary column is
  `footprint` binder-errored in stream mode while working in file mode. It now reads
  `primary_column` from the metadata before falling back to `STANDARD_GEOMETRY_NAMES`.
- `api/table.py` hardcoded `tile_size = 100_000` while `ops.py` used `DEFAULT_STR_TILE_SIZE`;
  now imports the constant, with a note that the parity harness structurally cannot catch that
  drift because the CLI has no `tile_size` option.
- The 12 positional arguments into `str_order()` are now keywords, so inserting a parameter
  cannot silently shift `compression` / `compression_level` / `row_group_size_mb` /
  `row_group_rows` / `profile`. (`sort column` and `sort quadkey` already do this.)

## Tests

- `test_str_order_reduces_tile_bbox_area_for_adversarial_input` **did not discriminate STR from a
  plain `ORDER BY x, y`**: on a full 8×8 grid, lexicographic order gives tiles of constant x, so
  bbox area is 0 and the `< 0.25 × scattered` threshold passes trivially. Worth knowing —
  **total tile bbox area cannot separate these orderings at all.** On the new fixture
  lexicographic scores *marginally better* (358.6 vs 356.8), because any ordering that partitions
  uniform data approaches the domain area. Replaced with a candidate-tile metric — mean tiles
  intersected by a query window swept over the extent, which is what pushdown pays for:
  **STR 1.58, `ORDER BY x, y` 2.27, unsorted 19.65**. It also asserts the area fact, so the
  misleading metric can't quietly come back.
- Added a file-path test asserting the written row order matches `str_order_table`'s in-memory
  result — previously every file-path test asserted only `num_rows`, so a bug losing the ordering
  between the sort and the writer would have passed all of them.
- The internal-column-collision test appended `__str_input_order`, but the real names are
  `__gpio_str_input_order_<uuid4hex>` — so it never exercised the uuid suffixing it exists to
  protect. Fixed.
- Plus tests for the verbose output, the honest passthrough message, the `--row-group-size`
  error naming, and the non-standard streaming primary column.

`skills/geoparquet.md` gained an STR entry in its hand-written sorting section (only the
generated table row had been updated, and that file is what LLM agents read), and the CHANGELOG
entry gained its `([#766])` link and lost the same alignment overclaim.

## Verification

`pre-commit run --all-files` clean including the pre-push stage (mypy, xenon, import-linter,
vulture, deptry); 4,701 fast tests passing; `diff-cover` **98%** with `str_order.py`,
`api/ops.py` and `api/table.py` all at 100%.

## Two things I deliberately did not touch — filed instead

- **#776** — about a third of `str_order.py` is a near-clone of `hilbert_order.py`, and it
  imports three private names from it. Extracting `core/sort_common.py` rewrites `sort hilbert`
  too, so it belongs in its own PR. There is a comment at the import site pointing there. That
  issue also carries the two gaps both commands share (streaming `--add-bbox`, and streaming
  dispatch happening before the overwrite guard).
- **#777** — the str:hilbert time ratio grows 1.19x → 2.86x from 50k to 8M rows, caused by three
  full-dataset materializations carrying the WKB payload. The fix (rank over
  `(input_order, cx, cy)` and join the payload back) is a rewrite of your algorithm's SQL, so it
  should be your call, not ours. That issue also explains why it is the *same* work as real
  per-tile packing.

Neither is a blocker for this PR. Thanks again — the AIS verification with the before/after
row-group screenshots is a genuinely good way to argue for a spatial sort, and wiring the command
into **both** parity harnesses with written-out rationales for the known divergences is more
codebase literacy than most external PRs bring.

PR #770, opened 2026-08-30
fix(metadata): judge geometries against the whole file's geo statistics (#721)

cholmes:
Refs #721. **Scope narrowed** — see [the CI analysis below](https://github.com/geoparquet/geoparquet-io/pull/770#issuecomment-5467742839).

## What this fixes

`native_geo_stats_contains_data_*` compared a sample spanning the **whole file** against the bounds
of whichever row group came **first**, so any correct multi-row-group native-geo file failed it — on
every platform, not just Windows. It now compares against the union over all row groups
(min-of-mins / max-of-maxes, skipping groups that carry no statistics rather than folding in a
zero), and `native_geo_stats_*` reports that same file-wide bbox.

Also in scope, all platform-independent:

- a read failure no longer reports as *"column not found"* — `None` now means only that, and a
  raised error surfaces with its real message;
- `get_native_geo_statistics` takes the bbox and `geometry_types` from the **same** row group,
  rather than the bbox from the first chunk with statistics and the types from chunk 0;
- `_escape_sql_string` on the `path_in_schema` literal keeps its regression coverage.

## What this no longer does, and why

Earlier revisions routed local reads to pyarrow and deleted the `win32` validator exclusions, on the
premise that pyarrow reads these statistics correctly and DuckDB is the broken reader on Windows.
**CI refuted that premise on this branch** (run
[33299322424](https://github.com/geoparquet/geoparquet-io/actions/runs/33299322424), identical on
windows-latest × 3.11/3.12/3.13):

```
test_native_geo_stat_getters_return_correct_values          (local -> pyarrow)
  assert [0.0, 0.0, 1.0435e-320, 1.0435e-320] == approx([0.0, 0.0, 3.0, 3.0])

test_native_geo_stat_getters_return_correct_values_over_duckdb
  [XPASS(strict)] -- the win32 xfail expecting DuckDB to misread did NOT fail
```

That fixture is **DuckDB**-written. On it, DuckDB reads its own statistics correctly and pyarrow
returns a denormal. The #748 discriminator used **pyarrow**-written files, where the polarity is
reversed.

**On Windows each reader misreads the statistics the other's writer produced.** It is a
writer/reader interop bug, not one bad reader — so no reader swap fixes it, and the `win32`
exclusions are load-bearing on both sides. They are restored here, with their comments corrected to
say so, and the existing #721/#748 changelog entry is corrected in place.

The interop bug gets its own investigation: both readers against files from **both** writers, then
an upstream report to whichever side is at fault.

## Tests

`tests/test_native_geo_statistics.py` keeps the union coverage. Sabotage-verified: reducing the
union to the first row group reproduces the original failure verbatim —
`2 of 3 geometries fall outside geospatial statistics`.

Targeted suite: 326 passed, 1 skipped, 2 xfailed. Full fast suite green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation now evaluates native geospatial bounds across all row groups instead of only the first.
  * Improved reporting when geometry statistics are unavailable, unreadable, or contain only empty geometries.
  * Retained Windows-specific exclusions for known native geospatial statistics interoperability limitations.

* **Documentation**
  * Added guidance for validating native geospatial statistics in `GEOMETRY` and `GEOGRAPHY` columns.
  * Updated the changelog with corrected Windows behavior and file-wide validation details.


cholmes:
## Windows CI contradicts this PR's premise — narrowing the scope

This PR argues that pyarrow reads native geospatial statistics correctly and DuckDB is the broken
reader on Windows, and on that basis routes local reads to pyarrow and deletes the `win32`
exclusions. The CI run for the current head ([33299322424](https://github.com/geoparquet/geoparquet-io/actions/runs/33299322424))
says the picture is more complicated. Identical on windows-latest × 3.11/3.12/3.13, so not flake:

```
test_native_geo_stat_getters_return_correct_values          (local -> pyarrow after this PR)
  assert [0.0, 0.0, 1.0435e-320, 1.0435e-320] == approx([0.0, 0.0, 3.0, 3.0])

test_native_geo_stat_getters_return_correct_values_over_duckdb
  [XPASS(strict)] -- the win32 xfail expecting DuckDB to misread did NOT fail
```

The fixture there (`_native_geo_fixture`) is written by **DuckDB**
(`write_parquet_with_metadata`, `parquet-geo-only`). On that file DuckDB reads its own statistics
correctly, and **pyarrow** returns `1.0435e-320` — a denormal, i.e. bytes read at a wrong offset or
width, not a zeroed value.

The #748 discriminator that motivated this PR used **pyarrow-written** files, where the polarity is
reversed.

**Read together: on Windows each reader misreads the statistics the other's writer produced.** That
is a writer/reader interop bug in the Parquet geospatial-statistics encoding, not "gpio's reader was
wrong" — and switching local reads to pyarrow does not fix Windows, it changes *which* files are
misread. The other three failures are that consequence, with gpio's own DuckDB-written 2.0 output
now judged invalid:

- `test_convert_v20_output_valid`
- `test_a1_auto_version_is_self_consistent[cli-native_geo_only]` and `[api_table_write-native_geo_only]`
- `test_native_stats_containment_ignores_empty`

all reporting *"2 of 2 geometries fall outside geospatial statistics"* against stats read as
`[0, 0, 0, 0]`. Those files are fine.

So the `win32` exclusions this PR removes are load-bearing on **both** readers, not just DuckDB.

### Plan

Narrowing this PR to the part that is platform-independent and demonstrably correct:

- **keep** the whole-file row-group union fix — `native_geo_stats_contains_data_*` compared a sample
  spanning every row group against the *first* row group's bounds, which fails correct
  multi-row-group files on every platform;
- **keep** the SQL-escaping regression coverage and the reader-honesty fixes (a read failure no
  longer reports as "column not found"; bbox and `geometry_types` come from the same row group);
- **drop** the pyarrow-for-local-files reader switch and the removal of the `win32` exclusions.

The interop bug gets its own investigation: testing both readers against files from **both** writers,
then reporting upstream to whichever side is at fault. A discriminator that varies only the writer
tests half the matrix, which is how the first reading came out one-sided.

PR #781, opened 2026-08-30
fix(duckdb): opt out of a5 extension telemetry that segfaults at exit (#779)

### LINKED ISSUE #779: Intermittent SIGSEGV (exit 139) in a5 docs examples during long test sessions
Running the `docs/guide` example suite end to end on `fix/737-unpin-duckdb` (PR #778,
DuckDB 1.5.5) intermittently kills a `gpio ... a5` example with **exit 139 (SIGSEGV)**.
It is session-length dependent: it never reproduces in isolation.

## Evidence

Five full `pytest docs/guide -m "not network"` runs on the branch, macOS, duckdb 1.5.5,
pyarrow 25.0.1:

```
run 1: FAILED docs/guide/add.md::guide/add.md:L294[bash]
run 2: FAILED docs/guide/partition.md:L351[bash]
       FAILED docs/guide/process-aggregate.md:L301[bash]
run 3: clean
run 4 (4-page subset): bash example failed (exit 139) at docs/guide/add.md:270
run 5 (4-page subset): bash example failed (exit 139) at docs/guide/process-aggregate.md:301
```

Every failure is `exit 139` and every one is an **a5** example — `gpio add a5`,
`gpio partition a5 --auto`, `gpio process aggregate a5`. The specific example varies
run to run, which points at accumulated session state rather than any one command.

Does **not** reproduce in isolation:

```
8 standalone CLI runs of `gpio add a5 in.parquet out.parquet --resolution 12`:  rc=0 x8
6 runs of the single failing pytest item (process-aggregate.md:L301):           0 failures
```

## What it is not

- **Not a stale/mismatched extension.** DuckDB stores community extensions per version
  (`~/.duckdb/extensions/v1.5.5/...`), and `a5` is published for both:
  `v1.5.5/osx_arm64/a5.duckdb_extension.gz` -> HTTP 200, `v1.5.1` -> HTTP 200.
- **Not a download failure.** Those surface as `ExtensionUnavailableError` and are
  converted to skips; this is a hard SIGSEGV with no Python traceback.

## Not established

Whether the DuckDB 1.5.1 -> 1.5.5 bump in #778 introduced it. One full run of the same
pages on `origin/main` (duckdb 1.5.1) was clean (`48 passed, 46 skipped`), but a single
clean run proves little against a flake this intermittent — the branch also had a clean
run. Someone should get a failure rate on both versions before blaming the bump.

## Why it matters more than a normal flake

#778 makes `gpio add a5` the **recommended substitute for S2** while the `geography`
extension is unpublished — the changelog, the runtime error message, and eight docs
pages now point users at a5. If a5 has an intermittent hard crash on the DuckDB this
release pins, that recommendation needs re-examining.

Note the `slow-tests` job is `skipping` on PRs, so this does not block a merge; it would
surface post-merge on main.

## Suggested next steps

1. Get a failure rate on 1.5.1 vs 1.5.5 (e.g. 10 full-guide runs each) to attribute it.
2. Capture a core dump / `lldb` backtrace to see whether the fault is in the a5
   extension, `spatial`, or DuckDB core.
3. Check for interaction with extension loading across many sequential DuckDB
   connections within one process — the harness creates a lot of them.

Found while reviewing #778.

cholmes:
Fixes #779.

## The crash

`gpio add a5` segfaults (exit 139) on roughly **1 invocation in 13**. Not a long-session
artifact — a single ordinary CLI call:

| build | trials | SIGSEGV | rate |
|---|---:|---:|---:|
| duckdb 1.5.5 (main) | 150 | 11 | 7.3% |
| duckdb 1.5.5 (interleaved) | 75 | 6 | 8.0% |
| duckdb 1.5.1 (interleaved, same loop) | 75 | 0 | 0% |
| **duckdb 1.5.5 + this fix** | **150** | **0** | **0%** |

The 75/75 runs alternated 1.5.5 / 1.5.1 / 1.5.5 / 1.5.1 to control for network conditions.

## Cause: the `a5` extension phones home on `LOAD`, from a detached thread

`QueryFarmSendTelemetry` POSTs to `https://duckdb-in.query-farm.services/` through httpfs's
bundled curl + OpenSSL. Faulting backtrace (`EXC_BAD_ACCESS`, `KERN_INVALID_ADDRESS at 0x0`):

```
 0 libsystem_pthread.dylib  pthread_rwlock_rdlock + 12580
 1 httpfs.duckdb_extension  CRYPTO_THREAD_read_lock + 8448936
 …
15 httpfs.duckdb_extension  duckdb::HTTPFSCurlClient::Post(duckdb::PostRequestInfo&) + 478260
19 a5.duckdb_extension      duckdb::HTTPUtil::Request(duckdb::BaseRequest&) + 2965708
20 a5.duckdb_extension      …QueryFarmSendTelemetry(duckdb::ExtensionLoader&, …)::$_0>>(void*)
21 libsystem_pthread.dylib  _pthread_start + 27660
```

The regression is in the **per-version `a5` build**, which #778's DuckDB bump selects — not
in DuckDB core. Symbol tables show the threading model changed:

- **v1.5.1 a5**: `__thread_proxy<…__async_assoc_state<void, __async_func<QueryFarmSendTelemetry…>>>`
  — a `std::async` whose future destructor **blocks**, so telemetry is joined before `LOAD` returns.
- **v1.5.5 a5**: `__thread_proxy<…QueryFarmSendTelemetry…::$_0>` — a plain **detached** thread that
  outlives `LOAD` and outlives the command.

So the process finishes its work and exits, teardown frees OpenSSL's global state, and the
still-running telemetry thread — mid-TLS-handshake — dereferences a null rwlock.

**The output file is already written correctly** in every crash (verified 6/6). The only
symptom is the exit code, which is exactly what breaks scripts and CI.

Controls, all inside the same measurement window:

```
40x LOAD a5 per process, normal exit                        20/20 segv
same, with QUERY_FARM_TELEMETRY_OPT_OUT=1                     0/20
40x LOAD spatial instead (same instance count)                0/20
lsof: 11 established TLS connections during a5 loads;  0 with opt-out
```

## The fix

`load_community_extension()` sets `QUERY_FARM_TELEMETRY_OPT_OUT` before `LOAD`. The extension
opts out on the variable's **presence**, whatever the value, so `setdefault` leaves a
user-chosen setting untouched.

Second commit: `core/partition/auto_resolution.py` ran raw `INSTALL a5 FROM community` / `LOAD a5`,
bypassing the helper — which is why `gpio partition a5 --auto` was among the failing examples. It
now routes through `load_community_extension`, which also gains it the `ExtensionUnavailableError`
translation it was missing.

Only `a5` carries telemetry; `h3` and `sqlite_scanner` have zero `QueryFarmSendTelemetry` symbols.

## Why this mattered for the release

#778 makes `gpio add a5` the **recommended substitute for S2** while the `geography` extension is
unpublished — the changelog, the runtime error message and eight docs pages point users at it. As
shipped it exited non-zero on ~8% of invocations.

## Not established

- **No bisect of 1.5.2 / 1.5.3 / 1.5.4** — only the endpoints were compared. Their a5 builds do
  contain the telemetry symbols; their threading model was not checked.
- **macOS arm64 only.** Not tested on Linux or in CI.
- The amplified pure-duckdb reproducer (`N=40` loads) is **not reproducible on demand** — identical
  config measured 20/20 at one hour and 0/25 an hour later, because the race depends on telemetry
  endpoint latency versus process-exit timing. The real-path `gpio add a5` measurement held steady
  at 7-8% across both windows, and the before/after runs were adjacent in time.

## Upstream

This belongs to Query Farm's `duckdb-a5`, not DuckDB. Minimal reproducer:

```python
import duckdb
cons = [duckdb.connect() for _ in range(40)]
for c in cons:
    c.execute("INSTALL a5 FROM community;")
    c.execute("LOAD a5;")
# process exits here; detached telemetry threads are still in the TLS handshake
```

The ask: join the telemetry thread (as the 1.5.1 build's `std::async` did), or register an
atexit / extension-unload hook that waits for it. A detached thread doing TLS must not outlive
the process.

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabled telemetry from community extensions during installation and loading by default.
  * Preserved any existing user-configured telemetry preference.
  * Standardized extension loading across grid processing and resolution workflows.
  * Improved handling when required community extensions are unavailable.

* **Tests**
  * Added coverage for telemetry opt-out behavior and consistent extension loading.


cholmes:
Reviewed and pushed two changes.

**The fix missed three call sites.** `load_community_extension()` is where the opt-out lives, but `process aggregate` (both entry points in `grid_common.py`) and `process overview` (`detect.py`) ran raw `INSTALL a5 FROM community` / `LOAD a5`. Both load the a5 extension, so `gpio add a5` was fixed while `gpio process aggregate a5` and `gpio process overview` still exited 139 at the same rate — and those two are exactly what this release headlines. They now go through the helper, which also gains them the `ExtensionUnavailableError` translation they were missing.

A guard test fails on any new `FROM community` outside the helper. This is the second time a call site skipped it (`auto_resolution.py` was the first, fixed in this PR's second commit), so it seemed worth pinning. `duckdb_utils` and `exceptions` carry the string in docstrings; `extract_bigquery` needs `FORCE INSTALL`, which the helper does not express, and bigquery is not a Query Farm extension. Sabotage-checked: reintroducing a raw `INSTALL` fails the test.

`test_ensure_grid_extension_statements` pinned the raw SQL down to the punctuation, so the helper's trailing semicolons broke it. It now asserts what matters — a5 is installed and loaded, and the opt-out is in the environment before `LOAD`.

**Dropped the hand-written changelog entry.** Changelog sections are generated from PR titles now (#780), so this PR's title is its entry; the hand-written one would have conflicted with the regenerated v1.4.0 section.

One thing I did not change: `_opt_out_of_extension_telemetry()` mutates `os.environ` for the life of the process, which is the sort of global-state leak #660 went after. Scoping it to the `LOAD` and restoring afterwards would introduce a race between concurrent loads under `--workers`, and the variable only disables vendor telemetry, so the permanent `setdefault` looks like the right trade. Worth a second opinion if you disagree.

PR #821, opened 2026-09-03
fix(arcgis): correct polygon ring winding order

### LINKED ISSUE #809: ArcGIS mask geometry loaded incorrectly due to wrong ring winding order in f=geojson output
### What happened?

Running the following command on branch `release/v1.0.0b0`:

```
portolan extract arcgis https://services.arcgis.com/OnnVX2wGkBfflKqu/ArcGIS/rest/services/Mask_HHSK/FeatureServer mask --license CC-BY-4.0
```

The resulting mask polygon geometry (holes/inner rings) is not rendered/loaded correctly. See screenshots: the actual output geometry differs from the expected geometry.

Actual: 

Image

Expected:

Image

### Why it matters

Polygons with holes (e.g. mask layers used to clip/exclude areas) are extracted with incorrect ring winding, causing GIS clients and downstream tools that strictly follow the GeoJSON spec to misinterpret which rings are outer boundaries and which are holes. This can result in inverted or missing holes in the extracted mask geometry, making the output unusable for masking purposes.

### Evidence

- ArcGIS FeatureServer `f=geojson` output uses: outer ring = clockwise, inner ring = counter-clockwise.
- GeoJSON RFC 7946 (section 3.1.6) requires the opposite winding order: outer ring = counter-clockwise, inner ring = clockwise.
- Because ArcGIS's `f=geojson` response does not conform to RFC 7946 winding order rules, consumers that rely on winding order (rather than just ring nesting) to distinguish exterior/interior rings will misinterpret the geometry.

### Implementation details

When extracting an ArcGIS layer via `f=geojson`, the ring winding order needs to be corrected (reversed) to conform to RFC 7946 before writing the output. Alternatively, fetch the layer using `f=json` (Esri JSON format) and convert to GeoJSON explicitly, applying correct winding order during conversion.

### Done

- [ ] ArcGIS extraction (`portolan extract arcgis`) produces polygon geometries with correct RFC 7946 winding order (outer ring = CCW, inner ring = CW)
- [ ] Regression test covering a layer with holes (e.g. a mask layer similar to `Mask_HHSK`) verifying correct ring orientation in the output
- [ ] Verified against the `Mask_HHSK` FeatureServer example from this issue

bertt:
## Summary

ArcGIS's `f=geojson` output winds exterior (shell) rings **clockwise** and interior (hole) rings **counterclockwise** — the reverse of what [RFC 7946 §3.1.6](https://www.rfc-editor.org/rfc/rfc7946#section-3.1.6) and the GeoParquet spec's default `orientation` (counterclockwise exterior, clockwise interior) require. GDAL's GeoJSON/ESRIJSON readers pass ring order through unchanged, so consumers that infer shell/hole from winding direction rather than ring nesting misread these polygons — e.g. a masked area rendering as a hole and vice versa.

Reported in portolan-sdi/portolan-cli#809.

## Fix

`geoparquet_io/core/arcgis.py`'s `_json_doc_to_table` (shared by the GeoJSON and EsriJSON page converters) now reorients every polygon ring after `ST_AsWKB`: the first ring of each polygon is forced counterclockwise, every following ring (a hole) clockwise, based on each ring's own shoelace signed area.

DuckDB's spatial extension has no force-orientation function — `ST_Normalize` only reorders points into a canonical *comparison* form, it doesn't fix winding direction, so it can't be used here. The fix instead reorients rings directly on the WKB bytes with the standard library only (`struct`), no shapely/geopandas dependency, following the same per-ring signed-area approach as opengeos/GeoLibre#2069 (`orientRing`/`signedRingArea`), adapted from TypeScript/GeoJSON coordinates to Python/WKB bytes.

Handles Polygon and MultiPolygon, including Z/M variants; other geometry types and anything that doesn't parse as plain (non-EWKB) WKB pass through unchanged.

## Testing

- New unit test: `tests/test_arcgis.py::TestStreamingConversion::test_geojson_page_to_table_normalizes_arcgis_ring_winding` — feeds a mask polygon wound the ArcGIS way and asserts the resulting WKB is RFC 7946 / GeoParquet conformant.
- `uv run pytest tests/test_arcgis.py` — 121 passed.
- Manual end-to-end check against a real, previously-affected service:

``` 
uv run gpio extract arcgis https://services.arcgis.com/OnnVX2wGkBfflKqu/ArcGIS/rest/services/Mask_HHSK/FeatureServer/0 mask_hhsk.parquet
```

Output verified with  

```
gpio inspect meta mask_hhsk.parquet
```
 :  Geometry Types: MultiPolygon ,  Orientation: Not present - counterclockwise (default value) , correct bbox.

Closes portolan-sdi/portolan-cli#809.

## Summary by CodeRabbit

- **Bug Fixes**
  - Polygon geometries loaded from ArcGIS GeoJSON now use standardized ring orientation: exterior boundaries are counterclockwise and interior holes are clockwise.
  - MultiPolygon geometries now apply the same orientation correction consistently to each polygon.
  - Non-polygon, empty, undersized, or malformed geometries continue to pass through safely without disrupting page conversion.
  - Geometry conversion now better aligns ArcGIS-loaded data with GeoJSON and GeoParquet orientation standards.

### DIFF
--- a/geoparquet_io/core/arcgis.py
+++ b/geoparquet_io/core/arcgis.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 
 import json
 import os
+import struct
 import tempfile
 import uuid
 from collections.abc import Generator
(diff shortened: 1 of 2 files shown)


bertt (commented):
looks good

<sub>Commented in [CodeRabbit Change Stack](https://app.coderabbit.ai/change-stack/geoparquet/geoparquet-io/pull/821)</sub>

cholmes:
> [!NOTE]
> **This review was AI-generated** (Claude, run by @cholmes as part of a multi-agent review pass). Every finding below was reproduced against the branch before being reported — nothing here is speculation — but please read it with that context.

Thanks for this fix @bertt — the diagnosis is right, and the byte-level implementation checks out well: we probed donuts, multipolygons with mixed windings, Z/M variants, big-endian WKB, zero-area/duplicate-vertex/minimum rings, and garbage passthrough, and everything comes out geometry-equal with correct ring roles. The sabotage test also confirms GDAL genuinely does not normalize this, so the tests are load-bearing. Two things should be addressed before merge, though:

### 1. `ST_MakeValid` runs *after* the normalization and re-winds rings clockwise (important)

`_json_doc_to_table` normalizes per page (`arcgis.py:1116`), but `arcgis_to_table` then calls `repair_arrow_table_geometry` on the assembled table (`arcgis.py:1509`), and `--repair-geometry` defaults **on**. Reproduced end-to-end with a self-intersecting ArcGIS-wound polygon:

```
after normalization, before repair: ring areas [68.0]                      # CCW, correct
after ST_MakeValid repair:          ring areas [-3.43, -35.71, -35.71]     # all CW again
```

Invalid geometry is common in exactly the mask/boundary layers that motivated portolan-sdi/portolan-cli#809, so as it stands the fix doesn't hold for a large share of real inputs — and the PR's two fixtures are both valid, so the tests can't see it. Suggested fix: move the normalization to a single whole-table pass **after** the repair step, and add a regression test with an invalid ArcGIS-wound polygon asserting CCW after repair.

### 2. The normalization is never declared in the geo metadata (important)

The output's `geo` block (`arcgis.py:1488-1500`) sets `encoding`/`crs`/`geometry_types` but not `"orientation": "counterclockwise"` — and per the GeoParquet spec, absence is not an assertion, so consumers can't actually rely on the work this PR does. Once finding 1 is fixed, add the orientation key to that block. (Order matters: declaring it while finding 1 stands would produce files that are actively wrong, and gpio's `_check_orientation_matches_data` is currently a stub that wouldn't catch it.)

### Minor notes (non-blocking)

- The per-page Python round-trip runs unconditionally — roughly 95s of overhead on a 1M-feature polygon layer, and it's paid for point/line layers too. `layer_info.geometry_type` is already in scope at both call sites, so gating on `esriGeometryPolygon` is cheap (and the move-after-repair fix naturally makes it one table-level pass).
- Z/M, big-endian, and the EsriJSON page path (what `--output-crs` uses) all work by direct probe but have no tests; the EsriJSON case is the one most worth adding.
- The PR body cites `gpio inspect meta`'s "counterclockwise (default value)" line as spec evidence — that's gpio's own display text, not the spec, which makes no assertion when the key is absent.

Everything else is clean: CI fully green, merges cleanly with current main, changed lines covered by fast-suite tests, title is a valid changelog entry. Happy to push the two fixes to your branch if you prefer (maintainer edits are enabled) — otherwise the shape above should be all that's needed.

bertt:
Pushed a fix for both blocking review items:

1. Winding now re-normalized after repair.  ST_MakeValid  (default  repair_geometry=True ) was re-winding rings clockwise after the per-page normalization. Added  _normalize_table_polygon_winding() , run on the whole table right after repair in  arcgis_to_table  (polygon layers only). Verified with a self-intersecting polygon: post-repair ring areas went from  [-25.0, -25.0]  (CW) to  [25.0, 25.0]  (CCW).
2.  orientation: counterclockwise  added to  geo  metadata for polygon layers, now that the guarantee actually holds after repair.

Added a unit test for the new helper and an end-to-end regression test (invalid bowtie polygon through the full pipeline). 117/117 arcgis tests pass, ruff clean.

bertt:
closing this PR it is already working correctly in version v1.4

Working: 

```
uvx --from git+https://github.com/geoparquet/geoparquet-io@v1.4.0 gpio extract arcgis "https://services.arcgis.com/OnnVX2wGkBfflKqu/ArcGIS/rest/services/Mask_HHSK/FeatureServer/0" mask_hhsk.parquet
```

Not working:

```
uvx --from git+https://github.com/geoparquet/geoparquet-io@v1.3.0 gpio extract arcgis "https://services.arcgis.com/OnnVX2wGkBfflKqu/ArcGIS/rest/services/Mask_HHSK/FeatureServer/0" mask_hhsk.parquet
```

cholmes:
Ah, gotcha. Thanks for the testing and the PR! And for figuring out that it's already working. I'd guess the fix came as part of a bunch of PR's that @cayetanobv did from sucking down a number of arcgis servers.

PR #835, opened 2026-09-04
fix(wfs): request the format the server advertises and parse GML responses

### LINKED ISSUE #818: _build_wfs_url ignores detected output format - hardcoded to GeoJSON
The `_detect_best_output_format()` function correctly checks the server's advertised output formats via GetCapabilities, but the detected format is never passed to `_build_wfs_url()`. The URL builder is hardcoded to `outputFormat=application/json`.

Additionally, `_fetch_wfs_page()` only handles JSON responses and rejects XML/GML content types. This means WFS servers that don't advertise GeoJSON support (common with WFS 1.0.0/1.1.0) will fail even though the code correctly detects that only GML is available.

cholmes:
## What changed

`gpio extract wfs` asked every service for `outputFormat=application/json` regardless of what it said it could produce, and rejected any response whose content type was not JSON. A GML-only service — NextGIS, and most WFS 1.0.0/1.1.0 deployments — was therefore unusable: gpio demanded GeoJSON, got GML at HTTP 200, and reported it as an error page.

Three separate defects combined to cause that, and all three are fixed.

**The capabilities lookup was case-sensitive.** `get_layer_info` read `op.parameters["outputFormat"]`, but OWSLib keys that dict by the literal `name` attribute in the capabilities XML. The WFS 2.0 spec spells it `outputFormat`; NextGIS and deegree publish `OutputFormat`. The lookup found nothing, so gpio believed those servers advertised no formats at all. It is now case-insensitive.

**The negotiated format was computed and thrown away.** `_detect_best_output_format` picked a format and `_build_wfs_url` then hardcoded `application/json` anyway. The negotiated value is now threaded through the request. Matching also tolerates the spelling variation servers really publish (`application/gml+xml;version=3.2` vs `application/gml+xml; version=3.2`, any capitalization) and falls back to a prefix match so an unlisted version suffix still resolves to its family.

The two ways of finding no match are answered differently, because they say different things:

- The server advertised formats and none is readable → omit `outputFormat` and take the server's own default, rather than naming a format it just said it does not support. The old code invented `"GML3"` here; picking the first advertised format instead would be no better, since on GeoServer that can be `SHAPE-ZIP`.
- The server advertised nothing at all → keep asking for `application/json`, which is exactly what gpio sent before. There is no advertisement to honor, so this preserves the fast path for every service whose capabilities gpio cannot parse.

**Parsing was chosen from the request, not the response.** `_fetch_wfs_page` now dispatches on the response `Content-Type`, so a server that ignores the requested format and answers with GML is read instead of rejected. GML is parsed through GDAL (already bundled with DuckDB's spatial extension — no new dependency) into the same shape the GeoJSON path returns: geometry as WKB in a `geometry` column, every server attribute kept, the server-declared CRS carried up through the existing metadata channel, and `gml:id` used as the dedup key for tiled fetches. Everything downstream — Hilbert sorting, bbox, geometry repair, type inference, tile deduplication — works unchanged.

An `ows:ExceptionReport` arrives under the same content types as a `FeatureCollection`, so the GML parser is what tells them apart: a body GDAL cannot open, or one with no geometry column, still produces the same "server returned an error page" message with a body preview that the JSON path has always given.

The response body is now staged in a temporary *directory* rather than a `NamedTemporaryFile`, because GDAL writes a `.gfs` schema sidecar next to schema-less GML and only a directory cleanup is guaranteed to take it along.

## Why

Fixes #818. The investigation found the bug is entirely gpio-side — no upstream fix in OWSLib or GDAL is needed, and DuckDB's bundled GDAL already parses WFS GML natively.

## Verification

End to end against the service from the issue, which advertises only `text/xml; subtype=gml/2.1.2`:

```
$ gpio extract wfs https://demo.nextgis.com/api/resource/9748/wfs \
      Vector_Layer__UK_Cities uk_cities.parquet --verbose
Available formats: ['text/xml; subtype=gml/2.1.2']
Using output format: text/xml; subtype=gml/2.1.2
Server reports 57 features
Parsed 57 GML rows in 0.4s (total: 0.6s)
Server declared CRS EPSG:4326 in its response
Wrote 57 features to uk_cities.parquet
```

```
$ gpio inspect head uk_cities.parquet
Rows: 57
GeoParquet Version: 1.1.0
CRS: OGC:CRS84 (default)
Geometry Types: Point
Bbox: [-7.333284, 50.133722, 1.300013, 60.150035]
Columns (5): gml_id, NAME, POP_MAX, geometry, bbox
```

Before this change that command failed outright.

- `uv run pytest -n auto -m "not slow and not network and not meta"` — 5417 passed, 65 skipped, 2 xfailed
- `uv run pre-commit run --all-files` and the `pre-push` stage (mypy, vulture, xenon, import-linter, deptry) — all pass
- `diff-cover --compare-branch=origin/main` — **97.6%** on changed lines (gate is 90%)

Two unrelated tests (`test_pipe_integration.py::test_column_selection_through_pipe`, `test_geojson_stream.py::test_streaming_handles_broken_pipe`) each failed once under `-n auto` *with* coverage instrumentation and pass serially; neither touches WFS. This is the known xdist cold-run flake.

## New tests

22 added in `tests/test_wfs.py`, covering the case-insensitive capabilities lookup against a real OWSLib parse of NextGIS-shaped XML, format negotiation and normalization, the exact pre-#818 URL as a regression pin, content-type dispatch, GML parsing (axis order, attributes, WKB, CRS, `gml:id`, no leaked temp files), and a network-marked end-to-end extraction from the live NextGIS demo.

## Summary by CodeRabbit

- **New Features**
  - WFS downloads now negotiate supported response formats, including GeoJSON and GML.
  - GML responses preserve coordinate reference systems, axis order, and feature identifiers.
  - Large responses are streamed to temporary storage for local processing.
  - Added guidance on response formats, local parsing, CRS handling, and GML-only public services.

- **Bug Fixes**
  - Improved handling of case-insensitive and parameterized format names, including server-default responses.
  - Prevented unrelated formats from being matched incorrectly.
  - Improved parsing of empty GML responses and features without identifiers.
  - WFS error responses are now rejected more reliably.

### DIFF
--- a/docs/guide/extract.md
+++ b/docs/guide/extract.md
@@ -1205,7 +1205,7 @@ To find service URLs:
 
 ## Extracting from WFS Services
 
-Web Feature Service (WFS) is an OGC standard for serving vector geospatial data over HTTP. Many government agencies and organizations publish data via WFS. gpio uses DuckDB's httpfs extension to stream JSON directly over HTTP, making extraction very fast.
+Web Feature Service (WFS) is an OGC standard for serving vector geospatial data over HTTP. Many government agencies and organizations publish data via WFS. gpio streams each response to disk and parses it locally with DuckDB, making extraction very fast.
 
 ### Basic Usage
 
@@ -1358,6 +1358,24 @@ gpio extract wfs https://geo.example.com/wfs cities output.parquet \
     --wfs-version 1.1.0
 ```
 
+### Response Formats
+
+gpio reads the formats a service advertises in its `GetCapabilities` response
+and asks for the best one it can parse: GeoJSON first, because it is the
+fastest to read, then GML. Nothing needs to be configured — a GeoJSON service
+and a GML-only service are extracted with the same command.
+
+Older WFS 1.0.0/1.1.0 deployments frequently offer GML and nothing else, and
+some servers answer with GML even when a different format was requested. gpio
+decides how to parse a response from the `Content-Type` the server actually
+sent, so either case works. GML is read through GDAL, which resolves the
+geometry, the attributes and the CRS the server used.
+
+If a service advertises formats but none is one gpio can read, gpio omits
+`outputFormat` from the request entirely and takes the server's own default,
+rather than demanding a format the service has just said it does not offer. If
+a service advertises nothing at all, gpio asks for GeoJSON as it always has.
+
 ### Axis Order
 
 WFS 1.1.0+ with URN-format CRS (e.g., `urn:ogc:def:crs:EPSG::4326`) uses lat,lon axis order per OGC spec. gpio auto-detects this, but you can override:
@@ -1386,7 +1404,8 @@ gpio extract wfs https://geo.example.com/wfs cities output.parquet \
 ```
 
 When a server reports the CRS it actually used (GeoServer and most WFS servers
-echo it in the GeoJSON `crs` member), gpio trusts that declaration verbatim —
+echo it in the GeoJSON `crs` member, and a GML response carries an `srsName` on
+its geometries), gpio trusts that declaration verbatim —
 it never second-guesses a server-honored CRS by inspecting coordinates. This
 matters for projected systems: a bounding box alone cannot distinguish, say,
 EPSG:22174 (Argentine Gauss-Krüger) from EPSG:3857 (Web Mercator), so guessing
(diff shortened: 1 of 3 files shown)


cholmes:
Both CodeRabbit findings were valid and are fixed in f0022af.

**Prefix fallback was too loose.** `_first_matching_format`'s second pass used a bare `startswith`, so `application/json-seq` (RFC 7464 JSON text sequences — a stream of records, not one FeatureCollection) resolved to `application/json`, and `gml3-alternate` resolved to `gml3`. Both would have been requested and then failed to parse, or worse, parsed to a single record. The fallback now accepts a longer name only when the extra part is a media-type parameter (begins with `;`). The case it exists for — an unlisted version suffix such as `;version=3.2.1` — still matches, and there are now tests pinning both directions.

**Accept contradicted the request.** `_accept_header_for(None)` returned `application/json`. `None` is precisely the case where gpio has deliberately left `outputFormat` off the request so the server can choose, so naming JSON in `Accept` contradicted that and risked a 406 from a server that cannot produce it. It now sends `*/*`.

Re-verified end to end against the GML-only service from #818 after both changes: 57 features, valid GeoParquet 1.1.0.

*This comment was written by Claude (AI).*

Issue #678, opened 2026-08-21
wfs: _get_feature_count swallows all errors, silently disabling auto-tiling


cholmes:
## Summary

`_get_feature_count` in `geoparquet_io/core/wfs.py` (~lines 1207-1225) wraps its entire body in a bare `except Exception: pass` and falls through to `return None`. Because `wfs_to_table` treats `None` as "the server does not report a count", any transient server hiccup, malformed response, or bug inside the probe silently degrades auto-tiling to no-tiling. No warning is logged.

```python
try:
    content = _make_request(clean_url, params=params)
    # ... regex for numberOfFeatures / numberMatched ...
except Exception:
    pass

return None
```

## Why this matters

`wfs_to_table` calls `_get_feature_count` twice as its auto-tiling probe, then uses the result to decide whether to tile around a server's `maxFeatures` / `startIndex` cap. When the probe silently returns `None`, tiling never engages — which is the same silent-truncation failure mode the `auto_tile` CHANGELOG entry already warns about:

> With it off, a server that caps responses (`maxFeatures` / `startIndex` limits) returned a **silently truncated** table and reported success.

A user gets a short table and a success message, with nothing in the logs to indicate the count probe failed.

## How this was found

While fixing six tests in `tests/test_wfs.py` that each spent ~6.15s in real retry backoff (#665, item 2). Those tests mocked the fetch path but not `_get_feature_count`, so they issued real HTTP GETs to a fake host, failed DNS resolution, and retried with a 1s+2s backoff — twice per test. The bare `except` swallowed every failure, so the tests **passed while unknowingly making live network calls**. The swallow is what kept a plainly broken path invisible for so long; the same property hides real server failures in production.

## Suggested fix

1. Narrow the catch to the exceptions actually expected from the probe (for example `RemoteAccessError` / `WFSError` and the parse failure), and let genuinely unexpected errors surface.
2. Emit `warn()` on the swallowed path so a failed count probe is visible rather than silent, and the caller can distinguish "server reports no count" from "the count probe failed".
3. Consider whether `wfs_to_table` should treat a *failed* probe differently from a *successful* probe that returned no count.

## Follow-up

[...]

PR #836, opened 2026-09-04
fix(wfs): keep existing query params (e.g. apikey) when building WFS GetFeature URLs

### LINKED ISSUE #828: WFS extraction fails with HTTP 403 when the service URL already has a query parameter (e.g. an apikey)
**Problem:**
When `service_url` already contains a query parameter (e.g. an API key), every `GetFeature` request built by `geoparquet-io` fails with **HTTP 403**, though the identical request built by hand returns 200. This breaks all WFS servers that require a key/token in the URL, and affects single fetches and both pagination modes, since they all share one URL builder.

**Cause:**

geoparquet_io/core/wfs.py`, `_build_wfs_url` (line ~1892-1939 in v1.4.0):

```
clean_url = _clean_service_url(service_url)   # keeps "?apikey=mykey"
return f"{clean_url}?{urlencode(params)}"     # appends a 2nd "?" instead of "&"
```

A second  ?  has no delimiter meaning in a query string, so everything after the first  ?  becomes one query string. Everything from  mykey?service=...  onward becomes the literal value of  apikey , so the server gets a corrupted key and rejects the request.

Resulting broken URL:

https://example.com/geo/wfs?apikey=mykey?service=WFS&version=2.0.0&request=GetFeature&typeNames=layer&outputFormat=application/json

Reproduction (no network needed):

from geoparquet_io.core.wfs import _build_wfs_url
print(_build_wfs_url("https://example.com/geo/wfs?apikey=mykey", "layer", "2.0.0"))

Verification:

$ curl ".../wfs?apikey=mykey?service=WFS&version=2.0.0&request=GetFeature&typeNames=layer&outputFormat=application/json"
→ 403
$ curl ".../wfs?apikey=mykey&service=WFS&version=2.0.0&request=GetFeature&typeNames=layer&outputFormat=application/json"
→ 200

Fix: merge new params into the existing query of  service_url  (as  _clean_service_url  already does with  parse_qs / urlencode ) and join with  & , instead of appending a second  ? .

Environment:  geoparquet-io==1.4.0 .

bertt:
Fixes #828

## Problem

When `service_url` already contains a query parameter (e.g. an API key),
`_build_wfs_url` built the final request URL with:

```python
clean_url = _clean_service_url(service_url)
return f"{clean_url}?{urlencode(params)}"
```

`_clean_service_url` only strips WFS-specific keys (`service`, `request`,
`version`, `typename(s)`); it does not remove an existing query string. So a
`service_url` like `https://example.com/geo/wfs?apikey=mykey` produced:

```
https://example.com/geo/wfs?apikey=mykey?service=WFS&version=2.0.0&...
```

A second `?` has no delimiter meaning in a query string, so everything after
the first `?` is parsed as one query -- the `apikey` value became
`mykey?service=WFS&version=2.0.0&...`, and servers that validate the key
returned HTTP 403. This affected the single-fetch path and both pagination
modes, since they all build URLs through `_build_wfs_url`.

## Fix

Added `_merge_query_params(url, params)`, which parses the existing query
string, merges in the new WFS params (new params win on key collision), and
rebuilds a single, correctly `&`-joined query string. `_build_wfs_url` now
uses this helper instead of raw string concatenation.

`_get_feature_count` / `_make_request` were not affected -- they pass `params`
separately to httpx's `client.get()`, which already merges correctly with an
existing URL query.

## Tests

Added `TestBuildWfsUrlQueryParamMerging` in `tests/test_wfs.py`:
- Exact reproduction from the issue (`apikey` preserved, single `?`)
- No-existing-query-param case (non-regression)
- Multiple pre-existing query params merged correctly
- WFS control params (`service`/`request`/`version`) in `service_url` are
  still stripped before merging, so `request=GetFeature` isn't overridden by
  a stale `GetCapabilities` value

## Validation

- `uv run pytest tests/test_wfs.py -n auto -m "not slow and not network and not meta"` -- 218 passed, 2 skipped
- `uv run pytest -n auto -m "not slow and not network and not meta"` -- 5327 passed (remaining failures pre-exist on `main`, unrelated to this change)
- `uv run pre-commit run --files geoparquet_io/core/wfs.py tests/test_wfs.py` -- ruff, duckdb-antipatterns, no-click-echo, doc-sync all pass
- No CLI/API surface change (private core helper); no `docs/guide/` or `api/` updates needed

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed WFS request URL construction when the base URL already contains query parameters.
  - Existing parameters, such as API keys, are now preserved while WFS parameters are added or updated.
  - Improved handling of multiple query parameters and separators.

- **Tests**
  - Added regression coverage for query-parameter merging and replacement behavior.

### DIFF
--- a/geoparquet_io/core/wfs.py
+++ b/geoparquet_io/core/wfs.py
@@ -319,6 +319,48 @@ def _clean_service_url(url: str) -> str:
     )
 
 
+def _merge_query_params(url: str, params: dict) -> str:
+    """
+    Merge additional query params into a URL that may already have a query string.
+
+    Building the final URL by naive string concatenation (``f"{url}?{...}"``)
+    corrupts any pre-existing query string (e.g. ``?apikey=mykey``) by adding
+    a second ``?``, which has no delimiter meaning in a query string — the
+    server sees a single, garbled query and everything after the first ``?``
+    (including the real ``apikey`` value) is misparsed (Issue #828).
+
+    New params take precedence over any same-named param already present in
+    ``url`` (WFS control params like ``service``/``request``/``version``
+    should always reflect the request being built, not a stale value carried
+    over from the caller's URL).
+
+    Args:
+        url: Base URL, with or without an existing query string.
+        params: Additional query params to merge in.
+
+    Returns:
+        URL with a single, correctly merged query string.
+    """
+    parsed = urlparse(url)
+    existing_params = parse_qs(parsed.query, keep_blank_values=True)
+    # dict values from parse_qs are lists; flatten before merging with new
+    # scalar values so urlencode doesn't emit e.g. "apikey=['mykey']".
+    merged = {k: v[0] if len(v) == 1 else v for k, v in existing_params.items()}
+    merged.update(params)
+
+    new_query = urlencode(merged, doseq=True)
+    return urlunparse(
+        (
+            parsed.scheme,
+            parsed.netloc,
+            parsed.path,
+            parsed.params,
+            new_query,
+            "",
+        )
+    )
+
+
 def get_wfs_capabilities(service_url: str, version: str = "1.1.0"):
     """
     Get WFS capabilities using OWSLib.
@@ -2367,8 +2409,6 @@ def _build_wfs_url(
     replies in its own default — the only thing that works on a service which
     advertises no format gpio recognizes (issue #818).
     """
-    from urllib.parse import urlencode
-
     clean_url = _clean_service_url(service_url)
 
     params = {
@@ -2404,7 +2444,7 @@ def _build_wfs_url(
     if sort_by and version != "1.0.0":
         params["sortBy"] = sort_by
 
-    return f"{clean_url}?{urlencode(params)}"
+    return _merge_query_params(clean_url, params)
 
 
 def _single_fetch_mode(
--- a/geoparquet_io/core/wfs.py
+++ b/geoparquet_io/core/wfs.py
@@ -305,6 +305,48 @@ def _clean_service_url(url: str) -> str:
     )
 
 
+def _merge_query_params(url: str, params: dict) -> str:
+    """
+    Merge additional query params into a URL that may already have a query string.
+
+    Building the final URL by naive string concatenation (``f"{url}?{...}"``)
+    corrupts any pre-existing query string (e.g. ``?apikey=mykey``) by adding
+    a second ``?``, which has no delimiter meaning in a query string — the
+    server sees a single, garbled query and everything after the first ``?``
+    (including the real ``apikey`` value) is misparsed (Issue #828).
+
+    New params take precedence over any same-named param already present in
+    ``url`` (WFS control params like ``service``/``request``/``version``
+    should always reflect the request being built, not a stale value carried
+    over from the caller's URL).
+
+    Args:
+        url: Base URL, with or without an existing query string.
+        params: Additional query params to merge in.
+
+    Returns:
+        URL with a single, correctly merged query string.
+    """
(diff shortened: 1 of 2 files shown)


cholmes:
Thanks for the fix @bertt — the core change is sound and the tests pin it well. We ran an adversarial review pass over the branch; **the findings below come from an AI review agent** (verified by running repros against the locked deps: httpx 0.28.1, owslib 0.35.0), posted here for transparency.

### S2 — the feature-count probe still drops the apikey (`geoparquet_io/core/wfs.py:1261`)

The PR body's "not affected" note about `_get_feature_count` doesn't hold: httpx 0.28.1 **replaces** a URL's existing query when `params=` is passed to `client.get()`, it doesn't merge. Repro:

```python
httpx.Client().build_request("GET", "https://ex.com/wfs?apikey=mykey",
                             params={"service": "WFS", "resultType": "hits"})
# → https://ex.com/wfs?service=WFS&resultType=hits   (apikey gone)
```

So on the exact #828 scenario (server 403s without the key), the `resultType=hits` probe at wfs.py:1249–1261 gets a 403 → `WFSAuthenticationError` is swallowed by the `except (WFSError, httpx.HTTPError, OSError)` at wfs.py:1262 → count comes back `None` → pagination/auto-tiling are disabled, which on a response-capping server means silently truncated output (the #678 failure mode). Suggested fix: route that probe's URL through `_merge_query_params` too (it's the only `_make_request` caller passing `params`; `_probe_start_index_limit` already goes through the fixed `_build_wfs_url`), plus a test.

(Capabilities/DescribeFeatureType are fine — OWSLib's `capabilities_url()` preserves the existing query; verified.)

### S3 — mixed-case WFS control params escape the strip and now produce conflicting duplicate keys

`_clean_service_url` (wfs.py:290–292) pops only exact-lowercase/UPPERCASE variants — notably not `typeName`, the canonical WFS 1.1 casing users copy from a browser URL. Repro:

```
_build_wfs_url("https://ex.com/wfs?typeName=roads&apikey=k", "buildings", "2.0.0")
# → ...?typeName=roads&...&typeNames=buildings&...
```

WFS KVP keys are case-insensitive per the OGC spec, so which layer the server picks from `typeName=roads&typeNames=buildings` is server-dependent — potentially silently wrong data. On main these inputs produced a garbled double-`?` URL (loud failure); after this PR they're a well-formed request that may quietly do the wrong thing. Cheap fix: make the control-key strip case-insensitive.

### S4 (no action needed, noting for completeness)

- Invalid percent-sequences in existing values get corrupted by the decode/re-encode round trip (`?tok=%FF` → `?tok=%EF%BF%BD`) — only bites tokens that aren't valid UTF-8 when decoded, which real apikeys essentially never are.
- Cosmetic canonicalization: `%20` → `+`, bare `?flag` → `flag=`, fragments dropped. Multi-valued params and already-encoded `%2F`/`%2B` round-trip correctly; userinfo/port survive.
- Test-gap nits: no case for multi-valued existing params or mixed-case control keys.

### Overall

Every GetFeature path (single fetch, sequential and parallel pagination, the startIndex probe) goes through `_merge_query_params` correctly, and the new tests would catch a reintroduction of the bug (offline suite: 220 passed). The one ask before merge is the S2 — the count probe still hits the un-merged URL, so the headline use case can still silently truncate.

🤖 These findings were generated by an AI review agent; a maintainer will follow up.

cholmes:
Thanks @bertt! I'll take this one from here — applying the remaining fixes from the review above (the feature-count probe and the mixed-case control-key strip) and merging. No further action needed on your side.

bertt:
great!

Issue #738, opened 2026-08-27
sort hilbert --add-bbox writes the bbox column but no covering metadata at 2.0 output, so nothing can use it


tristangrupp:
`gpio sort hilbert --add-bbox` writes the bbox column but no `covering` metadata when the output is 2.0, so nothing can use it.

With a WKB-encoded GeoParquet **2.0.0** input, auto version mode keeps the output at 2.0.0 and `--add-bbox` adds the physical `bbox` column — but writes no `covering` entry in the `geo` metadata. Because the geometry stays WKB, the file also has no native `GeospatialStatistics`. The result is a file with **no discoverable spatial statistics at all**, carrying a bbox column that costs bytes and delivers nothing.

The command reports success:

```
Added bbox column for optimized processing
Output includes bbox column and metadata for optimal performance
```

## Repro

Build a WKB 2.0.0 input (3,600 points), then sort it two ways:

```python
import json, pyarrow as pa, pyarrow.parquet as pq, shapely
pts = [shapely.Point(x/10.0, y/10.0) for x in range(60) for y in range(60)]
geo = {"version": "2.0.0", "primary_column": "geometry",
       "columns": {"geometry": {"encoding": "WKB", "geometry_types": ["Point"],
                                "crs": {"id": {"authority": "EPSG", "code": 4326}}}}}
t = pa.table({"id": pa.array(range(len(pts))),
              "geometry": pa.array([shapely.to_wkb(p) for p in pts], pa.binary())})
pq.write_table(t.replace_schema_metadata({b"geo": json.dumps(geo).encode()}), "in.parquet")
```

```bash
gpio sort hilbert in.parquet out_auto.parquet --add-bbox
gpio sort hilbert in.parquet out_11.parquet   --add-bbox --geoparquet-version 1.1
```

Observed (gpio 1.3.0):

```
in        v=2.0.0  schema=id,geometry       geomtype=binary  covering=ABSENT
out_auto  v=2.0.0  schema=id,geometry,bbox  geomtype=binary  covering=ABSENT
out_11    v=1.1.0  schema=id,geometry,bbox  geomtype=binary  covering={"bbox": {...}}
```

Only the explicit `--geoparquet-version 1.1` output is usable for row-group pruning.

## Why it matters

This is silent: the bbox column exists, the success message says metadata was included, and the file is valid. Nothing surfaces the gap until a validator or a reader asks for spatial statistics.

[...]

PR #877, opened 2026-09-05
fix(validate): enforce the covering.bbox path, field-order and field-type rules

jatorre:
## Problem

The `covering.bbox` checks accepted files that violate rules the spec states as MUST (GeoParquet 1.1 "bbox covering encoding" and "Bounding Box Columns", same text in PR opengeospatial/geoparquet#302 for 2.0):

| Spec rule | Before |
| --- | --- |
| "the second item in each element MUST be `xmin`, `ymin`, etc." | not checked: `"xmin": ["bbox", "minx"]` passed |
| "All values MUST use the same column name" | not checked: `xmin` on `bbox`, `ymax` on `other` passed |
| child fields "MUST be ordered" `xmin, ymin, xmax, ymax` / `xmin, ymin, zmin, xmax, ymax, zmax` | only the set of names was compared; wrong order, 5 children, 7 children all passed |
| "If `zmin` is present then `zmax` MUST be present and vice versa" | not checked |
| "The fields MUST be of Parquet type FLOAT or DOUBLE and all columns MUST use the same type" | only the first four children were typed; `zmax INT32` passed |

## Change

Three functions, no new ones:

- `_check_covering_bbox_paths`: after the existing shape check, `path[1]` must equal its key, and all four `path[0]` must be the same column.
- `_check_covering_bbox_structure`: the child field names, in schema order, must equal `["xmin", "ymin", "xmax", "ymax"]` or `["xmin", "ymin", "zmin", "xmax", "ymax", "zmax"]` (module constant `_BBOX_FIELDS`). Replaces the "at least 4 children containing these names" logic.
- `_check_covering_bbox_field_types`: iterate all children instead of the first four.

## Tests

`tests/test_validate_covering_strictness.py` (8 of 11 failed on `main`). Files are real Parquet written with pyarrow, one per violation: bad second element, mixed columns, swapped XY order, XYZ in the wrong order, `zmin` without `zmax`, an extra fifth child, `zmax INT32`; valid XY and XYZ files pass all three checks; `tests/data/austria_bbox_covering.parquet` still passes; the bad-paths file fails through `validate_geoparquet`.

Regression: validate, check, corpus and bbox suites — 497 passed, 8 xfailed. Pre-commit passes.

```
$ gpio check spec cov_bad.parquet         # xmin -> ["bbox","minx"], ymax -> ["other","ymax"]
  ✓ column "geometry" has valid covering object
  ✗ covering bbox path field must equal its key: ['xmin']
$ gpio check spec cov_zmax_int.parquet
  ✗ bbox fields must be FLOAT or DOUBLE (found: {'INT32'})
```

## Adversarial pass (done before opening)

- Coverings written by gpio itself pass unchanged: `gpio convert … --geoparquet-version 1.1` + `gpio add bbox` (28 checks green), and `gpio add bbox` on `fields_gpq2_crs84_zstd.parquet` at 2.0 (36 checks green); `gpio check bbox` unaffected.
- Both `get_schema_info` paths (pyarrow fast path and DuckDB `parquet_schema()`) give the same verdicts.
- A covering that points at a nested list-of-struct group (not a bbox) is now rejected by the structure check on both paths; before, it passed the paths and (as the existing test shows) failed only on types.
- Non-string `path[0]` values and `num_children: None` do not raise.

Found while mapping the OGC GeoParquet 2.0 abstract tests to `gpio` (opengeospatial/geoparquet#304).

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for covering bounding boxes.
  * Added support for correctly validating 3D bounding boxes.
  * Detects mismatched field names, invalid column paths, incorrect field ordering, extra fields, and incompatible data types.
  * Improved validation reporting for invalid bounding box paths.

* **Tests**
  * Added comprehensive coverage for valid and invalid 2D and 3D bounding box configurations.

### DIFF
--- a/geoparquet_io/core/validate.py
+++ b/geoparquet_io/core/validate.py
@@ -1539,6 +1539,24 @@ def _check_covering_bbox_paths(col_meta: dict, col_name: str) -> ValidationCheck
                 category="geoparquet_1_1",
             )
 
+    wrong_field = [k for k in required_keys if bbox_covering[k][1] != k]
+    if wrong_field:
+        return ValidationCheck(
+            name=f"covering_bbox_paths_{col_name}",
+            status=CheckStatus.FAILED,
+            message=f"covering bbox path field must equal its key: {wrong_field}",
+            category="geoparquet_1_1",
+        )
+
+    columns = sorted({str(bbox_covering[k][0]) for k in required_keys})
+    if len(columns) > 1:
+        return ValidationCheck(
+            name=f"covering_bbox_paths_{col_name}",
+            status=CheckStatus.FAILED,
+            message=f"covering bbox paths must name a single column (found: {columns})",
+            category="geoparquet_1_1",
+        )
+
     return ValidationCheck(
         name=f"covering_bbox_paths_{col_name}",
         status=CheckStatus.PASSED,
@@ -1592,10 +1610,17 @@ def _check_covering_bbox_column_exists(
     )
 
 
+_BBOX_FIELDS = {
+    4: ["xmin", "ymin", "xmax", "ymax"],
+    6: ["xmin", "ymin", "zmin", "xmax", "ymax", "zmax"],
+}
+
+
 def _check_covering_bbox_structure(
     col_meta: dict, col_name: str, schema_info: list
 ) -> ValidationCheck:
-    """Check 1.1-4/5: covering bbox column must be a struct with xmin/ymin/xmax/ymax."""
+    """Check 1.1-4/5: covering bbox column is a struct of xmin/ymin/xmax/ymax or
+    xmin/ymin/zmin/xmax/ymax/zmax, in that order."""
     covering = col_meta.get("covering")
 
     if covering is None or "bbox" not in covering:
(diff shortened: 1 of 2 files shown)


cholmes:
Thanks @jatorre — this one holds up very well under adversarial review. **The findings below come from an AI review agent** (all verified by execution in a synced worktree), posted for transparency.

### Spec anchoring: verified, all five rules

Given this repo's history with covering validation (an earlier PR invented rules and cited unmerged spec text), the reviewer fetched `format-specs/geoparquet.md` at tag **v1.1.0** and matched every enforced rule to merged text. All five check out:

- path second element (`"the second item in each element MUST be xmin, ymin…"`),
- single column (`"All values MUST use the same column name."`),
- 4-or-6 children in exact order (`"…MUST be named xmin, ymin, xmax, and ymax and MUST be ordered in this same way."`),
- z pairing + XYZ order (`"If zmin is present then zmax MUST be present…"`),
- FLOAT/DOUBLE with same type across columns.

So the "extra fifth child fails" and "swapped order fails" behaviors are spec-mandated, not inventions. Version scoping was probed at runtime and is right: 1.0 files skip covering checks entirely; 2.0 files carrying a covering get checked per the existing #738 decision.

### Probes that came back clean

- xmin path pointing at the xmax field → caught; nonexistent column → caught; valid XY/XYZ and non-`bbox`-named covering columns → pass (no invented naming rule); extra covering-JSON keys tolerated (correct leniency).
- Both schema paths (pyarrow fast path and DuckDB `parquet_schema()`) give the same verdicts — the tests only exercise the pyarrow path, but the DuckDB path was probed independently and matches.
- 8 of the 11 new tests fail on main, exactly as the PR body claims; mixed float32/float64 z-fields now correctly feed the same-type check.
- Merge with current main is clean; 280 validate-suite tests pass.

### Two pre-existing S4s (not this PR's fault — filed as a follow-up issue, no change requested here)

1. `_check_covering_bbox_structure` / `_check_covering_bbox_field_types` (validate.py:1646/1699) take the *first* flat-schema entry whose name matches, so a struct column with a **child** field named `bbox` appearing before the real root `bbox` column shadows it → cryptic false FAIL (`"…in that order (found: [])"`). Reproduces on main too.
2. `_check_covering_bbox_column_exists` (validate.py:1594) says "exists at schema root" but matches any flat-schema name, so the spec's "MUST be at the root of the schema" is still not truly enforced.

### Verdict

Ready to merge once the pending CI jobs (notably the ubuntu-3.11 coverage/diff-cover job) go green — the new tests are unmarked fast-suite tests covering every changed line, so the gate should clear. Nice work; a maintainer will follow up on the merge.

🤖 These findings were generated by an AI review agent.

cholmes:
Thanks @jatorre! Per the maintainer: we'll take this one from here — applying the remaining fixes from the review above and merging. No further action needed on your side.

PR #900, opened 2026-09-06
fix(partition): support explicit quadkey resolutions in directory mode

### LINKED ISSUE #854: partition quadkey directory mode (--min-size) forwards only --resolution, so every per-file run fails unless --auto is used
Found while adding the Python API for directory sub-partitioning (#811, PR #853). Pre-existing on `main`, confirmed on the CLI before any change; #853 mirrors the behaviour exactly rather than fixing it, so the API has the same limit.

## Repro

```
gpio partition quadkey by_country/ --min-size 1KB --resolution 13 --in-place
```

Every file over the threshold fails with:

```
must specify either --auto or both --resolution and --partition-resolution
```

## Cause

`partition quadkey` (single-file mode) requires **two** resolutions — `--resolution` for the cell column and `--partition-resolution` for the directory split — or `--auto`. Directory mode (`core/sub_partition.py`, the `partition_funcs` registry entry `"quadkey": (partition_by_quadkey, "resolution")`) forwards **one** value under the name `resolution` and never passes `partition_resolution`, so the per-file call always trips the single-file gate. h3/a5/s2 take one resolution and are unaffected.

`--auto` works because the auto path computes both.

## Direction

Either thread `--partition-resolution` through directory mode for quadkey (the registry's single `(func, param_name)` shape needs to grow to carry a second parameter, or quadkey gets a small adapter), or make directory mode derive `partition_resolution` from `resolution` the way single-file `--auto` does and say so in the help. Then remove the "use `auto=True`" workaround note #853 added to `docs/guide/sub-partitioning.md` and the test that pins the failure on both doors.

Refs #811, #853, #733, #790

be-student:
## Description

Fix directory quadkey partitioning with explicit resolutions. Previously, `gpio partition quadkey by_country/ --min-size 1KB --resolution 13 --partition-resolution 6` dropped the partition resolution and failed every selected file. The CLI now forwards both values, and `ops.sub_partition_by_quadkey(..., resolution=13, partition_resolution=6)` supports the same operation.

## Technical Details

The parameter now travels through both entrypoints and their shared directory dispatcher to the existing quadkey partitioner. It retains the single-file rules: both explicit resolutions are required, partition precision cannot exceed column precision, and explicit resolutions cannot be combined with auto mode. Existing `auto=True` behavior and the source-row preservation guard remain in place. New parameters are keyword-only or appended to existing positional signatures.

Real GeoParquet integration tests cover CLI and API output at partition precisions 0, 6, and 13, exact output prefix lengths, row preservation, threshold filtering, and safe in-place replacement. Invalid values and auto conflicts must leave the originals byte-for-byte intact and create no output. Both entrypoints' shared-dispatch contract also covers the new parameter. The guide and Python reference document the supported explicit mode, with executable examples.

## Breaking Changes

## Related Issue(s)
Fixes #854.

## Validation

- Before implementation: all six explicit-resolution CLI/API regressions failed (CLI lost the argument; API rejected the keyword).
- Sub-partition suite: 56 passed, 1 skipped.
- Executable sub-partition guide: 13 passed.
- Full fast suite (`uv run pytest -n 4 -m "not slow and not network and not meta"`): **6,193 passed, 77 skipped, 9 xfailed**, with 110 warnings. Slow, network, and meta lanes were not run locally.
- All commit-stage and pre-push hooks passed, including type, architecture, dependency, complexity, and documentation checks; generated doc check and `git diff --check` passed.
- Changed executable lines: 100% covered (`diff-cover`, required threshold 90%).
- An independent adversarial AI review checked parameter propagation, positional compatibility, invalid/auto combinations, and data-loss paths. Its requests for an exact zoom-zero partition count and invalid-input source-preservation tests were implemented.

This contribution was implemented and validated with Codex. Hosted CI and CodeRabbit feedback will be checked after submission.

## Checklist
- [ ] All tests pass (`uv run pytest`)
- [x] CodeRabbit comments fixed or answered

## Summary by CodeRabbit

* **New Features**
  * Added support for specifying a separate Quadkey partition resolution through the API and command-line workflows.
  * Added validation to ensure partition resolution is within the allowed range.
  * Directory-based Quadkey partitioning now consistently applies the configured partition resolution.

* **Documentation**
  * Updated API and sub-partitioning guides with the new option, usage examples, and automatic resolution behavior.

* **Tests**
  * Added coverage for explicit resolutions, validation errors, filename prefixes, row preservation, and original-file handling.

### DIFF
--- a/docs/api/python-api.md
+++ b/docs/api/python-api.md
@@ -1148,7 +1148,7 @@ for file_path in large_files:
 
 **Returns:** List of file paths sorted by size (largest first)
 
-#### `sub_partition_directory(directory, partition_type, min_size_bytes, resolution=None, level=None, in_place=False, hive=False, overwrite=False, verbose=False, force=False, skip_analysis=True, compression='ZSTD', compression_level=None, auto=False, target_rows=100000, max_partitions=10000)`
+#### `sub_partition_directory(directory, partition_type, min_size_bytes, resolution=None, level=None, in_place=False, hive=False, overwrite=False, verbose=False, force=False, skip_analysis=True, compression='ZSTD', compression_level=None, auto=False, target_rows=100000, max_partitions=10000, partition_resolution=None)`
 
 Sub-partition large files in a directory using spatial indexing.
 
@@ -1193,6 +1193,7 @@ result = sub_partition_directory(
 - `partition_type` (str): Type of partition ("h3", "a5", "s2", "quadkey")
 - `min_size_bytes` (int): Minimum file size to process
 - `resolution` (int | None): Resolution for H3/quadkey (0-15 for H3)
+- `partition_resolution` (int | None): Quadkey partition prefix length (0-23, no greater than `resolution`); required with explicit quadkey `resolution`
 - `level` (int | None): Level for S2 (alias for resolution)
 - `in_place` (bool): Delete originals after successful sub-partition (default: False)
 - `hive` (bool): Use Hive-style partitioning (default: False)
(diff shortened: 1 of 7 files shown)


cholmes:
Thanks @be-student! This went through our adversarial review pipeline (AI review agent, findings verified by execution) and came back clean — no blocking findings. Highlights from the review:

Verified end-to-end: the exact #854 command fails on main and succeeds on this branch, with directory mode now matching single-file semantics precisely (column at `--resolution`, split at `--partition-resolution`, rows preserved 42/42). The regression tests genuinely fail on main — 11 of them — and updating the front-door parity CASES table was exactly right; nothing would have forced that, so thank you for doing it anyway. The invalid-input matrix asserting originals stay byte-for-byte intact is the kind of test we want more of.

Merging as-is. 🤖 Review findings were generated by an AI review agent.

Issue #421, opened 2026-04-29
[Bug]: pmtiles creation fails with exit 1 but no error logs


C-Loftus:
### Bug Description

I am trying to generate pmtiles from parquet data using the pmtiles create command but it fails without any logs.

### Steps to Reproduce

Install the latest gpio via `uv tool install geoparquet-io --force` and run the command in the logs below

My open data parquet file can be found here: https://storage.cloud.google.com/metadata-geoconnex-us/exports/geoconnex_features.parquet if there is some quirk of this specific dataset. It was validated with gpio for what its worth.

### Expected Behavior

pmtiles should have been created successfully; if there were errors there should have been error logging. I think the tippeanoe sub process command just doesn't stream logs out.

### Actual Behavior

```shell
gpio pmtiles create exports_geoconnex_features.parquet geoconnex.pmtiles
2949123 features, 272438626 bytes of geometry and attributes, 242183613 bytes of string pool, 970636416 bytes of vertices, 5863376 bytes of nodes
Choosing a maxzoom of -z9 for features typically 2364 feet (721 meters) apart, and at least 188 feet (58 meters) apart
  99.9%  7/28/45    
Error: /Users/cloftus/.local/share/uv/tools/geoparquet-io/bin/gpio failed with exit code 1
```

### Environment

```markdown
- gpio version: [run `gpio --version`]


 gpio --version
geoparquet-io, version 1.1.1


- Installation method: [pipx, pip, uv tool, from source]

uv tool
```

For what its worth, I have only 40GB of storage left on my machine. I am not sure if there are intermediate files that are created that would cause issues due to size? The geoparquet is <1 GB in size

nlebovits:
@C-Loftus thanks for flagging! Should be fixed--try it out?

C-Loftus:
Thank you @nlebovits it seems to log properly now! I appreciate the fix

PR #907, opened 2026-09-08
fix(geojson): write UTF-8 to stdout from the Python API too

cholmes:
## Description

`gpio.read(...).to_geojson()` with no output path crashed with
`UnicodeEncodeError` whenever the data held a character the console codec could
not spell:

```
UnicodeEncodeError: 'charmap' codec can't encode character 'ū' in position 157
```

GeoJSON is UTF-8 by definition ([RFC 7946 §11](https://www.rfc-editor.org/rfc/rfc7946#section-11)),
but Python gives `sys.stdout` the console's codec — cp1252 on a default Windows
console. The UTF-8 reconfigure lived in the CLI group callback
(`cli/main.py`), so CLI users were covered and Python API callers were not:
`core/geojson_stream.py` wrote straight to `sys.stdout` and died on the first
place name outside Latin-1, *after* some features were already written.

125 of the 766 rows in the canonical sample carry such a name, so this is most
real-world place data, not an exotic edge case.

## Technical Details

Both stdout paths in core — `_stream_to_stdout` and the FeatureCollection
branch of `_stream_feature_collection` — now write through a new
`_utf8_stdout()` context manager. It wraps the stream's binary buffer for the
duration and **detaches, never closes, it afterwards**, so `sys.stdout` is left
exactly as we found it rather than reconfigured process-wide behind the
caller's back. A stream that already encodes UTF-8 (under any spelling of the
name, via `codecs.lookup`), or that has no binary layer to wrap, is yielded
untouched.

The cleanup suppresses `OSError` so a broken pipe — already handled inside
`_stream_to_stdout` for the `| head` / tippecanoe case — does not resurface
past its handler.

The CLI's own reconfigure is left in place: it also covers stderr and every
other command's non-ASCII output.

## Breaking Changes

## Related Issue(s)
- Caught by the Windows `slow-tests` leg on `docs/guide/geojson.md:341`. That
  lane only runs post-merge, which is why `main` has carried this since Aug 30
  while PRs stayed green.

## Checklist

<sub>Tests: three new cases in `TestNonUtf8Stdout` drive a cp1252
`TextIOWrapper` as `sys.stdout` — the GeoJSONSeq path (the one that failed),
the FeatureCollection path, and a check that a UTF-8 stream is left alone. The
first fails on `main` with the exact CI traceback. Verified end to end under
`PYTHONIOENCODING=cp1252`: exit 0, 213,991 bytes of valid UTF-8. Docs lane
`docs/guide/geojson.md` passes (19 passed, 9 skipped). No doc change needed —
the guide text was already correct.</sub>

## Summary by CodeRabbit

* **Bug Fixes**
  * GeoJSON output now supports characters outside the system console’s default encoding.
  * Streaming, FeatureCollection, and pretty-printed output remain reliable across different console encodings.
  * Existing UTF-8 output behavior is preserved without changing the console configuration.
  * Broken-pipe handling now also covers errors encountered while preparing console output, improving reliability when output destinations close unexpectedly.


cholmes:
Good catch on the entry flush — confirmed and fixed in f5ea50e.

Reproduced it first: a cp1252 `sys.stdout` whose `flush()` raises `BrokenPipeError` made `_stream_to_stdout` propagate the error out of the context manager's `__enter__`, one level outside the handler — the exact exit-1 shape of #421. The `with` now sits inside the `try`, so both the entry flush and the exit detach are covered, and there's a regression test (`test_a_pipe_that_broke_before_the_first_write_is_still_handled`) that fails without the change.

Issue #901, opened 2026-09-07
test(docs-harness): extend the network-fence guard to python fences and decide the extension-install class


cholmes:
Follow-ups from the adversarial review of #895 (AI review agent; the guard itself was sabotage-verified to fail loudly and has zero false positives on today's corpus). Two scope gaps, neither with a live offender:

1. **The guard is bash-only.** `test_network_fences_carry_the_network_directive` filters `block.lang == "bash"`, but `EXECUTABLE_LANGUAGES` includes python and 11 python fences are network-marked today. A future unmarked Python-tab example calling `partition_by_admin(dataset="gaul")` triggers the same 724MB download and slips past — the exact recurrence the guard exists to prevent, in the other executable language. Fix: a second, python-shaped pattern set over python fences.
2. **Decision: the community-extension-install class.** `gpio add s2`/`partition s2` (and h3/a5) fences run `INSTALL … FROM community` — a live CDN fetch on any cold machine, including fresh CI runners. #893 made the s2 fences live. Either mark these fences `network`, or record that DuckDB-CDN extension installs are accepted in the default lane (small, first-run-only per machine; CI has been green with h3/a5 doing this all along). The guard's premise ("needs the internet ⇒ marked") currently excludes this class silently.

Also two cosmetic notes from the same review: `_NEEDS_NETWORK["--country-codes"]` matches nothing (no gpio command accepts that flag — the command is `add admin-divisions`) and its comment describes a nonexistent behavior; and prompt-prefixed (`$ gpio …`) lines would evade the `^\s*gpio` anchors (zero exist today).

Refs #895, #894.

Issue #739, opened 2026-08-28
ci: cache ~/.geoparquet-io/cache/admin in the slow job so journey 8 gates in CI


cholmes:
PR #725's journey 8 (add admin-divisions → partition admin → check leaves) deliberately rides the slow lane and skips when the Overture boundary cache is cold. On CI's ephemeral runners the cache is always cold, and a measured cold run (~3m05s, ending in a DuckDB HTTP timeout) exceeds the journey timeout — so in CI the journey chronically skips and the gate only holds on warm developer machines.

Fix: add an `actions/cache` step to the slow-tests job for `~/.geoparquet-io/cache/admin`, keyed on the pinned Overture release, so the second and later slow runs execute journey 8 for real. First run after a release bump repopulates the cache.

Found during the adversarial review of #725.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01KLN6JAz97UT1gGygg7EVTN

cderinbogaz:
Disclosure: I work on [Runnerhut](https://runnerhut.com). A second-run cache hit will show the warm path works, but the first run after an Overture release bump is still an important gate case. I would make cache population an explicit bounded operation and distinguish a failed population from an intentional skip in the required result. That avoids a green cold-cache run quietly omitting journey 8. Runnerhut also uses ephemeral environments, so changing runners alone would not supply this missing persistent-data path.

PR #908, opened 2026-09-08
test(ci): unbreak the slow-test lane, and 3 of the nightly's 43 failures

cholmes:
## Description

> **Correction (after opening):** this PR was opened claiming it unbreaks the
> nightly mutation lane. It does not. A full `mutmut run` with mutmut's `-x`
> overridden shows **43** tests that cannot survive the rewrite; the change here
> clears 3 of them. The nightly stays red until the rest are dealt with — see
> "What this does not fix" below. The slow-test lane *is* fully unbroken by this
> PR. Title and body updated accordingly.

Failures that have kept `main`'s post-merge lanes red. None of them run on a
PR, which is why they went unnoticed: `slow-tests` is gated on
`push`/`schedule`/the `run-slow-tests` label, and Mutation Testing is nightly.

| Failure | Lane | Red since | Fixed here |
|---|---|---|---|
| `test_regeneration_reproduces_the_committed_files` | slow-tests, all 3 OSes | Aug 30 | ✅ |
| `test_no_skipped_fence_hides_a_working_command[partition.md]` | slow-tests, all 3 OSes | ~Sep 5 | ✅ |
| The 3 front-door surface guards, under mutmut | Nightly | Sep 5 | ✅ |
| 40 further mutmut-incompatible tests | Nightly | unknown, ≥ Aug 28 | ❌ |

The fourth slow-lane failure — the Windows-only `docs/guide/geojson.md`
`UnicodeEncodeError` — is a real product bug and is fixed separately in #907.

### 1. Regenerate the canonical dataset on DuckDB 1.5.5

DuckDB stamps its own version into every parquet footer's `created_by`. The
committed files were generated on 1.5.1; #778 moved the floor to
`duckdb>=1.5.2` and the lock now resolves 1.5.5, so a regeneration could no
longer reproduce the committed bytes.

Rows, schema and `geo` metadata are byte-for-byte what was committed — checked
with `pyarrow.Table.equals` and `schema.equals(check_metadata=True)`. Only the
writer's version stamp moved. `examples/data/` is re-mirrored to match.

Worth a follow-up decision: as written, this test goes red on **every** DuckDB
patch bump. Its own `_drift_diagnosis` already separates "gpio changed" from
"the writer's stamp changed", so it could tolerate the latter instead of
requiring a regeneration commit each time.

### 2. Split the kdtree fence in `partition.md`

The fence was skipped as *"the 766-row sample is too small to partition
meaningfully"*, but that is true of only two of its four commands:

| command | on the 766-row sample |
|---|---|
| `gpio partition kdtree input.parquet output/` | **exit 0** — auto picks 2 partitions, warns |
| `... --partitions 32` | exit 1 |
| `... --partitions 16 --exact` | exit 1 |
| `... --hive --verbose` | **exit 0** |

Auto mode became this permissive with the kdtree auto-mode work (#855). Split
into a `menu` fence for the two that work — `menu` runs each statement in a
fresh workdir, so both writing `output/` is fine — and a skip, with an accurate
reason, for the explicit-count pair.

The Python tab has the same shape but is deliberately not split: the guard only
inspects `bash` fences, which is #901.

### 3. Let the front-door surface guards step aside under mutmut

`mutmut run` rewrites every file it mutates. Each function becomes
`x_<name>__mutmut_orig` plus one `x_<name>__mutmut_<n>` per mutant — so even a
private `_foo` mangles into a **public** `x__foo__mutmut_1` — and a trampoline
preamble binds `Annotated`, `Callable`, `ClassVar` and `MutantDict` at module
scope.

The guards read `dir()` to describe gpio's real public surface, and under
mutmut there is no real package to describe, only a rewritten copy. Asserting
on it kills no mutant and fails the **stats** phase, which aborts the run
before a single mutant is measured.

This is the second time it has happened: the nightly previously died on
`test_examples_mirror_is_byte_identical` failing from mutmut's `mutants/` cwd.

So the guards now step aside when `MUTANT_UNDER_TEST` is set (mutmut exports it
for every phase; nothing else does). The skip lives **inside the two
`dir()`-reading helpers** rather than on each test, so a guard added later
cannot forget it. Every other lane, PRs included, still runs them.

<details>
<summary>Why not filter the injected names instead</summary>

That was the first attempt, and it is worth recording why it was dropped.
Subtracting the four `typing` names cleared `test_core_leaks_through_ops_are_pinned`
and immediately hit `test_one_sided_public_names_are_acknowledged` — mutmut runs
pytest with `-x`, so only the first failure was ever visible. Chasing artifacts
one at a time means re-fixing this whenever mutmut's mangling changes, for a
result that is meaningless anyway: none of these guards can kill a mutant.

</details>

## What this does not fix

Overriding mutmut's `-x` (it stops at the first failure, which is how this
stayed invisible) gives the full picture: **43 failures, 6003 passed**. Four
classes, none of them a gpio bug — every one is a test that cannot describe a
package mutmut has rewritten, or read files mutmut did not copy:

| Count | Class | Example |
|---|---|---|
| 16 | `KeyError: 'MUTANT_UNDER_TEST'` — the test clears `os.environ` to exercise credential detection, deleting mutmut's own variable, and the trampoline reads it unconditionally | `test_upload.py::TestCredentialValidationFunctions::*` |
| 14 | Reads a repo path absent from mutmut's `also_copy` (`examples/`, `docs/`) | `test_notebook_signal.py::*`, `test_canonical_dataset.py::test_examples_mirror_is_byte_identical` |
| 10 | Spawns a subprocess or isolates an import; the child inherits `MUTANT_UNDER_TEST` and resolves the mutated package | `test_secondary_geometry_carriers.py::test_secondary_carrier_is_independent_of_geoarrow_import` |
| 3 | Asserts on gpio's own source text, which mutmut rewrote | `test_geoparquet_versions.py::test_both_write_paths_reference_the_constant` |

3 of the 43 are the surface guards this PR fixes. The remaining 40 are a real
piece of work — the `also_copy` ones are a config line, but the 16
environment-clearing tests need a decision (preserve `MUTANT_UNDER_TEST` across
the clear, or narrow the patching), and that is not a call to smuggle into this
PR. Happy to take it on as a follow-up if wanted.

## Breaking Changes

## Related Issue(s)
- Follow-ups noted above, not addressed here: #901 (extend the fence guard to
  python fences), and the DuckDB-stamp question in §1.

## Checklist

<sub>Tests: `tests/test_canonical_dataset.py` 29 passed ·
`tests/test_docs_examples_coverage.py` + `..._harness.py` 65 passed ·
`docs/guide/partition.md` 19 passed, 15 skipped (the newly-unskipped menu fence
runs) · `tests/test_api.py` 237 passed, including a new regression test that
injects mutmut's four names into `ops` and asserts the pin holds ·
`uv run mutmut run` locally reproduced the stats-phase abort on `main`, and
`MUTANT_UNDER_TEST=stats uv run pytest tests/test_api.py` confirms exactly the
three surface guards skip and the other 233 still run. The full `mutmut run` has since
completed: 43 failures, 3 of which this PR removes. The other 40 are catalogued
above.</sub>


cholmes:
**Correcting the claim I opened this with.**

I said this unbreaks the nightly mutation lane. It does not, and the title and body are now fixed.

I opened it while a full `mutmut run` was still going, on the strength of a targeted check. That run has finished, with mutmut's `-x` overridden so every failure is visible rather than just the first: **43 failed, 6003 passed**. This PR clears 3 of the 43. The nightly stays red.

The 40 remaining fall into four classes, none of them a gpio bug:

| Count | Class |
|---|---|
| 16 | `KeyError: 'MUTANT_UNDER_TEST'` — the test clears `os.environ` to exercise credential detection, which deletes mutmut's own variable; the trampoline reads it unconditionally |
| 14 | Reads a repo path missing from mutmut's `also_copy` (`examples/`, `docs/`) |
| 10 | Spawns a subprocess or isolates an import, so the child inherits `MUTANT_UNDER_TEST` and resolves the mutated package |
| 3 | Asserts on gpio's own source text, which mutmut rewrote |

Worth noting what this means for the nightly's history: because `-x` only ever showed the first failure, each fix has revealed the next one behind it. The lane has been failing for a long time and every previous fix looked like the last one needed. The 43-item list is the first look at the whole thing.

The `also_copy` class is close to a config line. The 16 environment-clearing tests need an actual decision — preserve `MUTANT_UNDER_TEST` across the clear, or narrow what those tests patch — which I did not want to smuggle into this PR.

cholmes:
Filed #910 for the wider question, with the full evidence.

Short version: fixing the remaining 40 stats-phase failures would not get the nightly to a kill rate. It has never produced one — the 119 green runs before 2026-07-06 were a no-op (`--no-progress` is not a mutmut 3 flag; it errored in 17s and 0/0 was read as a pass), and the 45 runs between 2026-07-08 and 08-27 marked "cancelled" each ran the full 5h and hit `timeout-minutes: 300`. So clearing the 43 returns the lane to timing out, not to a number.

The scale is the reason: `source_paths = ["geoparquet_io/"]` is 63,069 mutants across 114 files, against a 5-hour budget.

Nothing in this PR changes — the 3 guards it fixes are correct regardless of how #910 is decided, and the two slow-lane fixes are unrelated to mutation testing. I have just stopped claiming it unbreaks the nightly.

cholmes:
**Slow lane on this PR: ubuntu ✅ macOS ✅**, both of the failures this PR targets are gone.

Windows shows two, neither from this PR:

1. `docs/guide/geojson.md` — the `UnicodeEncodeError`, fixed in #907. Confirmed there: that fence now passes on the Windows leg of #907, whose only remaining failures are the two *this* PR fixes. The two PRs cover each other's Windows gap.
2. `test_journey_08_admin_divisions_then_partition` — `IO Error: Could not move file: Access is denied.` while downloading the Overture cache. A Windows file-lock on the cache move, not something this PR touches: the journey reads `tests/data/places_test.parquet`, which is unmodified here, and the traceback is entirely in `admin_datasets._download_per_level_caches`.

   It is environment-dependent rather than a clean regression — on the 2026-09-08 scheduled run on `main` the same test was *skipped* on Windows, and here it ran and hit the lock. That is the cache-state sensitivity #739 is about.

cholmes:
Closing this, as it's not actually working. We should revisit mutmut stuff and figure out what we want to do.

Issue #890, opened 2026-09-05
fix(reproject): secondary geometry columns lose geometry_types, making the output unreadable by DuckDB


cholmes:
Observed during the adversarial review of #884 (findings from an AI review agent; pre-existing on main, untouched by that PR's diff — filed separately per the repo's cross-cutting convention).

`gpio convert reproject` drops `geometry_types` from a SECONDARY geometry column's metadata; DuckDB then refuses to read the output outright: `Geoparquet column 'other_geom' does not have geometry types`. The primary column's entry is rebuilt correctly; the secondary's carried entry loses the key somewhere on the reproject metadata path.

Repro shape: two-geometry-column GeoParquet → `gpio convert reproject in.parquet out.parquet --dst-crs EPSG:3857` → `duckdb -c "SELECT * FROM read_parquet('out.parquet')"` errors.

Related: #884 scoped its edges-drop to the transformed column only; the same "touch only what you transformed" discipline applies here — the secondary's carried entry should survive intact (it isn't reprojected at all). Also worth an eventual decision: whether reproject should refuse or warn on multi-geometry inputs, since only the primary is transformed.

Refs #884, #601.

PR #911, opened 2026-09-08
fix(aggregate): cut grid cell rings at the antimeridian into MultiPolygons

cholmes:
## What changed

Reworked to **Option B**: a cell ring that crosses the antimeridian is **cut at ±180 into a `MultiPolygon`** (RFC 7946 §3.1.9), so the output is spec-clean rather than carrying coordinates outside the valid range. The earlier approach in this PR — unwrapping each vertex relative to the ring's first vertex — is gone; review found it fixed the common case and broke two others.

One invariant, stated once in the shared builder for every grid scheme:

> the polygon written for a cell is **valid, contiguous, and inside `[-180, 180]`**.

Getting there takes four steps, all in `core/process/aggregate/grid_common.py`:

1. **Cumulative unwrapping.** Each vertex is placed relative to the *previous* one, with the 360° steps accumulating along the ring. Relative-to-first is only correct while every vertex stays within 180° of the first, which a polar ring does not — there the `> 180` test fires on vertices that were never wrapped and tears a ring that was intact.
2. **Span guard.** The shift is kept only when it *narrows* the ring's longitude span, so a ring that legitimately spans more than 180° is left alone.
3. **Explicit pole seam.** A ring whose longitude winding comes to ±360 encircles a pole; it spans every longitude, so no unwrapping closes it. It is closed through a seam that runs up to ±90 and back, covering the polar cap instead of self-intersecting.
4. **Cut at ±180.** Whatever still crosses the line is intersected with the two half-world boxes and reassembled, giving a MultiPolygon with one part either side.

Consequences worth calling out for reviewers:

- **`geometry_types` now reads `["Polygon", "MultiPolygon"]`** (or just `["MultiPolygon"]`) for any dataset reaching the antimeridian. It is computed from the data, so no code change was needed — but downstream consumers have to expect both. Documented in `docs/guide/process-aggregate.md`.
- **The `geo` bbox uses the RFC 7946 §5.2 wrap form (`xmin > xmax`)** for a dataset astride the seam. A plain min/max over cut geometry reads `[-180, …, 180, …]`, claiming the whole globe for a Fiji-sized dataset; the parts, not the whole geometries, are what make the crossing visible, so `write_geoparquet_table` grew an optional `geo_bbox` and the aggregate/overview writers compute one. gpio's own validator already reads this form (#876/#930). Data that does not reach both edges never pays for the part-level scan and is reported exactly as before.
- **`process overview` inherits all of it** — grid rollups regenerate parent geometry through the same builder, and a coarser parent cell straddles the seam more readily than its children.
- The scheme descriptors now hand the builder a **GEOMETRY** rather than a vertex list. That lets H3 read `h3_cell_to_boundary_wkb` instead of parsing WKT, and `--out-geometry centroid` build no boundary at all.

## Why

Found on a global aggregate of NASA FIRMS fire detections: one cell near Fiji covered 359.413° of longitude and rendered as a stripe across the map at 16°S. Review of the first fix then turned up three more problems it did not solve, and one it caused.

### Global sweep — 50,000 scattered points, h3 and a5, resolutions 0/1/3/5/7/9

| | `main` | this PR |
|---|---:|---:|
| invalid cell polygons | **82** | **0** |
| cells outside `[-180, 180]` | **305** | **0** |
| cell ids and row counts | — | identical in all 12 runs |

Per-scheme, cells with `ST_IsValid = false` / outside `[-180, 180]`:

| | h3 r0 | r1 | r3 | r5 | r7 | r9 | a5 r0 | r1 | r3 | r5 | r7 | r9 |
|---|---|---|---|---|---|---|---|---|---|---|---|---|
| `main` invalid | 11 | 10 | 46 | 18 | 2 | 1 | 2 | 0 | 0 | 0 | 0 | 0 |
| `main` out of range | 0 | 0 | 0 | 0 | 0 | 0 | 4 | 9 | 32 | 126 | 104 | 30 |
| this PR, both | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 | 0 |

Total spheroid area recovers the whole sphere wherever the cells cover it — a5 r0 goes **448.6M km² → 510.07M km²**, which is 4πR² to six figures.

### The A5 polar cells

These are the ones the relative-to-first unwrap regressed valid → invalid, and the ones `main` has always written out of range. At a5 r1 the 9 cells that leave `[-180, 180]` on `main` all become in-range MultiPolygons, with per-cell area held to within 0.03%:

```
cell                   main xmin/xmax   this PR              main area   this PR area
360287970189639680      87.0 / 267.0    MULTIPOLYGON ±180    8501090     8503440
14195346025471803392   123.0 / 267.0    MULTIPOLYGON ±180    8501090     8497008
72057594037927936     -201.0 /  -93.0   MULTIPOLYGON ±180    8501090     8498755
… 6 more, all in range
total                                                      510065622   510065622   km²
```

### The fixtures the new tests use, run against `main` and against this PR

`widest_part` is the widest *part* after `ST_Dump`, so a seam cut cannot hide a tear behind a globe-spanning envelope.

```
                              cells  invalid  out_of_range  widest_part  validate_failed  bbox_wrap
main    fiji   h3 r1              1        0             0       355.82                0      False
main    fiji   h3 r3              2        0             0       359.41                0      False
main    fiji   a5 r1              1        0             1        36.00                1      False
main    fiji   a5 r3              1        0             1         7.24                1      False
main    poles  h3 r1              3        2             0       321.18                0      False
main    poles  h3 r3              4        1             0       315.60                0      False
main    poles  a5 r0              2        2             2       358.74                1      False
main    poles  a5 r1              5        0             1       216.00                1      False
main    poles  a5 r2              5        0             1       189.65                1      False

this PR fiji   h3 r1              1        0             0         4.47                0      True
this PR fiji   h3 r3              2        0             0         1.23                0      True
this PR fiji   a5 r1              1        0             0        21.00                0      True
this PR fiji   a5 r3              1        0             0         4.77                0      True
this PR poles  h3 r1              3        0             0       360.00                0      False
this PR poles  h3 r3              4        0             0       360.00                0      False
this PR poles  a5 r0              2        0             0       360.00                0      False
this PR poles  a5 r1              5        0             0       273.00                0      False
this PR poles  a5 r2              5        0             0       273.00                0      False
```

Every new extension-backed test fails on `main`. `widest_part = 360.00` on the pole rows is the seam doing its job: a cell that holds a pole spans every longitude, so its polygon runs the full width and reaches ±90.

### `--out-geometry both`

The centroid comes from the grid's own cell-centre function, which reports lon/lat in range; the polygon is now in range too, so it holds its own centroid. Zero centroids outside their polygon over 15,738 h3 and 11,198 a5 cells (was: e.g. `839b5dfffffffff` with polygon lon `[-180.76, -179.55]` and centroid `179.84`).

### Performance

The 15× regression the review measured is gone — this is at parity or better, because reading the WKB boundary instead of parsing WKT offsets a repair that only the crossing cells pay for. Warm, 50,000 points, whole `aggregate_by_*` call:

| | h3 r5 | h3 r7 | h3 r9 | a5 r5 | a5 r7 | a5 r9 |
|---|---|---|---|---|---|---|
| `main` | 0.44s | 0.45s | 0.47s | 0.26s | 0.35s | 0.38s |
| this PR | 0.32s | 0.35s | 0.39s | 0.39s | 0.46s | 0.47s |

Isolated to the geometry SQL over 200k res-6 cells: `main` 0.58s, the previous approach in this PR 0.75s, now **0.14s**.

## Tests

`tests/test_process_aggregate_antimeridian.py` is rewritten. The tests it replaces asserted only `max(ST_XMax − ST_XMin) < 180` on a Fiji fixture, were marked `slow` **and** `network` (so the slow lane deselects them and the fast lane excludes them — only the non-blocking network job ever ran them), and the A5 one passed byte-identically on `main`.

**Fast lane, no community extension, no network** — 25 tests where diff-cover reads. A scheme whose boundary is a literal WKT ring drives the whole builder on plain DuckDB + spatial: real h3 rings for the torn and pole-enclosing cases, synthetic ones for the rest. Plus assertions on the generated SQL's shape (the guard, the cut, the one-evaluation-per-cell property, and that `--out-geometry centroid` builds no boundary) and on the wrap-form bbox.

**Extension lane** — 13 tests over both grids: validity, `ST_XMin >= -180 AND ST_XMax <= 180`, part width, `validate_geoparquet(..., validate_data=True)`, pole fixtures for both schemes, the `--out-geometry both` frame, the wrap-form bbox, and a `process overview` case (`rollup.py` inherits this through `GRID_SCHEMES` and had no coverage of it at all).

```
$ uv run pytest tests/test_process_aggregate_antimeridian.py -q
38 passed in 3.98s

$ uv run pytest tests/test_process_aggregate*.py tests/test_process_overview.py \
    tests/test_validate*.py tests/test_overview_levels.py -q -m "not network and not slow"
599 passed, 5 skipped, 57 deselected

$ uv run pytest -n auto -q -m "not slow and not network and not meta"
6478 passed, 75 skipped, 9 xfailed in 165.74s

$ uv run pytest -q -m meta
203 passed in 167.55s

$ uv run lint-imports
Contracts: 4 kept, 0 broken.

$ uv run xenon --max-absolute=E --max-modules=D --max-average=C geoparquet_io tests
$ uv run ruff check . && uv run vulture && uv run deptry .
```

Rebased onto `main` (the branch was ~10 commits behind and its CI predated #916, which is why it reported 2 import-linter contracts rather than 4).

### DIFF
--- a/docs/guide/process-aggregate.md
+++ b/docs/guide/process-aggregate.md
@@ -543,6 +543,30 @@ Regardless of the chosen bucket scheme, every output file contains:
 | `count_other` | BIGINT | Count for categories beyond `--breakdown-limit` |
 | `geometry` | WKB | Bucket polygon or centroid (absent when `--out-geometry none`) |
 
+### Cells at the antimeridian and the poles
+
+A grid cell is a region of the sphere, and a few of them cannot be drawn as a
+single flat polygon. Those are written the way [RFC 7946
+3.1.9](https://datatracker.ietf.org/doc/html/rfc7946#section-3.1.9) prescribes:
+
+- a cell straddling the antimeridian is **cut at ±180 into a `MultiPolygon`**,
+  with one part either side of the line, rather than a ring that stretches
+  across the whole map;
+- a cell holding a pole is closed through an explicit seam that runs up to
+  ±90, so it covers the polar cap instead of self-intersecting;
+- every coordinate written stays inside `[-180, 180]`, which is what
+  `gpio check spec` requires of a geographic CRS.
+
+So `geometry_types` in the output's GeoParquet metadata reads
+`["Polygon", "MultiPolygon"]` (or just `["MultiPolygon"]`) for any dataset that
+reaches the antimeridian, and consumers must expect both. A dataset that sits
+astride the line — rather than spanning the globe — also gets the RFC 7946 5.2
+wrap-around `bbox`, with `xmin > xmax`.
+
+`--out-geometry centroid` and `both` take the centroid from the grid library's
+own cell-centre function, which reports lon/lat in range, so it always falls
+inside the polygon on the same row.
+
 ---
 
 ## Common Options
--- a/docs/guide/process-overview.md
+++ b/docs/guide/process-overview.md
@@ -53,6 +53,8 @@ Outputs are written next to the input (override with `--output-dir`):
 
 Cells roll up by **true hierarchy** — `a5_cell_to_parent` / `h3_cell_to_parent` for grids, the ISO country prefix of `admin_code` (`US-CA` → `US`) for admin — and parent geometry is regenerated from the parent cell id (grids) or cached Overture country polygons (admin).
 
+A coarser parent cell straddles the antimeridian more readily than its children, so grid rollups go through the same seam handling as `gpio process aggregate`: crossing cells are cut into `MultiPolygon`s at ±180 and pole cells are closed through a seam. See [Cells at the antimeridian and the poles](process-aggregate.md#cells-at-the-antimeridian-and-the-poles).
+
 | Column | Rollup | Exactness |
 |--------|--------|-----------|
 | `count` | sum | exact |
--- a/geoparquet_io/core/common.py
+++ b/geoparquet_io/core/common.py
@@ -1600,6 +1600,7 @@ def _apply_geoparquet_metadata(
     verbose: bool = False,
     edges: str | None = None,
     geometry_info: dict | None = None,
+    geo_bbox: list[float] | None = None,
 ):
     """
     Apply GeoParquet metadata to an Arrow Table based on version.
@@ -1628,6 +1629,11 @@ def _apply_geoparquet_metadata(
             - "primary": primary geometry column name
             - "secondary": list of secondary geometry column names
             - "metadata": dict mapping column names to their metadata
+        geo_bbox: Pre-computed [xmin, ymin, xmax, ymax] for the primary column,
+            used instead of the plain min/max taken off the data. A caller that
+            knows the data crosses the antimeridian passes the RFC 7946 5.2
+            wrap form (xmin > xmax) here, which cannot be recovered from
+            per-geometry extents alone.
 
     Returns:
         pa.Table: Table with GeoParquet metadata applied
@@ -1775,7 +1781,7 @@ def _apply_geoparquet_metadata(
     if "geometry_types" not in col_meta:
         col_meta["geometry_types"] = _compute_geometry_types(table, geometry_column, verbose)
 
-    computed_bbox = _compute_bbox_from_data(table, geometry_column, verbose)
+    computed_bbox = geo_bbox or _compute_bbox_from_data(table, geometry_column, verbose)
     if computed_bbox:
         col_meta["bbox"] = computed_bbox
 
@@ -3197,6 +3203,7 @@ def write_geoparquet_table(
     verbose: bool = False,
     profile: str | None = None,
     edges: str | None = None,
+    geo_bbox: list[float] | None = None,
 ) -> None:
     """
     Write a PyArrow Table to a GeoParquet file with proper metadata.
@@ -3217,6 +3224,9 @@ def write_geoparquet_table(
         profile: AWS profile name (S3 only, optional)
         edges: Edge interpretation, "spherical" or "planar" (default None = planar).
                Use "spherical" for data from BigQuery or other S2-based sources.
+        geo_bbox: Pre-computed [xmin, ymin, xmax, ymax] for the geometry column.
+               Pass the RFC 7946 5.2 wrap form (xmin > xmax) for data that
+               crosses the antimeridian; None computes a plain extent.
     """
     # Auto-detect geometry column if not provided
     if geometry_column is None:
@@ -3285,6 +3295,7 @@ def write_geoparquet_table(
                 custom_metadata=None,
                 verbose=verbose,
                 edges=edges,
+                geo_bbox=geo_bbox,
             )
 
         # Write to disk with proper settings
--- a/geoparquet_io/core/process/aggregate/by_a5.py
+++ b/geoparquet_io/core/process/aggregate/by_a5.py
@@ -22,13 +22,15 @@ A5_SCHEME = GridScheme(
     max_resolution=30,
     default_column=DEFAULT_A5_COLUMN_NAME,
     key_template="a5_lonlat_to_cell(ST_X({pt}), ST_Y({pt}), {res})",
-    boundary_template="a5_cell_to_boundary({cell})",
-    latlng_template="a5_cell_to_lonlat({cell})",
-    # a5_cell_to_boundary returns DOUBLE[2][]; close the ring and build a polygon.
-    poly_wkb_template=(
-        "ST_AsWKB(ST_MakePolygon(ST_MakeLine("
-        "list_transform(list_append({bnd}, {bnd}[1]), p -> ST_Point(p[1], p[2])))))"
+    # a5_cell_to_boundary returns an open DOUBLE[2][] ring that stays contiguous
+    # but lets longitudes run outside [-180, 180] (267.0 is a real value), so the
+    # shared builder still has to bring the polygon back into range.
+    boundary_template=(
+        "ST_MakePolygon(ST_MakeLine(list_transform("
+        "list_append(a5_cell_to_boundary({cell}), a5_cell_to_boundary({cell})[1]), "
+        "p -> ST_Point(p[1], p[2]))))"
     ),
+    latlng_template="a5_cell_to_lonlat({cell})",
     # a5_cell_to_lonlat returns [lon, lat].
     centroid_wkb_template="ST_AsWKB(ST_Point({ll}[1], {ll}[2]))",
 )
--- a/geoparquet_io/core/process/aggregate/by_h3.py
+++ b/geoparquet_io/core/process/aggregate/by_h3.py
@@ -23,10 +23,11 @@ H3_SCHEME = GridScheme(
     default_column=DEFAULT_H3_COLUMN_NAME,
     # h3_latlng_to_cell_string takes (lat, lng) -> note Y before X.
     key_template="h3_latlng_to_cell_string(ST_Y({pt}), ST_X({pt}), {res})",
-    # h3_cell_to_boundary_wkt returns a WKT polygon directly.
-    boundary_template="h3_cell_to_boundary_wkt({cell})",
+    # The WKB form parses straight to a GEOMETRY -- no WKT round trip. Every
+    # vertex comes back wrapped into [-180, 180], so a cell straddling the
+    # antimeridian arrives torn; the shared builder repairs it.
+    boundary_template="ST_GeomFromWKB(h3_cell_to_boundary_wkb({cell}))",
     latlng_template="h3_cell_to_latlng({cell})",
-    poly_wkb_template="ST_AsWKB(ST_GeomFromText({bnd}))",
     # h3_cell_to_latlng returns [lat, lng]; ST_Point wants (lng, lat).
     centroid_wkb_template="ST_AsWKB(ST_Point({ll}[2], {ll}[1]))",
 )
--- a/geoparquet_io/core/process/aggregate/common.py
+++ b/geoparquet_io/core/process/aggregate/common.py
@@ -423,3 +423,65 @@ def build_breakdown_select(
                 other_clause = "TRUE"
             parts.append(f'COUNT(*) FILTER (WHERE {other_clause}) AS "count_other"')
     return ", ".join(parts)
+
+
+# Grid-cell output and the antimeridian
+# -------------------------------------
+# A cell that straddles the antimeridian is written cut into a MultiPolygon with
+# parts at both -180 and +180 (RFC 7946 3.1.9). Plain min/max over such a
+# dataset reports a bbox of [-180, ..., 180, ...] -- true, but it claims the
+# whole globe for data that only touches the seam. RFC 7946 5.2, which
+# GeoParquet's `bbox` follows, writes the crossing extent as xmin > xmax
+# instead; gpio's own validator already reads it that way (#876).
+_ANTIMERIDIAN_EPS = 1e-6
+
+_PLAIN_EXTENT_SQL = """
+SELECT min(ST_XMin(g)), max(ST_XMax(g)), min(ST_YMin(g)), max(ST_YMax(g))
+FROM (SELECT {geom_expr} AS g FROM {relation} WHERE {qcol} IS NOT NULL)
+"""
+
+# Widest longitude gap between the parts, found with a running max over the
+# sorted part extents -- the standard interval merge. The parts, not the whole
+# geometries: a cut cell's own extent already spans -180 to 180 and would hide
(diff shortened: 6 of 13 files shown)


cholmes:
Adversarial review findings — **recommend holding this one**. Findings below come from an AI review agent; everything was verified by running the real pipeline, not inferred.

## S1 — the output fails `gpio check spec`, and a compliant reader silently loses rows

Fiji fixture, `gpio process aggregate h3 … --resolution 3`, emitted ring:
```
POLYGON ((-179.860999 -16.781441, …, -180.447612 -15.776651, -180.760664 -16.316709, …))
```
`gpio check spec` on that file: `✗ coordinates outside valid range for CRS … min_x=-180.7607 < -180.0` → **23 passed, 1 failed**. On `origin/main` the same command gives **24 passed, 0 failed**. Third instance of the #890/#934 "gpio writes a file gpio rejects" shape.

It propagates past validation. The `geo` bbox becomes `[-180.760664, …]`, where GeoParquet v1.1.0 requires the RFC 7946 wrap form `[179.239336, …]` (xmin > xmax) for geographic CRS — the convention #876/#930 established. `gpio add bbox` then writes a covering with `xmin: -180.760664`, and an AOI query `bbox.xmin <= 179.9 AND bbox.xmax >= 179.2` returns **0 rows** while a cell genuinely covers that ground. `gpio convert geojson` also carries 4 out-of-range vertices out, violating RFC 7946 §3.1.9 — which validate's own docstring already cites as telling *producers* to cut.

## S1 — A5 polar cells regress from valid to invalid (the PR body says A5 is unchanged)

Verified false. North-pole fixture, res 5, cell `696369092382162944`:

| | main | this PR |
|---|---|---|
| `ST_IsValid` | **True** | **False** |
| width | 206.35° | **350.33°** |

Cause: relative-to-first unwrapping is only correct for rings spanning < 180° of longitude. Adjacent vertices at effectively the same longitude (`87` and `86.99999999999989`) land on opposite sides of the ±180 test and jump 360° apart on floating-point noise alone. Two A5 cells change at every resolution sampled (r1, r3, r5, r8), contiguous → torn each time. **Cumulative** unwrapping (each vertex relative to the previous, accumulating the offset) fixes this one: `is_valid=True`, x ∈ [-93, 267].

## S2 — pole-enclosing cells: H3 trades one broken cell for another

Res 5 north pole: `85032623fffffff` goes valid → **invalid** (max_x 187.72), while `8503263bfffffff` goes invalid → valid. Net zero, and the file still fails `check spec`. Cumulative unwrapping does not rescue this — a ring encircling a pole genuinely spans 360° and needs an explicit seam stitch or a MultiPolygon split.

## S3 — the tests cannot catch any of this, and never run in a blocking lane

Both assert only `max(ST_XMax − ST_XMin) < 180` on Fiji cells: no vertex coordinates, no `ST_IsValid`, no `validate_geoparquet`, no pole fixture, no row-count comparison. The A5 test **passes byte-identically on main** — it would not have caught the A5 regression. Both are marked `slow` *and* `network`: `slow-tests` selects `(slow or meta) and not network` so it skips them, the fast lane excludes them, and only the non-blocking `network-tests` job would run them (`gh pr checks` confirms both skipping). The green CI is not evidence here. diff-cover passed only because the changed lines are module-level constants executed at import.

## Cleared (checked, no finding)

CRS-blindness is **clean** — cell geometry is always built in lon/lat and projected input is reprojected to CRS84 before keying, so the ±180 logic never sees projected coordinates, and `_bbox_center_lon_sql` stays properly gated. No row or cell loss (2/2, 3/3, 4/4 across fixtures). The builder is correctly shared in `grid_common` rather than duplicated. 99 aggregate tests and 648 validate/spec tests pass.

## Pre-existing, tracked separately

A5 aggregate output **already** fails `check spec` on main (`min_x=-180.2016`, `max_x=180.4002`) because `a5_cell_to_boundary` returns continuous out-of-range coordinates that main passes through. This PR cites that as precedent for doing the same in H3 — but the precedent is itself a bug.

## Suggested shape

1. Cumulative unwrapping instead of relative-to-first (fixes the A5 regression).
2. Split the ring at ±180 into a MultiPolygon per RFC 7946 §3.1.9 and widen `geometry_types` accordingly — output then validates, GeoJSON export is legal, and the covering falls back into range.
3. Handle pole-enclosing rings explicitly (seam stitch), or detect and warn rather than emit a self-intersecting polygon.
4. Tests that assert `ST_IsValid`, vertex ranges and `validate_geoparquet`, with a pole fixture for **both** grids and an A5 case that fails on main.

Also: this branch is 10 commits behind main and its CI predates #916 — `lint-imports` reports 2 contracts here where main now enforces 4. Rebase before trusting the checks.

🤖 These findings were generated by an AI review agent.

cholmes:
Second review pass (data-integrity lens, AI review agent). Quantifies the earlier findings, clears the biggest worry, and adds three new ones — including a minimal fix that resolves the A5 regression **without** needing the cut-vs-unwrap decision.

## Cleared: no data loss, no measure drift

50,000 random global points, res 4, old vs new templates, both grids:
```
h3 res4: new_cells=43530 old_cells=43530 ids_equal=True measures_equal=True total_count 50000/50000
a5 res4: new_cells=3840  old_cells=3840  ids_equal=True measures_equal=True total_count 50000/50000
```
Identical cell-ID sets, identical `count`/`sum_v`/`mean_v` per cell. The separation holds in code — assignment is `key_template` on `__pt` before the `GROUP BY`, the boundary is derived from the cell id afterwards. Points at lon `180.0` and `-180.0` bin together at res 1/3/5/8, unchanged. `geometry_types` stays `['Polygon']` and `geometry_types_match_data` passes. Round-trip is clean in both pyarrow+shapely and DuckDB.

## A5 polar regression, quantified

Exactly the two pole-touching cells, every resolution, always valid → invalid:
```
A5 res1:    60 cells, changed=2, became invalid=2, area ratio new/old=3.73
A5 res2:   240 cells, changed=2, became invalid=2, ratio=4.59
A5 res3:   960 cells, changed=2, became invalid=2, ratio=4.51
A5 res4:  3840 cells, changed=2, became invalid=2, ratio=4.35
A5 res5: 11830 cells, changed=2, became invalid=2, ratio=3.97
```
The res-1 affected cell covers lat 52.75–90, so **any** global dataset aggregated at A5 res 1 hits it. A spatial join against those cells now over-selects by ~4×.

Worth knowing: res 0 escapes only because its pole rings sit at `max_delta` of **exactly** 180.0 against a strict `> 180`. That's a tie, not a design — any float drift in a future `a5` build flips res 0 into the same corruption.

## H3 out-of-range frequency

Sweeping 83,351 H3 cells across res 0,1,3,5,7,9 globally: **765 (0.92%)** carry out-of-range longitudes after the fix, 0 before. On the 50k res-4 fixture, 133 of 43,530. So roughly one cell in a hundred of any global H3 aggregate — and one is enough to fail the file.

Also: gpio does **not** emit the legitimate `xmin > xmax` antimeridian bbox form. `write_geoparquet_table` takes a plain min/max over unwrapped coordinates, giving `[-181.878, …, 190.078, …]`. The #876 validator already reads the wrap form correctly; the aggregate writer just doesn't produce one. Note `bbox_valid_geometry` **passes** on `xmax=197.25` — only the data-level `coordinates_valid_for_crs` check catches it.

## New — `--out-geometry both` splits polygon and centroid into different longitude frames

The polygon is unwrapped; the centroid still comes from `h3_cell_to_latlng`, wrapped into [-180, 180]:
```
839b5dfffffffff  poly lon[-180.7607, -179.5476]  centroid lon=179.8438  inside=False
```
The cell's own centroid falls outside its own polygon's longitude range. Anything pairing the two columns — label placement, a sanity join, point-in-polygon — breaks on exactly the cells this PR touches.

## New — blast radius is wider than the title

`core/process/overview/rollup.py:23-32,82` builds `GRID_SCHEMES = {"a5": A5_SCHEME, "h3": H3_SCHEME}` and calls the same `wrap_grid_geometry`, so **`gpio process overview` inherits both the H3 fix and the A5 polar regression**, with no test added there.

## New — ~15× slower boundary construction

`ST_Dump(ST_Points(ST_ExteriorRing(ST_GeomFromText(...))))` per cell, 200,000 res-6 cells: **0.64s → 9.39s**. Bounded by cell count, so ≈ +23s at the 500,000-cell `max_cells` default. Not a blocker; the same open ring is obtainable without a WKT parse plus dump-and-reassemble.

## Minimal fix for the regression, independent of the design question

Apply the shift only when it **reduces** the ring's longitude span:
- H3 crossing cells: 359.4 → 1.2 — kept
- A5 polar rings: 216 → 357.5 — rejected, old ring preserved
- H3 pole cells: 315.6 → 279.9 — kept, no worse

A cheap guard over values already in hand, no per-scheme branching, so the "state the invariant once" goal survives. That removes the A5 S1 entirely and leaves only the deliberate question of whether aggregate output should carry longitudes past ±180 at all.

## Test note

All four new tests are `slow` + `network`, so `-m "not slow and not network and not meta"` collects **0 selected / 4 deselected** — yet they pass in 0.88s offline with cached extensions, so the markers cost coverage without buying isolation. The assertion `max(ST_XMax - ST_XMin) < 180` is precisely what the unwrap minimises, so it cannot detect any of the above. `ST_XMin >= -180 AND ST_XMax <= 180`, or a `validate_geoparquet(..., validate_data=True)` assertion, would have caught the spec failure immediately.

🤖 These findings were generated by an AI review agent.

cholmes:
## Option B verified — merging

I refused to merge the previous approach because it traded one class of broken geometry for another. Every one of those objections is now resolved. The numbers below are **my own runs**, not the rework report's: a 57,000-point global fixture deliberately loaded at the seam (3,000 points in 179.5–180.5) and at both poles (2,000 each), aggregated through the real CLI on this branch and on a worktree at `main` (`3db4253`).

### The regressions I blocked on

| | main | this branch |
|---|---:|---:|
| h3 r1 invalid cell polygons | 10 | **0** |
| h3 r5 invalid cell polygons | 205 | **0** |
| a5 r1 cells outside [-180,180] | 9 | **0** |
| a5 r5 cells outside [-180,180] | 132 | **0** |
| **a5 r0 invalid / out-of-range** | **2 / 4** | **0 / 0** |
| cell counts (all runs) | 842 / 60 / 52448 / 14303 | identical |

The a5 r0 row is the specific regression I named — "2 cells invalid at every resolution". It is gone, and no cell ID or row count moved anywhere.

`gpio check spec` now returns **24 passed, 0 failed** on every output I generated (h3 r5, a5 r5, a5 r0, and the `--out-geometry both` run). That was the other blocker: cells going out of range flipped `check spec` from pass to fail on gpio's own output.

**Centroid frame:** 45,408 cells at h3 r4 with `--out-geometry both` — **0 centroids outside their own polygon**, 0 out of range.

### Nothing gated this before; now something does

My fourth objection was that all four tests were `slow`+`network`, so the behaviour rode in the non-blocking lane. The suite is now **26 fast, extension-free tests + 13 extension tests**. I sabotage-verified the fast ones by neutralising `_seam_repair_sql` to return its input: **1 failed, 18 errors** in the blocking lane. They bite.

The extension tests remain `slow`+`network`, matching every other h3/a5 aggregate test in the repo. I agree with leaving that convention alone here — changing it is repo-wide — now that the fast lane carries the behaviour.

### One claim I could not reproduce

The rework report states the a5 r0 total spheroid area comes to 510,065,622 km², exactly 4πR². I could not confirm it: `ST_Area_Spheroid` returns **NaN for 9 of the 12** a5 r0 cells, including plain polygons rather than only the new MultiPolygons. That is **not a regression** — main returns NaN for the same 9 cells — so it is a pre-existing limitation of the area function on cells this large, and it does not affect the merge. But the figure as written does not reproduce by my method, so I am not repeating it as verified.

### Gates (my runs)

Fast suite **6492 passed**, 2 failures both on the known cold-run xdist flake list (`test_journeys`, `test_geojson_stream`) — 76 passed serially · `diff-cover --fail-under=90` → **98% on 67 changed lines** · `lint-imports` **4 kept, 0 broken** (the branch's old CI predated #916) · rebased clean onto `3db4253`.

Also worth noting for the changelog: output `geometry_types` now legitimately includes `MultiPolygon`, and a seam-crossing aggregate writes the RFC 7946 wrap-form bbox (`xmin > xmax`). Both are correct per GeoParquet 1.1.0 and are called out in the PR body.

Review, rework and verification by Claude Code agents; every claim above independently reproduced before merging.

Issue #896, opened 2026-09-05
ci: guide examples do not gate PRs — most of the docs lane only runs after merge


cholmes:
Measured on `main` @ da331b6 while fixing #894.

## The gap

`pyproject.toml` sets `testpaths = ["tests", "docs/guide"]`, so guide fences are collected by every pytest invocation and sorted purely by marker. There is no job named for them. That leaves them split three ways:

| Where they land | Count | Runs on a PR? |
|---|---|---|
| Fast lane (`not slow and not network and not meta`) | **57** | Yes — all 10 OS x Python jobs |
| `network-tests` job (`-m network`) | 30 | Non-blocking |
| Slow lane (`slow or meta`) + runtime-skipped | ~398 | **No** |

`slow-tests` runs on push-to-main, schedule, dispatch, or an opt-in `run-slow-tests` label — deliberately not on pull requests (`.github/workflows/tests.yml:254-265`, and the reasoning there is sound for the suite as a whole).

The consequence is specific: **the large majority of documented commands are not verified before merge.** A PR that breaks a guide example is discovered after it lands on `main`, or on the nightly schedule. That is the exact drift docs-as-tests (#667, PR #732) was built to catch, and catching it post-merge means the guide is briefly wrong on `main` and the fix is a second PR.

## Why it is cheap to close

Docs are only **57 of 6,409** fast-lane items — 0.9%. There is no 10x-matrix waste to reclaim and no reason to change `testpaths`; the marker sorting is right. The whole non-network docs set runs in **~92 s** with `-n auto` (271 s serial), measured locally:

```
uv run pytest -m "docs_example and not network" -q
235 passed, 221 skipped, 6883 deselected in 271.10s
```

## Proposal

One `docs` job in `.github/workflows/tests.yml` running `-m "docs_example and not network"` on **ubuntu-latest / 3.11 only**, on every PR. One runner, not ten. It gives a named signal — "docs is red" means a guide example broke, not a unit test — and closes the pre-merge gap.

Two constraints worth honouring:

[...]

PR #912, opened 2026-09-08
bump: version 1.4.0 → 1.5.0

cholmes:
## What changed

- `CHANGELOG.md`: the generated `## v1.5.0` section — 93 pull requests since v1.4.0, grouped by conventional-commit type, under hand-written highlights. `docs/CHANGELOG.md` is synced from it by the `doc-sync` hook.
- `pyproject.toml` and `uv.lock`: version 1.4.0 → 1.5.0.
- `scripts/release_notes.py`: a fix that was needed to generate this release at all — see below.

## Why 1.5.0 and not 2.0.0

Six changes carry `!`. Each is narrow, and five are behaviour catching up with what the documentation already said — `partition_by_admin`'s annotation always promised a dict, the partition/kdtree API was silently picking values the CLI refused to assume, `resolve_file_url()` was double-encoding URLs. None asks a user to redesign anything. A major bump signals "your code needs rework", which would misdescribe this release, and would spend 2.0.0 on a bug-fix batch while the Python API redesign that deserves it has not happened. `cz bump --dry-run` independently reads the range as MINOR.

The breaking section leads with that reasoning and gives each change a migration action.

## The `release_notes.py` fix

`apply_overrides()` raised on any entry in `release_title_overrides.json` whose pull request is not in the current tag range. That is correct for the first generated release and wrong for every one after it, because the file only ever grows: 22 of its 24 entries belong to v1.4.0 and earlier, so generating v1.5.0 failed outright.

Out-of-range overrides are now reported on stderr and skipped. The signal the guard existed for — a mistyped number silently rewriting nothing — is preserved, since a number in no release at all still shows up in that list.

## Verification

```
$ uv run python scripts/release_notes.py 1.5.0 --previous v1.4.0
note: 22 override(s) are not in this range and were skipped ...
## v1.5.0 (2026-09-08)
...
```

No `### Uncategorized`. All six `!` changes are in `### Breaking`, including #841 — whose `!` was lost in its own squash merge and which reaches the section only through the override #868 added.

- `uv run pytest -n auto -m "not slow and not network and not meta"` — 6506 passed, 25 skipped, 10 xfailed. The one failure was `test_geojson_stream.py::test_streaming_handles_broken_pipe`, the known xdist flake; it passes serially (64 passed).
- `uv lock --check` agrees with the committed lockfile; the `uv.lock` change is the single `geoparquet-io` version line, not a format rewrite.
- `pre-commit` ran on the commit and passed.

Merging this fires `publish.yml`, which tags `v1.5.0`, publishes to PyPI and creates the GitHub release. The release body is then replaced with this changelog section per step 7 of the release skill.

## Summary by CodeRabbit

- **New Features**
  - Restored S2 support with DuckDB 1.5.5 or later.
  - Added GML parsing for WFS extraction and directory input for sorting commands.
  - Added Azure storage support and full AWS credential-chain resolution.
  - Added GeoParquet 2.0 conformance checks and Python partitioning API parity.

- **Bug Fixes**
  - Improved CRS integrity across writing, GeoArrow handling, and validation.

- **Breaking Changes**
  - Updated partitioning behavior, URL handling, sorting defaults, and the minimum DuckDB version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,115 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 
+## v1.5.0 (2026-09-08)
+
+What is newly possible in this release:
+
+- **S2 works again.** The `geography` DuckDB extension is published once more, so `gpio add s2` and `gpio partition s2` run instead of stopping with an explanation. This is what raises the DuckDB floor to 1.5.5. ([#893](https://github.com/geoparquet/geoparquet-io/pull/893))
+- **`gpio extract wfs` reads GML.** It asks each service for the format that service advertises and parses whatever comes back, which makes the many WFS 1.x deployments serving only GML usable for the first time. ([#835](https://github.com/geoparquet/geoparquet-io/pull/835))
+- **`gpio sort column` and `sort quadkey` take a directory** of GeoParquet files, with `--allow-schema-diff` for directories whose files disagree — and they now refuse to re-read their own output as input, a loop that silently doubled row counts on every run. ([#852](https://github.com/geoparquet/geoparquet-io/pull/852), [#873](https://github.com/geoparquet/geoparquet-io/pull/873))
+- **`az://account/container` uploads and copies work.** They were being handed to a container-first URL parser and could not build a store at all. ([#871](https://github.com/geoparquet/geoparquet-io/pull/871))
+- **S3 credentials resolve through the full AWS chain**, so `aws sso login`, assume-role and `credential_process` profiles authenticate the same way gpio's DuckDB reads already did. ([#874](https://github.com/geoparquet/geoparquet-io/pull/874))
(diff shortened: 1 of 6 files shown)


cholmes:
Addressed in `11c4fc6`. The finding was correct and the way it escaped is worth recording.

`tests/test_release_notes.py::TestOverrides::test_an_override_for_an_absent_pull_request_is_an_error` still asserted the behaviour this branch replaces, so the meta lane was red. Rewritten to the contract now in place — the number is reported on stderr, the entries come back untouched, generation continues — plus a second case for the shape the fix is actually about: a stale override alongside an in-range one, which must not stop the in-range one from applying.

```
uv run pytest tests/test_release_notes.py   45 passed
uv run pytest -m meta                      203 passed
```

**How it got past both me and CI.** That module is `pytestmark = pytest.mark.meta`. The fast suite runs `-m "not slow and not network and not meta"`, so my local verification never collected it — and `slow-tests`, which runs the meta lane, deliberately does not run on pull requests. None of this PR's 17 green checks covered the file the PR changes. It would have gone red on `main` after merge.

That is the same structural gap as #896 (guide examples not gating PRs), one lane over: a PR can change tooling whose only tests run post-merge. Worth its own issue — I have not filed one, since the immediate fix is in.

*Reviewed and fixed by Claude (AI).*

PR #914, opened 2026-09-09
refactor(cli): split cli/main.py, starting with the shared helpers and the sort group

cholmes:
## What changed

First step of splitting `geoparquet_io/cli/main.py` (7,730 lines) into per-group
modules. Two commits, no behaviour change.

**1. `refactor(cli): give the shared CLI helpers a home outside main.py`**

New `geoparquet_io/cli/_shared.py` holding the two helpers used by more than one
command group:

| helper | why it is shared |
|---|---|
| `_activate_s3` | 48 call sites, spread across most groups |
| `create_default_group` | the factory behind `check`, `extract` and `inspect`'s default-subcommand groups |

Why a new module rather than `decorators.py`: `parse_row_group_options`,
`GlobAwareCommand` and `SingleFileCommand` already live in `decorators.py` and
were **not** moved — that file is the home for reusable Click *option* decorators
and the `cls=` command classes, and it stays that way. `_shared.py` takes the
group-neutral *runtime* plumbing that is neither. Both are things a group module
imports; neither may live in `main.py`, because `main.py` imports the groups.

Helpers used by exactly one group stay in `main.py` for now and will travel with
their group: `OptionalIntCommand`, `ConvertDefaultGroup`, and the three
`create_default_group(...)` instances.

**2. `refactor(cli): move the sort group into its own module`**

New `geoparquet_io/cli/commands/` package, and `sort.py` inside it holding the
`sort` group and its four commands (`column`, `hilbert`, `quadkey`, `str`).

### The registration pattern the next ten PRs copy

```python
# geoparquet_io/cli/commands/<group>.py
@click.group()          # NOT @cli.group()
def <group>(ctx): ...

# geoparquet_io/cli/main.py
from geoparquet_io.cli.commands.<group> import <group>   # with the other imports
cli.add_command(<group>)                                 # in the registration block
```

Explicit, one line per group, all in one block under the root group. Not
self-registration: `@cli.group()` inside the group module would need `cli.main`,
which imports the group — a cycle. And `grep add_command geoparquet_io/cli/main.py`
lists the whole tree, which an import-time side effect would not.

`main.py` re-exports `sort`, so `from geoparquet_io.cli.main import sort` (used by
`tests/test_sort.py` and others) still resolves. The 330 moved lines are
byte-identical to their previous form apart from the group's decorator line —
verified programmatically, not by eye.

Two test files needed the patch site renamed, because they patch the CLI's `*_impl`
alias in the module that resolved it: `test_write_memory_forwarding` and
`test_cli_api_call_parity_scaffold` now carry the owning module per case
(`cli_main` or `cli_sort`), so the next group move is a one-line edit in each.
Three pre-commit hook file patterns (`doc-sync`, `validate-claude-md`,
`check-api-for-cli`) were widened to `geoparquet_io/cli/commands/`, so a change
confined to a group module still triggers them.

## Why

Deep Review item 3.2. A 7,730-line module is the single worst file in the repo to
review, test or navigate; every group PR after this one is mechanical once the
shape is settled. This PR is the pilot: it decides where shared helpers live and
how a group attaches itself, so the shape matters more than the 381 lines removed.

Explicitly **not** in scope (later steps of 3.2): no `@gpio_command` decorator, no
uniform exception mapping, no touching the 48 hand-rolled `_activate_s3` call
sites, and no moving `MultiFileCheckRunner` or the `_*_impl` functions into
`core/` — that belongs with the `check`/`inspect` group PRs.

## Verification

**The CLI surface snapshot is unchanged — not re-recorded.**

```
$ git diff --stat origin/main -- tests/data/cli_surface.json
                                     (no output)
$ uv run pytest tests/test_cli_surface.py
74 passed in 0.45s
```

**Identical output, before and after.** `gpio --help`, `gpio sort --help` and
`gpio sort hilbert --help` all diff clean against captures taken at
`origin/main`. A real run:

```
$ gpio sort hilbert tests/data/unsorted.parquet out.parquet --verbose
before sha256: 1dc5a7f061a751cfeea97e1cc7ebd1c77fc59ace70cf94c4ba3762b3af0434d1
after  sha256: 1dc5a7f061a751cfeea97e1cc7ebd1c77fc59ace70cf94c4ba3762b3af0434d1
```

The verbose *log* differs in exactly three places, all environment noise: a set's
iteration order in one debug line, a memory limit derived from free RAM at the
time (`1.3GB` vs `1.2GB`), and the output filename. The written file is
byte-identical.

**Plugin entry point still works.** `@with_plugins(entry_points(group="gpio.plugins"))`
is untouched, and coexists with `cli.add_command`. Verified by putting a fake
`gpio-probe32a` dist-info with a `gpio.plugins` entry point on `PYTHONPATH`:

```
$ python -c "from geoparquet_io.cli.main import cli; print(sorted(cli.commands))"
['add', 'benchmark', 'check', 'convert', 'extract', 'inspect', 'partition',
 'pmtiles', 'probe32a', 'process', 'publish', 'skills', 'sort']
```

`probe32a` is registered and renders in `gpio --help`.

**Line counts**

| file | before | after |
|---|---|---|
| `geoparquet_io/cli/main.py` | 7,730 | 7,349 |
| `geoparquet_io/cli/_shared.py` | — | 81 |
| `geoparquet_io/cli/commands/__init__.py` | — | 15 |
| `geoparquet_io/cli/commands/sort.py` | — | 359 |

**Suites** (run at both commits; commit 1 passes on its own)

```
uv run pytest -n auto -m "not slow and not network and not meta" -q
  6266 passed, 75 skipped, 9 xfailed in 183.70s
uv run pytest -m meta -q
  203 passed, 7080 deselected
```

The meta lane is deselected by the fast suite and does not run on PRs, so it was
run explicitly — it contains repo-structure checks a file move can break.

**pre-commit, both stages**

```
uv run pre-commit run --all-files              # all Passed
uv run pre-commit run --all-files --hook-stage pre-push
  deptry ......... Passed
  mypy ........... Passed
  vulture ........ Passed
  xenon .......... Passed
  import-linter .. Passed
  menard-check ... Failed  (pre-existing, see below)
  fast-tests ..... Passed
```

`import-linter` specifically:

```
$ uv run lint-imports
Core modules must not import Click KEPT (1 ignored import)
API modules must not import CLI KEPT
Contracts: 2 kept, 0 broken.
```

`menard-check` fails, and **failed identically on `origin/main` before this branch
existed** — it flagged the same three doc targets (`CLAUDE.md#CLI Command Groups`,
`geoparquet_io/skills/geoparquet.md#Command Reference`, `docs/contributing.md`)
attributed to `90038a9` and `f4a4b1d`, both already on main. `doc-sync` reports
nothing to regenerate, because none of the generated content changed. It is a
pre-push-only hook and is not run in CI.

Refs: none — this is Deep Review item 3.2, for which no GitHub issue exists.

## Summary by CodeRabbit

* **New Features**
  * Added the `gpio sort` command group with Hilbert, STR, column, and quadkey sorting subcommands.
  * Added sorting options for compression, row groups, GeoParquet versions, overwriting, memory limits, and S3 configuration.

* **Refactor**
  * Organized CLI commands into dedicated modules while preserving sorting behavior and help handling.

* **Tests**
  * Expanded CLI parity and option-forwarding coverage for the sorting commands.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -233,7 +233,7 @@ repos:
           - -c
           - |
             # This is a reminder hook - it doesn't fail, just warns
-            if git diff --cached --name-only | grep -q "geoparquet_io/cli/main.py"; then
+            if git diff --cached --name-only | grep -qE "geoparquet_io/cli/(main\.py|commands/)"; then
               echo "⚠️  Reminder: If you added new CLI commands, ensure you also:"
               echo "   1. Added corresponding methods to api/table.py"
               echo "   2. Added corresponding functions to api/ops.py"
@@ -249,7 +249,7 @@ repos:
         description: Validate CLAUDE.md against the actual codebase
         entry: uv run python scripts/validate_claude_md.py
         language: system
-        files: ^(CLAUDE\.md|geoparquet_io/cli/main\.py|pyproject\.toml)$
+        files: ^(CLAUDE\.md|geoparquet_io/cli/main\.py|geoparquet_io/cli/commands/.*\.py|pyproject\.toml)$
         pass_filenames: false
 
       - id: doc-sync
@@ -257,7 +257,7 @@ repos:
         description: Auto-update generated doc sections (CLAUDE.md, SKILL.md, contributing.md)
         entry: uv run python scripts/doc_sync.py --update
         language: system
-        files: ^(CHANGELOG\.md|geoparquet_io/cli/main\.py|geoparquet_io/cli/decorators\.py|pyproject\.toml|geoparquet_io/core/.*\.py|scripts/doc_sync\.py|docs/contributing\.md)$
+        files: ^(CHANGELOG\.md|geoparquet_io/cli/main\.py|geoparquet_io/cli/commands/.*\.py|geoparquet_io/cli/decorators\.py|pyproject\.toml|geoparquet_io/core/.*\.py|scripts/doc_sync\.py|docs/contributing\.md)$
         pass_filenames: false
 
       - id: sync-dependencies
--- a/geoparquet_io/cli/_shared.py
+++ b/geoparquet_io/cli/_shared.py
@@ -1,0 +1,81 @@
+"""CLI helpers shared between ``cli/main.py`` and the per-group command modules.
+
+``cli/main.py`` owns the root ``gpio`` group and imports each command group from
+:mod:`geoparquet_io.cli.commands`. Anything a group module also needs therefore
+cannot keep living in ``main.py``: importing it back would make the dependency
+cyclic. It lives here instead, so both sides import downwards.
+
+The split against the neighbouring modules:
+
+* :mod:`geoparquet_io.cli.decorators` - reusable Click *option* decorators and
+  the ``click.Command`` subclasses commands declare with ``cls=``
+  (``GlobAwareCommand``, ``SingleFileCommand``, ``parse_row_group_options``).
+  That is still their home; nothing was moved out of it.
+* this module - the group-neutral runtime plumbing that is not a decorator and
+  not a Click option: S3 activation, and the default-subcommand group factory.
+
+Only helpers used by more than one command group belong here. A helper used by a
+single group travels with that group into ``cli/commands/<group>.py``.
+"""
+
+import os
+from contextlib import contextmanager
+
+import click
+
+
+@contextmanager
+def _activate_s3(ctx, aws_profile=None, s3_endpoint=None, s3_region=None, s3_no_ssl=False):
+    """Resolve and activate S3 config from ctx + per-command overrides.
+
+    Properly saves/restores AWS_PROFILE env var to avoid credential leaks.
+    """
+    from geoparquet_io.core.duckdb_utils import s3_config_scope
+    from geoparquet_io.core.remote import resolve_s3_config
+
+    config = resolve_s3_config(
+        s3_endpoint=s3_endpoint or ctx.obj.get("s3_endpoint"),
+        s3_region=s3_region or ctx.obj.get("s3_region"),
+        s3_no_ssl=s3_no_ssl or ctx.obj.get("s3_no_ssl", False),
+        aws_profile=aws_profile or ctx.obj.get("aws_profile"),
+    )
+    previous_profile = os.environ.get("AWS_PROFILE")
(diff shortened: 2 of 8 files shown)


cholmes:
Reviewed. The invariant holds and the pattern is right — one correction to the PR body.

**Verified independently:**

- `tests/data/cli_surface.json` is absent from the diff entirely; not re-recorded. The only change to `tests/test_cli_surface.py` is a one-word docstring correction (`cli/main.py` → `cli/_shared.py`) — no assertion touched.
- `uv run pytest -m meta` — 203 passed. Worth stating because that lane is deselected by the fast suite *and* does not run on pull requests, so nothing in this PR's checks covers it; a file move is exactly the shape of change that breaks its repo-structure tests.
- `main.py` 7,730 → 7,349; `_shared.py` 81, `commands/sort.py` 359, `commands/__init__.py` 15.

**Correction — claim 4 in the report is wrong.** `menard-check` does **not** fail on `main`, and does not fail on this branch either:

```
$ git rev-parse --short HEAD        # 72d3157, current main
$ uv run pre-commit run menard-check --all-files --hook-stage pre-push
menard-check.............................................................Passed

$ # same, on this branch
menard-check.............................................................Passed
```

The report cites commits `90038a9`/`f4a4b1d`, which predate the v1.5.0 release merge — `main` moved underneath while the work was in progress. There is no pre-existing failure to document, so that paragraph should come out of the PR body rather than leaving a reader hunting for a phantom.

**The two incidental findings are the valuable part of this PR**, and both are things the next ten moves would have hit blind:

1. **Three pre-commit hooks matched only `cli/main.py`** (`doc-sync`, `validate-claude-md`, `check-api-for-cli`), so a change confined to a group module would have skipped them silently. Widening those patterns is arguably more important than the move itself — without it, the split would have progressively disabled its own doc-sync enforcement, one group at a time.
2. **Two test modules patch a command's `*_impl` alias in the module that resolved it**, so a group move breaks them (14 failures here). Now parameterised by owning module, making each future move a one-line edit in each file. Every group PR will hit this; it belongs in the recipe.

`OptionalIntCommand` confirmed dead — one definition, zero references repo-wide, and vulture does not flag it. Agreed it is out of scope here; it should go when `inspect` moves.

*Reviewed by Claude (AI).*

PR #945, opened 2026-09-09
fix(crs,reproject): recover the real geometry column from malformed 'geo' metadata

### LINKED ISSUE #887: fix(crs,convert): remaining malformed-geo-block readers crash outside #883's four fixed sites
Follow-up from the adversarial review of #883 (findings from an AI review agent, all crash-verified on real files). #883 fixes the four write-path readers via `sanitize_geo_metadata`; these readers still index the raw block unguarded (pre-existing on main, cross-cutting → separate PR per repo convention):

1. `crs_utils.extract_crs_from_parquet` (crs_utils.py:620-628): `gpio convert geoparquet` fails with `'list' object has no attribute 'get'` (block=list), `argument of type 'NoneType' is not iterable` (columns=null), `'str' object has no attribute 'get'` (entry=string). Also reached by `convert reproject` (reproject.py:248), `format_writers.py:123`, `process/aggregate` (by_admin.py:273, grid_common.py:584).
2. `crs_utils.extract_crs_from_table` (crs_utils.py:~575 — distinct from streaming's): AttributeError on block-list and entry-string; used by aggregate grid keying (grid_common.py:692).
3. `primary_column: 123` (non-string) → `quote_identifier` TypeError in convert/sort; `primary_column` naming a nonexistent column → raw DuckDB BinderException traceback in sort/add.
4. Kept entry with `encoding: 123` → `convert.py:276 _calculate_bounds` AttributeError.

Routing these through `sanitize_geo_metadata` (or a thin shared guard) plus a clean error for the bad-primary_column cases closes the remaining #771 taxonomy. Note: DuckDB-read commands (sort/extract/add) also die with DuckDB's own "Geoparquet metadata is not an object" for block-level malformations — beyond gpio's reach, worth mentioning in #771's resolution.

Refs #771, #883.

cholmes:
## What changed

Five more readers of a carried `geo` block now go through the shared
`sanitize_geo_metadata` check that #883 established:

| Site | Category | Before |
|---|---|---|
| `crs_utils.extract_crs_from_parquet` | write-path → sanitize | `argument of type 'NoneType' is not iterable` (`columns: null`), `list indices must be integers` (`columns` an array), `'str' object has no attribute 'get'` (entry not an object), `'list'/'str' object has no attribute 'get'` (block not an object) |
| `crs_utils.extract_crs_from_table` (the aggregate grid-keying one, distinct from `streaming`'s) | write-path → sanitize | `AttributeError` on a non-object block and on an entry that is not an object |
| `crs_utils.geoparquet_crs_is_null` | write-path → sanitize | `AttributeError` on every non-object `columns` and every non-object block |
| `crs_utils.crs_string_from_geo_meta` (so `crs_string_from_table` and `stream_io` too) | write-path → sanitize | same |
| `convert.detect_all_geometry_columns` | write-path → sanitize | `'NoneType'/'list'/'str' object has no attribute 'items'`; and it passed a non-string `primary_column` to `quote_identifier` and a non-string `encoding` to `_calculate_bounds`'s `encoding.lower()` |
| `geometry_detection.detect_parquet_geometry_column` / `find_primary_geometry_column` | **validation-shared → guard, not sanitize** | returned `primary_column: 123` verbatim, which failed in `quote_identifier` |

The read-only readers keep the line #883 drew. `get_geo_metadata` and
`parse_geo_metadata_from_schema` still hand the block back exactly as the file
holds it, because `gpio check` reads through them. `geometry_detection` is the
one site that is shared between the two worlds — `check spatial` and
`check row-group` ask it what the file calls its geometry — so it does not
sanitize either; it only refuses to hand back the one thing that cannot be a
column name, a non-string `primary_column`, and falls through to schema
detection instead. The pre-1.0 list-shaped block gets the same guard.

`convert.detect_all_geometry_columns` also falls back to schema detection when
nothing usable is left of `columns`, which is the answer a block-less file
already gives.

## Why

The `geo` key on an input file is arbitrary JSON written by somebody else's
tool. #883 fixed the four readers on the write path proper; these run *earlier*
— `crs_utils` resolves the input CRS and `convert` resolves the geometry columns
before any output metadata exists — and still indexed the raw block. Every
crash above was reproduced on a real file before the fix and is pre-existing on
`main`; per repo convention the cross-cutting fix gets its own PR.

One honest limit, already recorded in #887: a file whose `columns` or `encoding`
is the wrong type still cannot be converted, because DuckDB's own GeoParquet
reader refuses to open it (`Invalid Input Error: Geoparquet metadata does not
have a columns object`). That is outside gpio's reach. What changed is that the
failure is now a domain error naming the real cause, alongside gpio's warning
saying which key was malformed — not a `TypeError` raised from indexing the
block three frames deep. `primary_column: 123`, whose file DuckDB *can* read,
now converts to completion.

Sweep of the remaining unguarded raw-block readers, reported not fixed (other
agents hold those files, and each is larger than a one-liner): `common.py`
(2774–2794, 3435), `add/bbox_metadata.py` (222–253, 338–344), `add/bbox.py`
(210–216), `reproject.py` (76–124, 289–295, 383, 825–831), `stac.py` (216–219),
`stream_io.py` (281–286), `api/table.py` (2558). The `primary_column`-names-a-
nonexistent-column case in `sort`/`add`, which surfaces a raw DuckDB
`BinderException`, is also left for a separate change.

## Verification

- `uv run pytest -n auto -m "not slow and not network and not meta" -q` — 6503 passed, 75 skipped, 9 xfailed
- `uv run pytest -m meta -q` — 203 passed
- `uv run pre-commit run --all-files` and `--hook-stage pre-push` — clean
- `diff-cover coverage.xml --compare-branch=origin/main --fail-under=90` — **100%** (15/15 changed lines)
- 30 new tests in `tests/test_malformed_geo_metadata.py`, each of the eight
  malformed shapes against each of the five readers, plus the end-to-end
  `convert` contract: never a `TypeError`/`AttributeError` anywhere in the
  cause chain

Closes #887

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed GeoParquet metadata across CRS extraction, geometry detection, conversion, and write operations.
  * Falls back to schema-based geometry detection when metadata is missing or invalid.
  * Prevents crashes caused by malformed metadata while preserving valid CRS and geometry information.
  * Provides clearer warnings and domain-specific errors for invalid metadata.

* **Tests**
  * Added regression coverage for malformed metadata scenarios across supported workflows.

### DIFF
--- a/geoparquet_io/core/convert.py
+++ b/geoparquet_io/core/convert.py
@@ -179,6 +179,15 @@ def _detect_geometry_column(con, input_file, verbose, is_parquet=False, layer=No
     return None
 
 
+def _schema_geometry_column(input_file: str, verbose: bool, is_parquet: bool = True) -> str | None:
+    """``_detect_geometry_column`` on a connection of its own, closed either way."""
+    con = get_duckdb_connection(load_spatial=True, load_httpfs=needs_httpfs(input_file))
+    try:
+        return _detect_geometry_column(con, input_file, verbose, is_parquet=is_parquet)
+    finally:
+        con.close()
+
+
 def detect_all_geometry_columns(input_file: str, verbose: bool = False) -> dict:
     """Detect all geometry columns from a GeoParquet file.
 
@@ -206,14 +215,10 @@ def detect_all_geometry_columns(input_file: str, verbose: bool = False) -> dict:
     # Only GeoParquet files can have multiple geometry columns with metadata
     if not _is_parquet_file(input_file):
         # For non-parquet, detect single geometry column the standard way
-        con = get_duckdb_connection(load_spatial=True, load_httpfs=needs_httpfs(input_file))
-        try:
-            geom_col = _detect_geometry_column(con, input_file, verbose, is_parquet=False)
-            if geom_col:
-                result["primary"] = geom_col
-                result["metadata"][geom_col] = {"encoding": "WKB"}
-        finally:
-            con.close()
+        geom_col = _schema_geometry_column(input_file, verbose, is_parquet=False)
+        if geom_col:
+            result["primary"] = geom_col
+            result["metadata"][geom_col] = {"encoding": "WKB"}
         return result
 
     # A write-path reader: the column names and encodings this returns are
@@ -227,18 +232,19 @@ def detect_all_geometry_columns(input_file: str, verbose: bool = False) -> dict:
     if not geo_meta or not geo_meta.get("columns"):
         # No GeoParquet metadata -- or nothing left of it once the malformed
         # parts were dropped. Either way, detect the column from the schema.
-        con = get_duckdb_connection(load_spatial=True, load_httpfs=needs_httpfs(input_file))
-        try:
-            geom_col = _detect_geometry_column(con, input_file, verbose, is_parquet=True)
-            if geom_col:
-                result["primary"] = geom_col
-                result["metadata"][geom_col] = {"encoding": "WKB"}
-        finally:
-            con.close()
+        geom_col = _schema_geometry_column(input_file, verbose)
+        if geom_col:
+            result["primary"] = geom_col
+            result["metadata"][geom_col] = {"encoding": "WKB"}
         return result
 
     # Extract from GeoParquet metadata
-    primary_col = geo_meta.get("primary_column", "geometry")
+    primary_col = geo_meta.get("primary_column")
+    if not isinstance(primary_col, str):
+        # Sanitizing dropped a malformed `primary_column` and could not repair
+        # it from a single column. The literal "geometry" would name a column
+        # this file may not have, so ask the schema (#887 review).
+        primary_col = _schema_geometry_column(input_file, verbose)
     columns = geo_meta.get("columns", {})
 
     result["primary"] = primary_col
--- a/geoparquet_io/core/convert.py
+++ b/geoparquet_io/core/convert.py
@@ -199,6 +199,7 @@ def detect_all_geometry_columns(input_file: str, verbose: bool = False) -> dict:
             - "metadata": dict - per-column metadata from input (crs, encoding, etc.)
     """
     from geoparquet_io.core.duckdb_metadata import get_geo_metadata
+    from geoparquet_io.core.geo_metadata import sanitize_geo_metadata
 
     result = {"primary": None, "secondary": [], "metadata": {}}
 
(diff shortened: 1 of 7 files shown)


cholmes:
## Multi-agent adversarial review — two S2s found and fixed in 9feb0b0

The core work was sound and strictly better than main on every shape tested, but the review found two blockers. I verified both myself against a worktree at the pre-fix commit rather than taking the fixer's word for it — the before/after numbers below are my own runs.

### S2 — the recovery turned a loud crash into silently wrong coordinates

This is the one that mattered. When `sanitize_geo_metadata` dropped a malformed `primary_column`, four readers fell back to the **literal string `"geometry"`**. Every test in the PR used a column literally named `geometry`, so the suite was blind to it — but for a file whose column is `geom` (or `wkb_geometry`, or `shape`, all in gpio's own detection list) the lookup misses and the CRS is reported as absent. Main crashed; this would have shipped quiet corruption.

**(a) The explicit-null-CRS guard was bypassed.** `crs: null` on column `geom`, `primary_column: 123`, `gpio convert reproject --dst-crs EPSG:3857`:

| | before | after |
|---|---|---|
| control (`primary_column: "geom"`) | `rc=1` — `Error: Input has an explicit null CRS (unknown)… --assume-crs84` | same |
| malformed | **`rc=0`, "Reprojected 1 features / Source CRS: EPSG:4326"** | `rc=1`, identical to control |

The file declared its CRS *unknown*; gpio invented `EPSG:4326`, reprojected on that basis, and exited 0.

**(b) A projected file was grid-keyed as lon/lat.** EPSG:5070 point on column `geom`, `gpio process aggregate h3 --resolution 5`:

| | before | after |
|---|---|---|
| control | `85268287fffffff` | `85268287fffffff` |
| malformed (`primary_column: 7`) | **`85b6593bfffffff`** | `85268287fffffff` ✓ |

**Fix:** `sanitize_geo_metadata` now *repairs* rather than only drops — when a malformed `primary_column` is dropped and the sanitized `columns` has exactly one entry, that key becomes the primary, which is unambiguous per the spec. Where it cannot repair, the four readers ask the file's schema instead of guessing a name. I probed the repair boundaries directly: 1 column → repaired; **2 columns → refuses to guess** and falls through to schema detection; 0 columns → `None`; a well-formed primary → untouched. Conservative in the right direction.

### S2 — a crash listed in #887 still reproduced through the public API

`reproject.py` read `primary_column` off the raw block with no string guard, so `_detect_geometry_column_from_table` returned the raw integer, which reached `quote_identifier` and died with `TypeError: argument of type 'int' is not iterable` — while the PR claimed `Closes #887`. Verified before and after:

```
before   _detect_geometry_column_from_table(table) -> 7
after    _detect_geometry_column_from_table(table) -> 'geom'
```

The fix also caught **two more sites of the same class the review had not named**: both metadata-update paths (`reproject_table`, `_reproject_streaming`) decoded the raw block and handed it to a function that indexes `columns`, where a list-shaped block raises an `AttributeError` their `except (JSONDecodeError, KeyError)` does not catch.

### S3 — the guard dropped silently

The three `isinstance(primary, str)` guards now go through one `carried_column_name` helper that warns naming the ignored key, matching the rule #883 established. Confirmed live: the malformed file now emits `Ignoring malformed 'geo' metadata on the input: 'primary_column' is a number, expected a string`.

### One correction to the fix report

The claim that two malformed files in one process now each get a warning holds only for the `geometry_detection` path. `sanitize_geo_metadata` takes no `source` parameter, so readers that sanitize rather than guard still emit an un-sourced message and still dedupe globally — I reproduced one warning for two distinct malformed files. Pre-existing from #883 and diagnostic rather than correctness, so filed as **#957** rather than folded in here.

### A test that encoded the bug

`test_detect_geometry_column_from_table_handles_invalid_json` asserted that a table whose only column is `geom` resolves to `"geometry"` — a column that is not in the schema. It now asserts `"geom"`, with a docstring saying why. Worth flagging explicitly since changing an existing assertion is normally how a regression gets hidden; here the old assertion *was* the defect.

The readers' parametrization is now over `GEOMETRY_COLUMN_NAMES = ["geometry", "geom"]`, and that single change alone catches the S2. 42 new/changed tests failed before the fix and pass after.

### Gates (my runs, not the fixer's)

Fast suite **6611 passed, 75 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100% on 108 changed lines, 0 missing** · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push hooks clean.

PR #948, opened 2026-09-09
refactor(cli): extract the duplicated command prologue, group callback and kdtree option resolution

### LINKED ISSUE #919: refactor(cli): extract the duplicated command prologue and option stacks the module split exposed
Follow-ups from the multi-agent review of #916 (AI review agents; every count below was measured, not estimated). None were fixed in #916 because they would change behavior or destroy the byte-exactness that made the move cheap to verify — but the split is what made them legible, which is a point in its favor.

Ordered by impact:

**1. The same 13-line output-validation prologue appears verbatim 7 times (91 lines).**
```python
from geoparquet_io.core.streaming import StreamingError, validate_output
try:
    validate_output(output_parquet)
except StreamingError as e:
    raise click.ClickException(str(e)) from None
validate_parquet_extension(output_parquet, any_extension)
row_group_mb = parse_row_group_options(row_group_size, row_group_size_mb)
```
At `add.py:443,559,642,725,965`, `extract.py:223`, `sort.py:92`. The 6-line `validate_output` guard alone appears 9 times (adds `extract.py:680`, `sort.py:174`). Related counters: `parse_row_group_options(...)` at 26 call sites, `validate_parquet_extension(...)` at 19. Proposal: a `_shared.prepare_output(path, any_extension, row_group_size, row_group_size_mb) -> float | None` doing all three steps. 91 lines → 7.

Important nuance from a second reviewer: only the **guard-side** `except StreamingError` handlers (9, wrapping `validate_output`) should be absorbed. The **call-side** ones (7, wrapping the core `*_impl` call) are healthy explicitness — each wraps a different core function and the `from None` is deliberate traceback suppression a shared decorator would hide. Also noted: `add` wraps both guard and impl call, while `sort hilbert`/`sort str` wrap only the guard, so a `StreamingError` escaping `hilbert_impl` surfaces as a traceback rather than a ClickException — worth aligning.

**2. The same 7-decorator option stack appears 7 times in `add.py` alone (49 lines).**
`@output_format_options @geoparquet_version_option @overwrite_option @dry_run_option @verbose_option @any_extension_option @show_sql_option` at `add.py:86,291,380,521,604,687,926`; near-identical variants at `partition.py:316,487,660,1006` and `partition.py:74,195,829`; `sort.py:53,135` and `sort.py:209,296`. ~120 decorator lines total. `decorators.py` already composes at this altitude (`output_format_options` bundles compression + row-group + write-memory), so the pattern exists — it just stops one level short. A `standard_write_options` composite takes each stack from 7 lines to 1.

**3. Identical group-callback body, 8 times.**
`ctx.ensure_object(dict)` + `timestamps = ctx.obj.get(...)` + `setup_cli_logging(...)` at `partition.py:44`, `inspect.py:76`, `sort.py:36`, `add.py:39`, `convert.py:123`, `publish.py:200`, `extract.py:62`, `main.py:151`. Eight modules import `setup_cli_logging` solely for this. One `_shared.init_group_context(ctx)` removes the import from seven of them.

**4. `add kdtree` and `partition kdtree` duplicate ~20 lines of option validation, now across files.**
`add.py:847-870` vs `partition.py:889-914` — same three checks, same messages, diverging only in the tail. The split made this cross-file, so drift is now *likelier*. Both also use a sentinel comparison for exclusivity (`if exact and approx != 100000`), so `--exact --approx 100000` passes silently; the codebase already knows the right idiom at `add.py:462-464` (`ctx.get_parameter_source(...) is not ParameterSource.DEFAULT`).

**5. `bbox_option` in `decorators.py:332` has zero production callers** while three commands declare its byte-identical literal inline (`add.py:62-64`, `sort.py:50-52`, `sort.py:132-134`). Its only reference is `tests/test_decorators.py:23`. Adopt it in the three places or delete it and its test.

**6. Triplicated bbox parsing in `extract.py`** (`:464-473`, `:986-998`, `:1203-…`) is also the sole cause of all 5 `arg-type` mypy errors there, because `tuple(parts)` is `tuple[float, ...]` and the length-4 invariant is invisible to the checker. A `_parse_bbox_option(...) -> tuple[float, float, float, float] | None` helper that indexes the four elements fixes both — no `cast`, no ignore, and the extract override in `pyproject.toml` disappears entirely. Note the three copies have slightly different error messages today ("Expected 4 values" vs "bbox must have exactly 4 values"), so unifying is user-visible; `tests/test_extract_cli_layer.py` (added in #916) now pins the current strings, which is the safety net this needs.

Refs #916.

cholmes:
## What changed

Four extractions, one commit each, in `cli/commands/*.py`, `cli/_shared.py` and
`cli/decorators.py`. Together they remove **257 lines from the command modules**
and add 143 to the two shared modules.

**1. `_shared.prepare_output(...)` — the output-validation prologue** (`-133/+30`)

The same three checks — guard the output against a missing destination, reject a
non-`.parquet` extension, parse the mutually exclusive row-group options — were
spelled out in full at the top of **ten** call sites across nine commands: `add bbox`, `add h3`,
`add a5`, `add s2`, `add quadkey`, `sort hilbert`, `sort str`,
`extract geoparquet`, `extract bigquery` and `convert` (10 sites; `convert`
folded its import in with `should_stream_output`). One helper now does all three
and returns the row-group size in MB. **No `validate_output` call remains in
`cli/`.**

Only the **guard-side** `except StreamingError` is absorbed. The five
**call-side** handlers in `add.py` — the ones wrapping the core `*_impl` call —
stay put: each wraps a different core function and the `from None` there is
deliberate traceback suppression a shared helper would hide.

**2. `_shared.init_group_context(ctx)` — the group callback** (`-45/+24`)

`add`, `check`, `convert`, `extract`, `inspect`, `partition`, `publish` and
`sort` all opened with the same three statements, and eight modules imported
`setup_cli_logging` for nothing else. The root `gpio` callback in `main.py` keeps
its own body — it *writes* `ctx.obj["timestamps"]` rather than reading it.

**3. `bbox_option` adopted for `--add-bbox`** (`-9/+5`)

The decorator had zero production callers while three commands declared its
byte-identical literal inline. Adopted rather than deleted.

**4. `decorators.resolve_kdtree_options(...)`** (`-70/+8`)

`add kdtree` and `partition kdtree` each turned the same
`--partitions`/`--auto`/`--approx`/`--exact` quartet into the same
`(iterations, sample_size, auto_target)` triple with their own ~25-line copy,
including three `UsageError` messages that had to stay worded identically with
nothing enforcing it. The two copies were already written differently, and #916
moved them into separate files, so the next divergence would have been invisible.

Behaviour is preserved exactly, including the quirk that `--exact` is checked
against `--approx`'s *default value* rather than against whether the user typed
it, so `--exact --approx 100000` is still accepted silently. That is now written
down in one place with a pointer to this issue instead of being an unremarked
line in two files — fixing it needs a `ParameterSource` check and is a
user-visible change.

41 new unit tests cover the three new helpers (20 kdtree, 11 group-context, 10 prepare-output).

## Why

The #916 module split did not create any of this duplication; it made it
legible. The counts in #919 were re-measured against `main` before starting —
line numbers had shifted, and the prologue turned out to appear at **ten** sites
rather than seven, with no guard-only residue left over.

### Judged not worth extracting

**Issue item 2, a `standard_write_options` composite decorator** (~120 lines).
Declined. There is no single stack to name: `add` uses a 7-decorator stack (7
sites), `partition` a differently-ordered 4 (7 sites), `sort` a 6 with
`default_rows=` (4 sites), and the one-offs in `convert`/`extract` match none of
them. Naming each group's current accident gives three composites that differ by
one or two options, and the first time a group gains one the composite grows a
boolean parameter — worse than the duplication. It would also only halve the
copy-paste it targets: every option still has to be named in the function
signature. `output_format_options` is the right precedent for a *cohesive* bundle
(five knobs that all describe the output file's physical layout); "everything
`gpio add` happens to accept" is not one.

**A second helper for the extension + row-group pair** (6 adjacent sites).
Declined. Those two statements name exactly what they do; replacing them with one
four-argument call that wraps to three lines is longer *and* less clear. Adding a
`require_output=False` flag to `prepare_output` is worse still: none of these
sites calls the streaming guard today, so adopting the helper would *add*
behaviour rather than preserve it, and passing `-` would newly emit a
"Writing binary Arrow IPC data to terminal" warning.

(An earlier draft of this paragraph said OUTPUT is a required argument at all of
these commands. That is true for some — `add h3`'s OUTPUT is required — but not
for `add admin-divisions`, `add geometry-metrics`, `convert reproject` or
`extract wfs`, which all declare `required=False, default=None`. Each exclusion
still stands; the reason is the one above, that the guard was never there to
begin with, not that it could not apply.)

**Issue item 6, the triplicated bbox parsing in `extract.py`.** Declined, with a
measurement worth recording: the three copies raise **different** inner
`ValueError` messages ("bbox must have exactly 4 values" vs "Expected 4 values"),
and `arcgis` interpolates its one into the user-facing text, so unifying them
changes a user-visible string. More importantly, the issue's claim that this
makes the `pyproject.toml` mypy override "disappear entirely" does not hold: with
the override lifted, `extract.py` emits 5 `arg-type` errors, and only **4** are
the bbox tuple. The fifth is
`row_group_size_mb: float | None` vs `int | None` on
`convert_arcgis_to_geoparquet` — a `core/arcgis.py` annotation, so the override
survives either way. This belongs with a `core/` annotation pass.

**Aligning `sort hilbert`/`sort str` to wrap their impl call in
`except StreamingError`**, as a second reviewer suggested in the issue. That is a
behaviour change (a traceback becomes a `ClickException`), not a refactor.

## Verification

- `git diff --stat origin/main -- tests/data/cli_surface.json` is **empty** — the
  snapshot is byte-identical, and `tests/test_cli_surface.py` was green after
  every one of the four commits. The snapshot was never re-recorded.
  → **6489 passed, 75 skipped, 9 xfailed** in 317s. Two subprocess-pipe tests
  (`test_streaming_handles_broken_pipe`, `test_zero_row_pipe_to_add_bbox`) failed
  on the saturated parallel run and pass both serially and on a re-run under
  `-n auto` — machine saturation, not the change.
- Meta lane: `uv run pytest -m meta -q` → **203 passed**.
- `uv run pre-commit run --all-files` clean; `--hook-stage pre-push` clean
  (deptry, mypy, vulture, xenon, import-linter, menard-check, fast-tests).

Closes #919

### DIFF
--- a/geoparquet_io/cli/_shared.py
+++ b/geoparquet_io/cli/_shared.py
@@ -103,9 +103,13 @@ def prepare_output(
        that ``--row-group-size`` and ``--row-group-size-mb`` are mutually
        exclusive and converts the size string to MB.
 
-    The order matters and is part of what this helper pins: a user who typed
-    neither an output nor a ``.parquet`` name should get the streaming hint,
-    which tells them what to do, rather than the extension complaint.
+    The order matters and is part of what this helper pins, though not between
+    steps 1 and 2: those two can never both fire, since ``validate_output``
+    raises only for a missing output and ``validate_parquet_extension`` returns
+    immediately on one. The live constraint is step 1 before step 3 -- a user
+    who piped nothing anywhere and also passed both row-group options should be
+    told how to name an output, which is actionable, rather than which pair of
+    size flags conflict.
 
     Args:
         output_path: The command's output argument. ``None`` means "stream to
--- a/geoparquet_io/cli/_shared.py
+++ b/geoparquet_io/cli/_shared.py
@@ -25,6 +25,25 @@ from contextlib import contextmanager
 
 import click
 
+from geoparquet_io.core.logging_config import setup_cli_logging
+
+
+def init_group_context(ctx: click.Context) -> None:
+    """Prime ``ctx.obj`` and set up CLI logging for a ``gpio`` subgroup.
+
+    Every group callback needs both, for the same two reasons:
+
+    * ``ctx.obj`` is filled by the root ``gpio`` callback, but a group object
+      invoked on its own -- ``CliRunner().invoke(add, [...])``, or any
+      programmatic caller -- never runs it, and reading ``ctx.obj`` then fails
+      on ``None`` (#922).
+    * ``--timestamps`` is a root option, so a group has to read it back out of
+      ``ctx.obj`` and hand it to the logger. ``verbose`` stays ``False`` here;
+      individual commands raise it to DEBUG themselves.
+    """
+    ctx.ensure_object(dict)
+    setup_cli_logging(verbose=False, show_timestamps=ctx.obj.get("timestamps", False))
+
 
 @contextmanager
 def _activate_s3(ctx, aws_profile=None, s3_endpoint=None, s3_region=None, s3_no_ssl=False):
--- a/geoparquet_io/cli/_shared.py
+++ b/geoparquet_io/cli/_shared.py
@@ -13,7 +13,8 @@ The split against the neighbouring modules:
   helpers such as ``parse_row_group_options``. That is still their home;
   nothing was moved out of it.
 * this module - the group-neutral runtime plumbing that is not a decorator and
-  not a Click option: S3 activation, and the default-subcommand group factory.
+  not a Click option: S3 activation, the shared write-command prologue, and the
+  default-subcommand group factory.
 
 Only helpers used by more than one command group belong here. A helper used by a
 single group travels with that group into ``cli/commands/<group>.py``.
@@ -60,6 +61,67 @@ def _activate_s3(ctx, aws_profile=None, s3_endpoint=None, s3_region=None, s3_no_
             os.environ.pop("AWS_PROFILE", None)
         else:
             os.environ["AWS_PROFILE"] = previous_profile
+
+
+def prepare_output(
+    output_path: str | None,
+    any_extension: bool,
+    row_group_size: int | None,
+    row_group_size_mb: str | None,
+) -> float | None:
+    """Run the three checks every write command performs before it does any work.
+
+    In order:
+
+    1. :func:`~geoparquet_io.core.streaming.validate_output` - refuses a missing
+       output when stdout is a terminal, and warns when binary Arrow IPC would
+       be written to one. Its ``StreamingError`` is re-raised as a
+       ``ClickException`` with ``from None``: the message is already written for
+       a user, and the traceback behind it is noise.
+    2. :func:`~geoparquet_io.core.file_utils.validate_parquet_extension` -
+       rejects a non-``.parquet`` output unless ``--any-extension`` was given.
+    3. :func:`~geoparquet_io.cli.decorators.parse_row_group_options` - enforces
+       that ``--row-group-size`` and ``--row-group-size-mb`` are mutually
+       exclusive and converts the size string to MB.
+
+    The order matters and is part of what this helper pins: a user who typed
+    neither an output nor a ``.parquet`` name should get the streaming hint,
+    which tells them what to do, rather than the extension complaint.
+
+    Args:
+        output_path: The command's output argument. ``None`` means "stream to
+            stdout" and ``"-"`` means it explicitly.
+        any_extension: Value of ``--any-extension``.
+        row_group_size: Value of ``--row-group-size`` (rows).
+        row_group_size_mb: Value of ``--row-group-size-mb`` (size string).
+
+    Returns:
+        The row group size in MB, or ``None`` when it was not requested -- pass
+        it to the core writer as ``row_group_size_mb``.
+
+    Raises:
+        click.ClickException: No output was given and stdout is a terminal.
+        click.UsageError: Both row-group options were given, or the size string
+            is invalid.
+        InvalidParameterError: The output has a non-``.parquet`` extension and
+            ``--any-extension`` was not given.
+    """
+    from geoparquet_io.cli.decorators import parse_row_group_options
+    from geoparquet_io.core.file_utils import validate_parquet_extension
+    from geoparquet_io.core.streaming import StreamingError, validate_output
+
+    try:
+        validate_output(output_path)
+    except StreamingError as e:
+        raise click.ClickException(str(e)) from None
+
+    # ``validate_parquet_extension`` returns immediately on ``None``, but is
+    # annotated ``str``. Skipping the call keeps the None-handling visible here
+    # rather than relying on an annotation that does not admit it.
+    if output_path is not None:
+        validate_parquet_extension(output_path, any_extension)
+
+    return parse_row_group_options(row_group_size, row_group_size_mb)
 
 
 def create_default_group(default_subcommand: str, description: str) -> type:
--- a/geoparquet_io/cli/commands/add.py
+++ b/geoparquet_io/cli/commands/add.py
@@ -10,6 +10,8 @@ from click.core import ParameterSource
 
 from geoparquet_io.cli._shared import _activate_s3, init_group_context, prepare_output
 from geoparquet_io.cli.decorators import (
+    DEFAULT_KDTREE_APPROX,
+    DEFAULT_KDTREE_TARGET_ROWS,
     SingleFileCommand,
     any_extension_option,
     bbox_option,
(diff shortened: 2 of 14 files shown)


cholmes:
## Multi-agent adversarial review — merge-ready

Two independent agent reviews (adversarial + simplicity). No S1 or S2. Both corrected premises in my review brief, which is worth recording:

- `prepare_output` takes 4 params and returns a scalar, not 5 returning a tuple; there are no opt-out flags at any of the 10 call sites.
- The kdtree change did **not** switch to `ctx.get_parameter_source`. It deliberately preserved the sentinel comparison and pinned the quirk with a test, because switching would newly reject `--exact --approx 100000` — a user-visible change that does not belong in a behaviour-preserving refactor. Filed as #951.

### Behaviour preservation — verified, not asserted

- **Help output: 72/72 pages byte-identical** (full command tree walked on the merge base and on HEAD).
- **CLI surface: zero drift.** `tests/data/cli_surface.json` unchanged; `test_cli_surface.py` 74 passed.
- **Error text, exit codes and resolved values identical** across the prologue error matrix at all 10 former sites (including the combined cases that pin precedence) and a 22-combination kdtree grid against both `add kdtree` and `partition kdtree`. Only diff between base and branch was one temp-file UUID.
- **Call-side `except StreamingError` handlers were not absorbed**: 16 on main → 6 on this branch, exactly the 10 guard-side sites collapsing. `validate_output` is now called in precisely one place, so #919's guard-side/call-side line is respected exactly.

### Ratio

+459/−254 sounds like growth; production code is **net −37 lines**. 53% of the additions are tests, 18% docstrings. Roughly 35 executable lines replaced 253 duplicated ones. The simplicity reviewer's verdict was "the extraction is genuinely simpler — keep it, do not revert or narrow."

### Fixed in 9e00fe6 before merge

**S3 — the `--approx` sentinel became a silent cross-file coupling.** `_KDTREE_DEFAULT_APPROX = 100000` lived in `decorators.py` while the option's default stayed a bare literal in `add.py` and `partition.py`, which do not import it. Three copies, nothing linking them, and the new tests hardcode `100_000` too — so if either declaration drifted, a bare `--exact` would start reporting a conflict with an `--approx` nobody typed, and no test would fail. Both constants are now public and referenced at the declaration sites *including in the help text*, so drift is impossible rather than merely tested for. `{DEFAULT_KDTREE_APPROX}` and `{DEFAULT_KDTREE_TARGET_ROWS:,}` render the same `100000` and `120,000`, so the surface JSON stays byte-identical.

**S4 — `prepare_output`'s docstring justified its step order with an interaction that cannot happen.** `validate_output` raises only when the output is missing; `validate_parquet_extension` returns immediately on a missing output. Steps 1 and 2 can never both fire. The real constraint — and what the test actually exercises — is step 1 before step 3, so the test is renamed from `test_output_guard_runs_before_the_extension_check` to `..._before_the_row_group_parse`.

**S4 — PR body corrected.** It claimed the 9 non-adopting sites all have a required OUTPUT argument. True for some, but `add admin-divisions`, `add geometry-metrics`, `convert reproject` and `extract wfs` are all `required=False, default=None`. Every exclusion still stands, for a better reason: none of those sites calls the streaming guard today, so adopting the helper would *add* behaviour. Test count corrected 44 → 41.

### Gates

`lint-imports` 4 kept / 0 broken · fast suite **6491 passed, 75 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100% on 72 changed lines, 0 missing** (the gate that blocked #915/#916/#929 — coverage did not migrate from many covered call sites into one uncovered helper) · `pytest -m meta` 203 passed · pre-commit and pre-push hooks clean.

Note CodeRabbit's green check on this PR is not a signal — it hit its OSS rate limit and posted no review.

Reviews performed by Claude Code agents; findings verified independently before acting on them.

cholmes:
### CI follow-up: a Python 3.10-only failure, now guarded

`test (ubuntu-latest, 3.10)` failed while the other 10 matrix jobs passed:

```
AttributeError: 'Group' object has no attribute '_shared'
```

`tests/cli/test_shared_group_context.py` used the dotted-string form
`mock.patch("geoparquet_io.cli._shared.setup_cli_logging")`. `geoparquet_io/__init__.py` does `from geoparquet_io.cli.main import cli`, which binds the name `cli` inside the `geoparquet_io` package to the Click Group and shadows the `geoparquet_io.cli` **subpackage** as an attribute of its own parent:

```python
>>> import geoparquet_io, geoparquet_io.cli
>>> type(geoparquet_io.cli)
<class 'click.core.Group'>
```

`mock.patch` resolves a dotted target by importing the longest prefix it can and walking the rest with `getattr`. 3.10's `_get_target` uses `_importer()`, which getattr-walks; 3.11+ uses `pkgutil.resolve_name`, which imports the submodule first. So the test is green on 3.11/3.12/3.13 and red on 3.10 alone — which reads as a flake and invites the wrong fix. Switched to `patch.object`, which takes the module object and has no name to walk.

**This is the third time the shadowing has cost this repo.** `test_write_memory_forwarding.py` and `test_check_cli_guards.py` were each fixed previously with only a warning comment left behind, and the CLI split into `cli/commands/` made the exposed surface much larger than when the comments were written. Comments do not stop a defect class, so it is now guarded mechanically: `tests/test_patch_target_shadowing.py` AST-walks every `tests/test_*.py` for `patch("geoparquet_io.cli...")` string targets and names the offending file and line. Sabotage-verified — reintroducing the old form fails it with `tests/cli/test_shared_group_context.py:30`.

It is `meta`-marked, so it costs the fast suite nothing and does not touch diff-cover. It also carries a second test asserting the shadowing still exists, so if `__init__.py` ever stops rebinding the name, the guard fails and can be deleted rather than lingering as cargo cult.

PR #949, opened 2026-09-09
fix(extract,sort,partition): keep a secondary geometry column readable after a row filter

### LINKED ISSUE #934: fix(extract,sort,partition): deleting geometry_types on row filters makes multi-geometry output unreadable
Found during the #927 review (AI review agent; every row below reproduced on a two-geometry-column fixture). #927 fixed this for `convert reproject`; these call sites still carry the defect and are untouched by that PR.

## The defect

When a command invalidates derived stats, it **deletes** the `geometry_types` key rather than emptying it. For the *primary* column that is harmless — every write path recomputes it. For a **secondary** geometry column nothing recomputes on any file-write path, so the key stays gone and the output is unreadable:

```
Invalid Input Error: Geoparquet column 'other_geom' does not have geometry types
```

gpio's own validator rejects it too — 7 failed checks, led by `geometry_types_list_other_geom -> column "other_geom" must have a "geometry_types" list`. **gpio writes a file gpio rejects.**

## Reproduced (post-#927)

| command | secondary `geometry_types` | DuckDB read |
|---|---|---|
| `gpio extract geoparquet --where "id = 1"` | MISSING | FAILS |
| `gpio extract geoparquet --bbox 0,0,4,4` | MISSING | FAILS |
| `gpio sort quadkey '<glob>' out.parquet` | MISSING | FAILS |
| `sort hilbert` / `add bbox` / `add h3` / `convert geoparquet` | `['Point']` | OK |

Also reachable via `gpio partition *` (partition staging) and `gpio extract --repair-geometry` — the latter is worse, because repair touches only the primary column, so invalidating the secondary is not even semantically warranted.

## The fix

Invalidation *is* correct for a row filter — the rows really do change. The mechanism is what's wrong. Set `geometry_types: []` (the spec's valid "not known" sentinel) instead of deleting the key, so the block stays structurally valid and readers can still open the file. #927's `invalidate_derived_stats_columns` plumbing is the natural place to make that distinction: callers that pass nothing keep invalidating every column, but invalidation should empty rather than delete.

Worth checking whether `bbox` deserves the same treatment (it is also carried, also unrecomputed for secondaries) — the #927 review confirmed a carried secondary `bbox` stays correct and self-consistent when left alone, so the answer may differ per key.

Refs #927, #890.

cholmes:
## What changed

`geometry_types` is REQUIRED by GeoParquet 1.1, and the only column any file-write path recomputes it for is the **primary** geometry column. Marking the carried stats stale by *deleting* the key therefore worked for the primary and quietly corrupted every other geometry column.

- `strip_derived_stats` grows a `recomputed_columns` argument naming the columns whose absent key is a *request* for a recompute. Every other stripped column keeps `geometry_types` as `[]` — the spec's "not known" — and loses only `bbox`, which is optional and whose absence already means the same thing.
- `write_parquet_with_metadata` names the one column it recomputes, so `sort` and `partition` are fixed at the shared seam rather than at each call site. `partition/staging.py` stops stripping by hand and hands the invalidation to the write path for the same reason.
- *Which* columns are stale at all is now decided per site. `_output_stats_are_stale` becomes `_stale_stat_columns`.
- `backfill_derived_stats` reads an empty `geometry_types` as the gap it is, so the Arrow paths that *can* compute every column's types still do.

### Scoped vs emptied, per site

| Site | Carried stats | Choice |
|---|---|---|
| `extract` row filter (`--bbox`/`--geometry`/`--where`/`--limit`), single file | OVER-cover — output rows are a subset | **Scoped** to the primary. Over-cover is what the spec allows (the type list must be exhaustive, not exact; a wider bbox never hides a row), so the secondary's real `["Polygon"]` and its bbox survive. |
| `extract --repair-geometry` | OVER-cover; repair rewrites the primary only | **Scoped** to the primary — the secondary was never touched, so invalidating it was not even semantically warranted. |
| `extract` over a glob/directory | UNDER-cover — only the FIRST file's footer is carried | **Emptied.** Under-cover makes conformant readers skip rows, so nothing can be kept for any column; `[]` is the honest answer. |
| `sort column` / `sort quadkey` over a glob | UNDER-cover (same, `is_partition_path`) | **Emptied.** |
| `partition *` | The whole input's stats on a fraction of its rows — far too wide, and only the first file's when the input was a glob | **Emptied.** |

## Why

For a secondary geometry column nothing recomputes on any file-write path, so a deleted key stays gone and the output is unreadable:

```
Invalid Input Error: Geoparquet column 'boundary' does not have geometry types
```

`gpio check spec` rejected the same files (7 failed checks, led by `geometry_types_list_boundary`). **gpio wrote a file gpio rejects.** #927 fixed exactly this for `convert reproject`; these call sites were untouched by it.

## Verification

Two-geometry-column fixture (`geometry`: Point, primary; `boundary`: Polygon, secondary), on `origin/main` vs this branch.

| command | `boundary.geometry_types` before | DuckDB read before | after | DuckDB read after |
|---|---|---|---|---|
| `extract geoparquet --where "id = 1"` | MISSING | FAILS | `['Polygon']` | OK (1 row) |
| `extract geoparquet --bbox 0,0,4,4` | MISSING | FAILS | `['Polygon']` | OK (3 rows) |
| `extract geoparquet '<glob>'` | MISSING | FAILS | `[]` | OK (6 rows) |
| `extract` stdin → file with `--where` | MISSING | FAILS | `['Polygon']` | OK (1 row) |
| `sort quadkey '<glob>'` | MISSING | FAILS | `[]` | OK (6 rows) |
| `sort column '<glob>' id` | MISSING | FAILS | `[]` | OK (6 rows) |
| `partition quadkey --auto` | MISSING | FAILS | `[]` | OK (3 rows) |

The exact error before, on every "FAILS" row:

```
InvalidInputException: Invalid Input Error: Geoparquet column 'boundary' does not have geometry types
```

`gpio check spec` after, on each output: `Summary: 38-39 passed, 0 warnings, 0 failed` (was 7 failed).

The primary column is still retightened everywhere — `extract --where "id = 1"` gives `geometry.bbox == [0, 0, 0, 0]`, down from the input's `[0, 0, 2, 2]`.

Suites:

- Fast: `6465 passed, 75 skipped, 9 xfailed` (`-m "not slow and not network and not meta"`)
- Meta: `203 passed`
- `pre-commit run --all-files` and `--hook-stage pre-push`: clean
- diff-cover vs `origin/main`: **100%** (22 changed lines; the one initially-missed line got a test)

Closes #934

### DIFF
--- a/geoparquet_io/core/common.py
+++ b/geoparquet_io/core/common.py
@@ -2880,7 +2880,11 @@ def write_parquet_with_metadata(
             geometry columns but not others — reproject transforms only the
             primary column, so a secondary column reaches the output unchanged
             and its carried stats still describe it (#890). None (the default)
-            invalidates every column, which is right for a row filter or a merge.
+            invalidates every column, which is right for a merge: its carried
+            stats under-cover every column of the output. An invalidated column
+            other than the primary keeps ``geometry_types`` as the spec's empty
+            "not known" list rather than losing the key, since nothing here
+            recomputes it (#934).
         drop_nonplanar_edges_columns: Geometry columns whose non-planar
             ``edges`` declaration must neither be carried through nor
             re-attached to the output. Set by writes that invalidate the edge
@@ -2902,14 +2906,30 @@ def write_parquet_with_metadata(
 
     configure_verbose(verbose)
 
+    # Use geometry column from geometry_info if provided, otherwise auto-detect
+    # This ensures original column names are preserved (fixes #328)
+    # Resolved before the invalidation below, which has to name the one column
+    # the write strategies recompute derived stats for.
+    if geometry_info and geometry_info.get("primary"):
+        geometry_column = geometry_info["primary"]
+    else:
+        geometry_column = _detect_geometry_from_query(con, query, original_metadata, verbose)
+
     # Callers that transform geometry, filter rows, or merge multiple inputs
     # invalidate the carried bbox/geometry_types; drop them so the write
     # strategies recompute (or omit) them instead of describing the input. A
     # caller that transforms only some geometry columns names them, so an
     # untouched secondary column keeps the stats that still describe it (#890).
+    #
+    # Every strategy recomputes `geometry_column` and only that, so a stripped
+    # SECONDARY column would be left with no `geometry_types` at all -- a key
+    # GeoParquet 1.1 requires and DuckDB refuses to open a file without. Naming
+    # the recomputed column leaves the others the "not known" sentinel (#934).
     if invalidate_derived_stats:
         original_metadata = strip_derived_stats(
-            original_metadata, columns=invalidate_derived_stats_columns
+            original_metadata,
+            columns=invalidate_derived_stats_columns,
+            recomputed_columns={geometry_column} if geometry_column else set(),
         )
 
     # A write that invalidates the edge interpretation for the columns it
@@ -2920,13 +2940,6 @@ def write_parquet_with_metadata(
         original_metadata = strip_nonplanar_edges(
             original_metadata, columns=drop_nonplanar_edges_columns
         )
-
-    # Use geometry column from geometry_info if provided, otherwise auto-detect
-    # This ensures original column names are preserved (fixes #328)
-    if geometry_info and geometry_info.get("primary"):
-        geometry_column = geometry_info["primary"]
-    else:
-        geometry_column = _detect_geometry_from_query(con, query, original_metadata, verbose)
 
     # A column projection (e.g. ``extract --exclude-cols``) can drop the bbox
     # column a ``covering`` points at, a secondary geometry column, or the
--- a/geoparquet_io/core/extract.py
+++ b/geoparquet_io/core/extract.py
@@ -12,6 +12,7 @@ from __future__ import annotations
 
 import json
 import sys
+from collections.abc import Collection
 from pathlib import Path
 
 import pyarrow as pa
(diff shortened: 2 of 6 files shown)


cholmes:
## Multi-agent adversarial review — approved, one S3 fixed in b87c6e1

Reviewed by agents in an isolated worktree, differentially against `origin/main` (21b14fe): every claim below is a real run on a real file, read back with **both** DuckDB and pyarrow and put through `validate_geoparquet(validate_data=True)`.

### The defect is fixed on every path, and the fix is real

Twelve command shapes that produced a file DuckDB **refuses to open** on main — `extract --where` / `--bbox` / `--limit` / over a glob / `--repair-geometry`, stdin→file, `sort quadkey`, `sort column`, `partition quadkey` / `kdtree` / `string` — now all read cleanly and pass `check spec` with **0 failed** (8 failed each on main). geopandas confirmed separately.

Item 5 exceeds what #934 asked: on `--repair-geometry` with a genuinely self-intersecting bowtie primary, the secondary keeps its real `["Polygon"]` **and** its real bbox while the primary correctly becomes `["MultiPolygon","Polygon"]`.

**The primary is not degraded anywhere.** It still recomputes and retightens on every path (`[0,0,4,4]` → `[1,1,1,1]`), across all four `--write-strategy` values and `--geoparquet-version` 1.0/1.1/2.0, with branch and main byte-identical except for the secondary's keys.

**`[]` is the right sentinel.** GeoParquet v1.1.0 makes `geometry_types` REQUIRED and says an empty array "explicitly signals that the geometry types are not known" — so `[]` means *unknown*, not *no geometries*, and the key stays present as REQUIRED demands. gpio's own validator agrees on a non-empty file. No self-rejecting output is traded in.

### S3 fixed — the stdout stream was losing precision it used to have

The scoping was applied to both of `_extract_streaming`'s outputs, but the two paths are not equally capable. A file output can only recompute the primary column, so a secondary's carried stats must be left alone. The stdout stream goes through `backfill_derived_stats`, which recomputes **every** column from the rows it writes — so scoping it there meant the secondary was never stripped, backfill saw no gap, and the input's stats shipped verbatim:

```
gpio extract geoparquet rich.parquet - --where "id = 1"     # 1 Polygon survives
before   boundary  ['Polygon','MultiPolygon','LineString']  bbox [0,0,10,10]
after    boundary  ['Polygon']                              bbox [0,0,1,1]
```

A strict loss against main on the documented pipe path, for an answer that was free. Fixed by widening the strip on the stream branch only. `--limit` and `--bbox` verified likewise; an unfiltered extract still carries its stats verbatim (nothing stale, nothing stripped); the file path is unchanged.

Three tests added, including one pinning that the **file** path still carries the secondary — widening the strip there would reintroduce #934. Sabotage-verified: disabling the fix fails exactly the new test and nothing else. This slipped past a 100% diff-cover run because the changed lines were all executed, just never with `output_path=None`.

### Wording corrected

The "the type list must be exhaustive, not exact" note overstated its licence — the spec asks that `geometry_types` be "strictly correct" and its worked example is about *under*-declaring. Over-cover is *safe* (it costs a reader a rejected candidate, never a skipped row) rather than blessed, and now reads as the tolerated fallback it is.

### Filed separately, not fixed here

- **#952** — an emptied `geometry_types` is sticky: `_compute_missing_metadata` gates on the key being absent, so no file→file command ever recomputes `[]`. Only the Arrow/stdout path recovers it.
- **#953** — the secondary's `covering` is attached to the **primary** under `in-memory` and `streaming` write strategies. Pre-existing, reproduced byte-identically on main.
- **#954** — `partition quadkey` writes a `covering` with no `bbox` key, and `geopandas.read_parquet` does an unguarded `covering["bbox"]` → `KeyError`. Pre-existing.

### Gates

`lint-imports` 4 kept / 0 broken · `ruff` + `mypy` clean · fast suite **6465 passed, 0 real failures** (3 known cold-run xdist flakes in `test_pipe_integration` and `test_journeys`; all 38 pass serially) · `diff-cover --fail-under=90` → **95%**, the one uncovered line being a pre-existing `--verbose` debug call · no SQL changed, no new interpolation.

Note CodeRabbit's green check is not a signal here — it hit its OSS rate limit and posted no review.

Reviews performed by Claude Code agents; findings verified independently before acting on them.

Issue #867, opened 2026-09-04
fix(sort,extract): directory-input hardening — self-read duplication guard and --allow-schema-diff on sort


cholmes:
Follow-ups from the adversarial review of #852 (findings from an AI review agent, all verified by execution). #852's directory-input fix itself is sound; these are shared-surface gaps at its edges:

1. **S3 — an output written inside the input directory is silently re-read as input on the next run.** `gpio sort column parts/ parts/sorted.parquet name` twice: 766 → 1532 → 2298 rows, exit 0, no warning — and `--overwrite` reads the stale output while replacing it. `extract` has the identical hazard (verified: 4596 rows after two self-read reruns), so this is the inherited convention rather than #852's bug — which is why it's filed here per the separate-PRs-for-cross-cutting-fixes convention. Guard candidate: error (or at minimum warn) in `handle_output_overwrite`/`resolve_read_path` when the resolved output path lies inside the input directory/glob. The #852 docs example writes next to the parts, so this will get hit.
2. **S3 — sort has no `--allow-schema-diff`.** A directory where a later file adds a column: the merge succeeds and the extra column silently vanishes (DuckDB glob default). `extract` exposes `--allow-schema-diff` (union_by_name); sort plumbs nothing into `build_read_parquet_expr(allow_schema_diff=...)`. Decide whether sort should grow the same flag (CLI + API twins + docs) or the docs should state the asymmetry and point at extract-then-sort. (#852 now reports mismatched-schema directories cleanly and hints at extract; this issue covers the union_by_name escape hatch itself.)
3. **S4s for the same sweep:** empty-dir error should say "no .parquet files found" rather than "Cannot read file: <dir>"; `sort_quadkey.py` discards `resolve_read_path`'s options dict, so hive keys survive only via DuckDB's auto-detect default; `con.close()` in `sort_quadkey.py`'s `finally` raises UnboundLocalError if the connection constructor throws (pre-existing on main).

Refs #852, #817.

Issue #718, opened 2026-08-23
Input paths are interpolated into DuckDB SQL as raw string literals (55 sites)


cholmes:
## What

Across `geoparquet_io/core/`, file paths and URLs are interpolated straight into DuckDB SQL as string literals:

```python
query = f"""
    SELECT *, ... FROM '{input_url}'
"""
```

A quick count: **55 raw `FROM '{...}'` interpolations, against 1 that routes through `_escape_sql_string()`.**

```
$ grep -rn "FROM '{" geoparquet_io/core/ | wc -l
55
```

Examples: `core/sort_quadkey.py:247,376`, `core/check_spatial_order.py:48,66,106`, `core/add/quadkey.py:642`.

## Why it matters

CLAUDE.md already states the rule — `_escape_sql_string()` for SQL string literals, `quote_identifier()` for identifiers, never hand-rolled, escape exactly once — and the identifier half has been enforced and fixed repeatedly (#657, #662, #695). The string-literal half was never swept.

A path containing a single quote terminates the literal. That is a crash for an ordinary user with an apostrophe in a filename, and an injection surface anywhere a path reaches gpio from somewhere other than the operator's own shell: the Python API (`ops.*`, `Table.*`), a service wrapping gpio, or a remote URL assembled from user input.

```
in.parquet' UNION SELECT ... --
```

## Why this is its own issue

Found while reviewing #695, which correctly fixed the *identifier* quoting in `_add_quadkey_file_based` and left `FROM '{input_url}'` on the line below it untouched. Fixing it there would have been a one-line change to one of 55 sites, which is worse than leaving it alone: it converts a uniform, greppable problem into a partially-fixed one.

## Suggested shape

1. Add a helper that takes a path/URL and returns a fully-quoted SQL literal, so call sites read `FROM {sql_path(input_url)}` and the quotes stop being written by hand.
2. Sweep the 55 sites.
3. Add a pre-commit check to `duckdb-antipatterns` for `FROM '{`, the same way the identifier rules are enforced today.
4. Cover it with a hostile-path test mirroring the hostile-name tests #695 added.

Discovered during adversarial review of #695.

cholmes:
Closed by #803, which delivers asks 1, 3 and 4 of the suggested shape.

[...]

Issue #232, opened 2026-02-04
Performance regression: 80-130% slower since v0.7.0


nlebovits:
## Summary

Cross-version benchmarking reveals an **80-130% performance regression** (not 25% as originally estimated) in `gpio inspect` between v0.4.0 and v0.7.0+. The regression has persisted through v0.8.0 and v0.9.0.

**Related:** #129

---

## Benchmark Data

| Version | Tiny File | Small File | vs v0.4.0 |
|---------|-----------|------------|-----------|
| **v0.4.0** | 0.29s | 0.22s | baseline |
| v0.7.0 | 0.53s | 0.51s | +82% / +130% |
| v0.8.0 | 0.52s | 0.54s | +80% / +144% |
| v0.9.0 | 0.52s | 0.52s | +78% / +136% |

---

## Root Cause

The switch from PyArrow to DuckDB for **all** metadata reads in v0.7.0. Profiling breakdown:

| Component | Time | Notes |
|-----------|------|-------|
| `duckdb.connect()` | 17ms | Per connection |
| `LOAD spatial` | 105ms | Per connection |
| `detect_geoparquet_file_type()` | 85ms | Creates 2 connections per call |
| PyArrow equivalent | 0.03ms | 1000x faster |

`inspect` command breakdown:
- 92% of time spent in `get_duckdb_connection()`
- Only 5% on actual inspection logic

---

### 1. Use PyArrow for local file metadata (2-3 hrs, 25-40% gain)
**Files:** `geoparquet_io/core/duckdb_metadata.py`, `geoparquet_io/core/common.py`

- Add `is_remote_url()` check at start of metadata functions
- Use `pyarrow.parquet.read_metadata()` and `read_schema()` for local files
- Keep DuckDB for remote files (S3, HTTP) where PyArrow needs fsspec

### 2. Connection reuse in `detect_geoparquet_file_type()` (1 hr)
**Files:** `geoparquet_io/core/common.py`

- Accept and pass `con` parameter to avoid creating 2 connections per call
- Reuse single connection for both `get_geo_metadata()` and `detect_geometry_columns()`

### 3. Lazy spatial extension loading (1 hr)
**Files:** `geoparquet_io/core/common.py`

- Only `LOAD spatial` when actually needed for spatial operations
- Metadata queries don't need the spatial extension (saves 105ms per connection)

### 4. LRU cache on `detect_geoparquet_file_type()` (30 min)
**Files:** `geoparquet_io/core/common.py`

[...]

PR #955, opened 2026-09-09
fix(inspect): stop counting bbox struct children as top-level columns

### LINKED ISSUE #931: fix(inspect,metadata): struct children counted as top-level columns — 'Columns: 14' for a 10-column file
Found during the sibling sweep for #886 (AI review agent), then confirmed user-visible.

## The symptom

`gpio inspect` disagrees with itself about how many columns a file has:

```
$ gpio inspect head tests/data/austria_bbox_covering.parquet
Columns (10):

$ gpio inspect meta tests/data/austria_bbox_covering.parquet
Columns: 14
```

The extra four are `xmin`, `ymin`, `xmax`, `ymax` — the children of the `geometry_bbox` covering struct, counted as if they were top-level columns:

```python
>>> from geoparquet_io.core.duckdb_metadata import get_column_names
>>> get_column_names("tests/data/austria_bbox_covering.parquet")
['id', 'inspire:id', 'crop:code', 'crop:name', 'area', 'determination_datetime',
 'admin:country_code', 'crop:code_list', 'geometry', 'geometry_bbox',
 'xmin', 'ymin', 'xmax', 'ymax']
```

Any file with a bbox covering — which is what gpio's own `add bbox` writes, and what the guide recommends — reports four phantom columns.

## Cause

Three places filter the flat Parquet schema listing to "root columns" with a `"." not in name` test:

- `core/duckdb_metadata.py::get_column_names`
- `core/duckdb_utils.py::_DuckDBSchemaWrapper`
- `core/metadata_utils.py` (the column count)

Neither the DuckDB nor the pyarrow schema listing renders dotted paths, so the test never excludes anything and struct children pass straight through. The filter has presumably never worked.

## Fix

`core/validate.py` now has `_root_schema_index()` / `_schema_direct_children()` (added by #930 for the covering-bbox lookup), which walk the listing subtree by subtree and handle both schema shapes — DuckDB's listing prefixes the file's root group element, the pyarrow fast path starts at the first column. That is the correct primitive; these three call sites should use it rather than each growing their own.

## Related, larger

The same first-name-match scan appears in ~12 geometry-column lookups in `validate.py`, plus `common._schema_struct_child_names` and `common._find_bbox_column_in_schema`. Those are a shadowing hazard rather than a miscount (a nested field named `geometry` would win over the real one), and a dozen call sites is its own PR — noted here so the two are not fixed twice independently.

Refs #886, #930, #885.

*Filed by Claude (AI) from the #886 sweep; the inspect discrepancy verified directly.*

cholmes:
## What changed

`inspect head` and `inspect meta` now agree on the column count of a file with
a covering bbox.

Three call sites filtered the flat Parquet schema listing down to root-level
columns with a `"." not in name` test:

- `core/duckdb_metadata.py::get_column_names`
- `core/duckdb_utils.py::_DuckDBSchemaWrapper`
- the column count and schema string in `core/metadata_utils.py`

All three now call a shared `root_schema_columns()`.

The subtree walk that makes this correct already existed in `core/validate.py`
(`_root_schema_index` / `_schema_direct_children`, added by #930 for the
covering-bbox root-placement check). Rather than import private helpers out of
`validate.py` from three other core modules, they move to a new
`geoparquet_io/core/parquet_schema.py`: the concept is "walk the flat Parquet
schema listing", it is not DuckDB-specific (it handles both the DuckDB
`parquet_schema()` shape, which prefixes the listing with the file's root group
element, and the pyarrow fast path, which does not), and the module imports
nothing from the package, so all four consumers share it with no import cycle.
`validate.py` keeps its behaviour and now imports the public names.

`tests/test_metadata_utils.py::test_json_output` asserted `num_columns == 10`
for `places_test.parquet`, which has six root columns plus a four-field `bbox`
covering struct. That assertion had baked in the bug; it is now 6, with a
comment saying why.

## Why

Closes #931.

```
$ gpio inspect head tests/data/austria_bbox_covering.parquet
Columns (10):
$ gpio inspect meta tests/data/austria_bbox_covering.parquet
Columns: 14
```

The extra four are `xmin`/`ymin`/`xmax`/`ymax`, the children of the
`geometry_bbox` covering struct, counted as top-level columns:

```python
>>> get_column_names("tests/data/austria_bbox_covering.parquet")
[... 'geometry', 'geometry_bbox', 'xmin', 'ymin', 'xmax', 'ymax']
```

The listing is depth first — a struct's children follow their parent — and
neither shape renders dotted paths, so the `"." not in name` filter never
excluded anything. It has never worked. Every file with a bbox covering is
affected, which is what `gpio add bbox` writes and what the guide recommends.

Beyond the miscount, `get_column_names()` feeds membership checks in
`partition/by_string.py`, `constants.py` and `sort_quadkey.py`, where a struct
child could satisfy a "does this column exist" test for a column that is not
addressable at the root.

## Verification

After:

```
$ gpio inspect head tests/data/austria_bbox_covering.parquet
Columns (10):
$ gpio inspect meta tests/data/austria_bbox_covering.parquet
Columns: 10
```

New `tests/test_root_schema_columns.py` (9 tests) covers: the covering file's
column names and count; `inspect head` and `inspect meta` agreeing; the
`_DuckDBSchemaWrapper` count; a struct nested inside a struct; both listing
shapes (passing a connection forces the DuckDB path, which — unlike the pyarrow
fast path — renders grandchildren, so it is the shape that exercises the
recursive walk); and truncated/empty listings.

- Fast suite: `6475 passed, 75 skipped, 9 xfailed`. Two `-n auto` failures
  (`test_journeys.py::test_journey_01_...`, `test_geojson_stream.py::...broken_pipe`)
  were machine-saturation noise and pass serially.
- Meta lane: `203 passed`.
- `diff-cover coverage.xml --compare-branch=origin/main`: **100%** over 49
  changed lines.
- `pre-commit run --all-files`: clean. `--hook-stage pre-push`: clean except
  `menard-check`, which flags `docs/guide/check.md` as stale against
  `core/validate.py` — pre-existing on `main` (the doc last moved 2026-09-05,
  behind #876, #904, #930 and #939), not introduced here.

### Sweep

`"." not in` is now gone from `geoparquet_io/` entirely; the remaining `.split(".")`
hits are version and BigQuery table-id parsing, unrelated.

Not fixed here, worth its own PR: ~12 geometry-column lookups in `validate.py`,
plus `common._schema_struct_child_names` and `common._find_bbox_column_in_schema`,
still scan the listing for the *first* entry with a given name. That is a
shadowing hazard (a nested field named `geometry` beating the real root column)
rather than a miscount, and `root_schema_index()` is now the primitive that fixes
it across all of them.

Closes #931

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected column counting and listing so nested struct fields are excluded from top-level column results.
  * Improved consistency between DuckDB and PyArrow schema handling.
  * Ensured metadata displays accurate root-level column counts.
  * Improved handling of empty, truncated, and nested schema definitions.

* **Tests**
  * Added coverage for nested structures, covering-bounding-box fields, schema variations, empty names, and incomplete schema listings.

### DIFF
--- a/geoparquet_io/core/duckdb_metadata.py
+++ b/geoparquet_io/core/duckdb_metadata.py
@@ -166,12 +166,23 @@ def _pyarrow_get_schema_info(parquet_file: str) -> list[dict] | None:
             if isinstance(field_type, pa.BaseExtensionType):
                 field_type = field_type.storage_type
 
+            # Only field-bearing types are iterable, so a list/map column
+            # renders as a single row with no children while a struct renders
+            # its fields. ``num_children`` therefore has to count the rows this
+            # listing actually goes on to emit, not the type's fields --
+            # consumers read the listing as a depth-first tree and skip a
+            # subtree by that count, so a column claiming children it never
+            # emits swallows the next root column (issue #931). Nested struct
+            # children are emitted with ``num_children: 0`` for the same
+            # reason: their own fields are not rendered either.
+            children = list(field.type) if hasattr(field.type, "__iter__") else []
+
             col_info = {
                 "name": field.name,
                 "type": str(field_type),
                 "type_length": None,
                 "repetition_type": "OPTIONAL" if field.nullable else "REQUIRED",
-                "num_children": 0,
+                "num_children": len(children),
                 "converted_type": None,
                 "scale": None,
                 "precision": None,
@@ -181,28 +192,22 @@ def _pyarrow_get_schema_info(parquet_file: str) -> list[dict] | None:
                 "logical_type": logical_type,
             }
 
-            # Handle struct types
-            if hasattr(field.type, "num_fields"):
-                col_info["num_children"] = field.type.num_fields
-
             result.append(col_info)
 
-            # Add child fields for struct types
-            if hasattr(field.type, "__iter__"):
-                for child_field in field.type:
-                    child_info = {
-                        "name": child_field.name,
-                        "type": str(child_field.type),
-                        "type_length": None,
-                        "repetition_type": "OPTIONAL" if child_field.nullable else "REQUIRED",
-                        "num_children": 0,
-                        "converted_type": None,
-                        "scale": None,
-                        "precision": None,
-                        "field_id": None,
-                        "logical_type": None,
-                    }
-                    result.append(child_info)
+            for child_field in children:
+                child_info = {
+                    "name": child_field.name,
+                    "type": str(child_field.type),
+                    "type_length": None,
+                    "repetition_type": "OPTIONAL" if child_field.nullable else "REQUIRED",
+                    "num_children": 0,
+                    "converted_type": None,
+                    "scale": None,
+                    "precision": None,
+                    "field_id": None,
+                    "logical_type": None,
+                }
+                result.append(child_info)
 
         return result
     except ValueError as e:
--- a/geoparquet_io/core/duckdb_metadata.py
+++ b/geoparquet_io/core/duckdb_metadata.py
@@ -22,6 +22,7 @@ import pyarrow.parquet as pq
 from geoparquet_io.core.crs_utils import geoarrow_crs_to_projjson, merge_longitude_ranges
 from geoparquet_io.core.duckdb_utils import _escape_sql_string, sql_path
 from geoparquet_io.core.exceptions import GeoParquetError
+from geoparquet_io.core.parquet_schema import root_schema_columns
 
 # =============================================================================
 # PyArrow Fast Path for Local Files (Issue #232 Performance Fix)
@@ -553,8 +554,9 @@ def get_schema_info(parquet_file: str, con=None) -> list[dict]:
 def get_column_names(parquet_file: str, con=None) -> list[str]:
     """Get list of column names from schema."""
     schema = get_schema_info(parquet_file, con)
-    # Filter out empty names (schema root element) and nested struct fields
-    return [col["name"] for col in schema if col.get("name") and "." not in col["name"]]
+    # Filter out the schema root element and nested struct fields: the listing is
+    # depth first, so a struct's children have to be skipped as whole subtrees.
+    return [col["name"] for col in root_schema_columns(schema)]
 
 
 def get_usable_columns(parquet_file: str, con=None) -> list[dict]:
--- a/geoparquet_io/core/duckdb_utils.py
+++ b/geoparquet_io/core/duckdb_utils.py
@@ -15,6 +15,7 @@ from contextvars import ContextVar
 import duckdb
 
 from geoparquet_io.core.logging_config import warn
+from geoparquet_io.core.parquet_schema import root_schema_columns
 
 # Per-bucket cache for S3 buckets that require authentication
 # Buckets not in this set are accessed without credentials (works for public buckets)
@@ -791,7 +792,7 @@ class _DuckDBSchemaWrapper:
     """Wrapper to provide PyArrow-like interface for DuckDB schema info."""
 
     def __init__(self, schema_info):
-        self._columns = [c for c in schema_info if c.get("name") and "." not in c.get("name", "")]
+        self._columns = root_schema_columns(schema_info)
 
     def __len__(self):
         return len(self._columns)
--- a/geoparquet_io/core/metadata_utils.py
+++ b/geoparquet_io/core/metadata_utils.py
@@ -12,6 +12,7 @@ from rich.table import Table
 from rich.text import Text
 
 from geoparquet_io.core.common import format_size
+from geoparquet_io.core.parquet_schema import root_schema_columns
 
 
 def _calculate_overall_bbox(row_group_stats: list[dict]) -> dict[str, float] | None:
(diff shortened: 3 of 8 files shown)


cholmes:
## Multi-agent adversarial review — S1 found and fixed in b12a88b

Reviewed in an isolated worktree, rebased onto current `main` (`4af17e8`) first. **The review found an S1 that every gate passed straight through**, including a 100% `diff-cover` result — which is worth dwelling on, because it is a clean example of line coverage certifying wrong lines.

### S1 — the fix silently deleted real root columns

`schema_subtree_end()` trusted `num_children` to say how many *emitted rows* follow an entry. That holds for the DuckDB shape. It does not hold for the pyarrow fast path — the path used for every **local** file, i.e. the normal CLI path. `_pyarrow_get_schema_info` set `num_children = field.type.num_fields` but only emitted child rows for iterable types, and only `StructType` is iterable:

| pyarrow type | `num_children` written | child rows emitted |
|---|---|---|
| `list`, `large_list`, `fixed_size_list`, `map` | 1 | **0** |
| `struct` | n | n |

So every root-level list or map column swallowed **the next root column**. On the repo's own Overture fixture:

```
$ gpio inspect head tests/data/country_partition/El_Salvador.parquet   ->  Columns (19):
$ gpio inspect meta tests/data/country_partition/El_Salvador.parquet   ->  Columns: 13
```

`names`, `socials`, `emails`, `phones`, `brand`, `operating_status` — gone. The PR re-created the exact head-vs-meta disagreement it exists to fix, and this time `meta` *under*-reported, hiding real columns rather than inventing phantom ones.

It was not cosmetic. Verified before and after:

| | before | after |
|---|---|---|
| `inspect head` / `inspect meta` | 19 / **13** | 19 / **19** |
| `partition string --column operating_status` | `Error: 'operating_status' not found in the Parquet file` | succeeds |
| `add quadkey` on `[tags list, quadkey, geometry]` | succeeded, wrote **both** `quadkey` and `quadkey_1` | `Error: Column 'quadkey' already exists` |
| `check spec`, covering bbox after a list column | **3 failed** (`bbox column "bbox" is not at the schema root`) | **0 failed** |

The blast radius was 11 call sites, not the 3 the body named — including `common.py:3856`/`:3976` and `constants.py:123`, where `ensure_vecorel_columns` would add a second `id` and rewrite the file **in place**.

### Why the gates missed it

Every fixture in the new `tests/test_root_schema_columns.py` was struct-only — no list, map, or fixed-size-list column anywhere. `diff-cover` reported **100% over 49 changed lines** while those lines were wrong: line coverage of a walk says nothing about whether `num_children` is trustworthy. The fix adds root-level `list`, `large_list`, `fixed_size_list`, `map` and `struct` cases, each followed by a plain column, asserted in **both** schema shapes and asserted to agree with each other.

### The fix, and a regression it would otherwise have introduced

`num_children` now counts the rows the listing actually emits. That was already the convention for nested struct children, so the root loop just obeys the same rule — and it makes the walk correct for both shapes with no special-casing, rather than teaching the consumer about producer quirks.

Falling out of that: `validate.py::_check_geometry_not_grouped` tested `num_children > 0`. Once `num_children` became honest, a geometry column stored as a **list** under a declared `WKB` encoding would have silently started passing. A new `schema_entry_is_group()` reads groupness off the type instead, with three tests pinning it.

Worth recording that the defensive guard I asked for would **not** have caught the original bug — `sources` renders as `list<...>`, so it does not look like a leaf. That was reported plainly rather than claimed as closing the class, and the real producer-side invariant is documented in the module instead.

### The S2 was pre-existing, and this fixes it

The covering-bbox false failure reproduces on `origin/main` — `validate.py:1783` carries the identical walk from #930 — so it arrived with that PR, not this one. Fixing the root cause resolves it.

- **The changed assertion is a false alarm; the PR is right.** `places_test.parquet` genuinely has 6 root columns plus a 4-field `bbox` struct = the old 10. The old assertion encoded the bug.
- **`validate.py`'s extraction is behaviour-preserving.** The four moved helpers are byte-identical modulo the underscore, and `validate_geoparquet()` over all 40 fixtures produced **zero differing check results** between revisions.
- **The `"." not in name` sweep is complete**, and the escaping contract at `constants.py:144` is intact — `parquet_schema.py` handles no paths at all, so there is no #718 exposure.

Unclaimed bonus the review turned up: a column literally named `a.b` is now correctly kept, where main dropped it.

### Filed as follow-up

**#963** — these 11 membership checks are really asking "can the SQL I am about to generate name this column?", which is a question about what `read_parquet()` exposes, not about the physical schema. `get_usable_columns()` answers it directly and `sort_by_column.py:204` already uses it. The walk structurally cannot see `--allow-schema-diff` union columns (#867) or hive partition columns. Correctly left out of this PR: it costs ~44 ms against ~0.1 ms on paths #232 exists to keep fast, and it is a real behaviour change. The issue also notes `constants.py:144`'s comment describes the escaping backwards.

### Gates

Fast suite **6738 passed, 75 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100% on 63 changed lines** · `pytest -m meta` 205 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push clean · 13 new tests, all verified failing before the fix.

Issue #892, opened 2026-09-05
fix(write): in-memory strategy drops the M suffix from geometry_types declarations


cholmes:
Found during the adversarial review of #888 (AI review agent; pre-existing on main, reproduced identically on main, the #888 branch, and a main+#888+#869 tree — untouched by any of them).

The in-memory write strategy declares `geometry_types: ['Polygon']` for `Polygon M` data, dropping the M suffix, while streaming and disk-rewrite correctly declare `['Polygon M']`. Per spec the dimensional suffixes (" Z", " M", " ZM") are part of the type name, and #879's (pending) geometry_types-vs-statistics check would flag exactly this drift.

Where: `geoparquet_io/core/write_strategies/arrow_memory.py` (its geometry-type computation). The repair helper's `_geo_code_to_type_name` (common.py:2064) has the correct spellings to borrow.

Refs #888, #785.

PR #956, opened 2026-09-09
fix(extract): compute arcgis geometry_types from the data, not a five-entry table

### LINKED ISSUE #928: fix(arcgis): unknown geometry types write "Geometry", which is not a valid geometry_types value
Found during the sibling sweep for #892 (AI review agent). Verified against the vendored GeoParquet 2.0.0 schema.

## The defect

`geoparquet_io/core/arcgis.py:1377` builds the geo block with:

```python
"geometry_types": [
    ARCGIS_GEOM_TYPES.get(layer_info.geometry_type, "Geometry")
],
```

`"Geometry"` is not a value the spec allows. The schema's enum pattern is:

```
^(GeometryCollection|(Multi)?(Point|LineString|Polygon))( Z| M| ZM)?$
```

so `"Geometry"` matches nothing, and any ArcGIS layer whose `geometryType` is outside the five-entry `ARCGIS_GEOM_TYPES` table (`arcgis.py:52-58`) produces a file that fails schema validation on its own `geo` block.

The table covers `esriGeometryPoint`, `esriGeometryMultipoint`, `esriGeometryPolyline`, `esriGeometryPolygon` and `esriGeometryEnvelope`. Anything else — and the Esri type list is longer than that — takes the fallback.

## Why the fallback is the wrong shape

The spec already has a way to say "I do not know the geometry types": **an empty array**. From the geometry_types section:

> An empty array explicitly signals that the geometry types are not known.

So the honest fallback is `[]`, not an invented type name. That is also strictly better than guessing, because a wrong declared type is worse than an absent one — `gpio check spec` enforces `geometry_types` against the data, and #879's geometry_types-vs-statistics check will flag a mismatch.

Worth considering as part of the fix: rather than falling back at all, compute the real types from the fetched WKB the way the write path does. `arcgis` already materialises the table before writing, so the information is in hand. That would make the mapping table a fast path rather than the only source of truth. Whoever picks this up should decide between "empty array fallback" (small, safe) and "compute from data" (better, slightly more work) — I would take the second if the write path's helper is reusable, the first otherwise.

A regression test wants a layer whose `geometryType` is outside the table; the resulting file should validate.

Refs #892, #879.

*Filed by Claude (AI) from the #892 sibling sweep.*

cholmes:
## What changed

`gpio extract arcgis` now computes the `geo` block's `geometry_types` from the fetched WKB instead of translating the layer's advertised `geometryType` through a five-entry table.

- New `_resolve_geometry_types()` in `core/arcgis.py` calls `_compute_geometry_types()` from `core/geo_metadata.py` — the same helper the write paths use, fixed in #892/#927 to recombine geoarrow's `geometry_type` and `dimensions` fields.
- `ARCGIS_GEOM_TYPES` survives only as a fallback for WKB that cannot be read. An Esri type outside it now yields `[]` — the spec's own way of saying the types are not known — never an invented name.
- Geometry repair (#506) moved above the geo block. `ST_MakeValid` can change a geometry's type (a bowtie Polygon comes back as a MultiPolygon), so the declaration has to describe the repaired data that actually gets written. The helper preserves schema metadata, so it is equally safe on either side.

**User-visible metadata change.** The table mapped `esriGeometryPolyline` → `MultiLineString` and `esriGeometryPolygon` → `MultiPolygon` unconditionally. Computed types win, so a polygon layer holding single Polygons is now declared `["Polygon"]` rather than `["MultiPolygon"]`, a mixed layer declares every type present, and Z/M layers get the `" Z"` / `" M"` / `" ZM"` suffixes the table could never express. In every case the new value is what the data actually contains.

## Why

`ARCGIS_GEOM_TYPES.get(layer_info.geometry_type, "Geometry")` fell back to the literal `"Geometry"`, which the GeoParquet schema does not permit:

```
^(GeometryCollection|(Multi)?(Point|LineString|Polygon))( Z| M| ZM)?$
```

Esri's geometry-type list is longer than the five entries in the table, so any layer outside it (`esriGeometryMultiPatch`, for one) wrote a file that failed validation on its own metadata. Before this change, `gpio check spec` on such a file reported two failures:

```
column "geometry" has invalid geometry_types: ['Geometry']
found undeclared geometry types: {'Point'} (1 checked)
```

The issue offered an empty-array fallback or computing from the data, and asked for the second if the write path's helper was reusable. It was, directly — the extraction path already materialises the table before writing, so the real types were in hand — and it is the better answer: a declared-but-wrong type is worse than an absent one, and #879's `geometry_types`-vs-statistics check would have flagged the guesses the old table made.

## Verification

- New `tests/test_arcgis_geometry_types.py` (6 tests, all failing before the fix): unknown Esri type computes `["Point"]` and matches the schema pattern; unreadable WKB on an unknown type declares `[]`; single Polygons stay `Polygon` against the `MultiPolygon` mapping; `LineString Z` keeps its suffix; mixed geometries are all declared; and end to end, a file written for `esriGeometryMultiPatch` passes `validate_geoparquet` with zero failed checks.
- Fast suite: `uv run pytest -n auto -m "not slow and not network and not meta" -q` → 6472 passed, 75 skipped, 9 xfailed. Two unrelated `-n auto` saturation failures (`test_journey_01_geojson_to_geoparquet_and_back`, `test_streaming_handles_broken_pipe`) both pass on a serial re-run.
- Meta lane: `uv run pytest -m meta -q` → 203 passed.
- `uv run pre-commit run --all-files` and `--hook-stage pre-push` both clean (mypy, vulture, xenon, import-linter, deptry included).
- `diff-cover coverage.xml --compare-branch=origin/main --fail-under=90` → 9 changed lines, 0 missing, **100%**.

Closes #928

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected GeoParquet geometry metadata for ArcGIS extracts by deriving geometry types from the actual written data.
  - Improved handling of repaired, mixed, 3D, and previously unsupported geometry types.
  - Prevented invalid generic geometry labels from appearing in metadata.
  - Ensured files with unknown or unreadable geometry types remain valid with an empty geometry-type list.

### DIFF
--- a/geoparquet_io/core/arcgis.py
+++ b/geoparquet_io/core/arcgis.py
@@ -28,6 +28,7 @@ from geoparquet_io.core.exceptions import (
     InvalidParameterError,
     RemoteAccessError,
 )
+from geoparquet_io.core.geo_metadata import _compute_geometry_types
 from geoparquet_io.core.geometry_repair import repair_arrow_table_geometry
 from geoparquet_io.core.http_retry import (
     DEFAULT_TIMEOUT,
@@ -1195,6 +1196,30 @@ def _stream_features_to_parquet(
             writer.close()
 
 
+def _resolve_geometry_types(table: pa.Table, esri_geometry_type: str, verbose: bool) -> list[str]:
+    """Geometry types for the geo block, read from the fetched WKB (#928).
+
+    The layer's advertised ``geometryType`` is only a fallback. It is coarser
+    than the data (Esri's ``esriGeometryPolyline`` covers both LineString and
+    MultiLineString, and says nothing about Z/M), it does not cover every Esri
+    type, and it used to fall back to the literal ``"Geometry"`` -- which the
+    GeoParquet schema does not allow, so unmapped layers wrote files that
+    failed validation on their own metadata.
+
+    Computing from the data instead is what every gpio write path does, gives
+    the spec's " Z"/" M"/" ZM" suffixes for free, and keeps the declaration
+    consistent with the statistics a reader checks it against. When the WKB
+    cannot be read the mapping table still serves as a fallback, and an Esri
+    type outside it yields ``[]`` -- the spec's way of saying the types are
+    not known.
+    """
+    computed = _compute_geometry_types(table, "geometry", verbose)
+    if computed:
+        return computed
+    declared = ARCGIS_GEOM_TYPES.get(esri_geometry_type)
+    return [declared] if declared else []
+
+
 def arcgis_to_table(
     service_url: str,
     auth: ArcGISAuth | None = None,
@@ -1348,6 +1373,14 @@ def arcgis_to_table(
                 table = table.select(cols_to_keep)
                 debug(f"Excluded columns: {cols_to_exclude}")
 
+        # Repair invalid geometry (issue #506) before the geo block is built:
+        # ST_MakeValid can change a geometry's type (a bowtie Polygon comes back
+        # as a MultiPolygon), so the types have to be read off the repaired data
+        # that is actually written. The helper preserves schema metadata, so it
+        # is equally safe on either side of the block.
+        if table.num_rows > 0:
+            table, _ = repair_arrow_table_geometry(table, "geometry", repair=repair_geometry)
+
         # Add CRS to metadata.
         # Default (GeoJSON path): always CRS84 (WGS84 lon/lat) because we request
         # f=geojson, which per RFC 7946 is always WGS84 regardless of the native SR.
@@ -1374,9 +1407,9 @@ def arcgis_to_table(
                     "geometry": {
                         "encoding": "WKB",
                         "crs": crs,
-                        "geometry_types": [
-                            ARCGIS_GEOM_TYPES.get(layer_info.geometry_type, "Geometry")
-                        ],
+                        "geometry_types": _resolve_geometry_types(
+                            table, layer_info.geometry_type, verbose
+                        ),
                     }
                 },
             }
@@ -1385,10 +1418,6 @@ def arcgis_to_table(
             existing_metadata = table.schema.metadata or {}
             new_metadata = {**existing_metadata, b"geo": json.dumps(geo_metadata).encode("utf-8")}
             table = table.replace_schema_metadata(new_metadata)
-
-        # Repair invalid geometry (issue #506). Helper preserves schema metadata.
-        if table.num_rows > 0:
-            table, _ = repair_arrow_table_geometry(table, "geometry", repair=repair_geometry)
 
         success(f"Converted {table.num_rows} features")
         return table
--- a/tests/test_arcgis_geometry_types.py
+++ b/tests/test_arcgis_geometry_types.py
@@ -69,14 +69,15 @@ def _stub_stream(tmp_path, geometries):
 def run_extract(tmp_path, monkeypatch):
     """Run ``arcgis_to_table`` over stubbed network calls, returning its geo block."""
 
-    def _run(geometry_type, geometries):
+    def _run(geometry_type, geometries, **kwargs):
         from geoparquet_io.core import arcgis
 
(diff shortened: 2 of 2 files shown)


cholmes:
## Multi-agent adversarial review — merge-ready, one test added in 142f59b

Reviewed in an isolated worktree, rebased onto current `main` (`4af17e8`) first since the branch predated #911/#945/#948/#949. Rebase produced exactly the two intended hunks with no phantom artifacts, and no interaction with #949's derived-stats changes was found (arcgis writes through `_apply_geoparquet_metadata`, which *carries* a present `geometry_types` rather than recomputing it — confirmed by watching the arcgis-level values reach the file byte-for-byte).

### My review brief was wrong on three points — worth recording

I asked the reviewer to attack paging, on the theory this might **under**-declare. It refuted that, and the risk runs the other way:

- **There is no partial-table path.** `_stream_features_to_parquet` writes every page to one temp parquet, which is fully materialised at `arcgis.py:1363` before `_resolve_geometry_types` is called at `:1409`. `--limit` truncates before the write too, so the declaration always describes exactly the rows written.
- **The OLD value was the under-declaration.** GeoParquet 1.1.0 says verbatim: *"it is not sufficient to specify `["MultiPolygon"]`, but it is expected to specify `["Polygon", "MultiPolygon"]`"*. A polygon layer of single Polygons declaring `["MultiPolygon"]` was under-declaring `Polygon`, and `gpio check spec` said so. This change moves gpio **toward** the spec.
- **The tests are not network-marked.** `tests/test_arcgis_geometry_types.py` carries no markers at all (unlike `tests/test_arcgis.py`), so it runs in the blocking fast lane. All 6 fail against the merge base.

Z/M suffixes verified end to end through the default path: a `LineString Z` layer writes `["LineString Z"]` and the WKB really carries Z, via the #892/#927 `geometry_type`+`dimensions` recombination.

### Added before merge: the reorder had nothing holding it in place

Moving the repair back below the geo block left **all six new tests green** while the code emitted the exact spec-failing under-declaration the reorder exists to prevent:

```
esriGeometryPolygon layer holding a bowtie POLYGON
repair above the block   declared ['MultiPolygon']   data MultiPolygon   ok
repair below the block   declared ['Polygon']        data MultiPolygon   <- undeclared type
```

A conformant reader skips rows whose type is not declared, so that is data loss rather than a cosmetic mismatch. The PR body spends a paragraph justifying this move; it now has the test that paragraph describes.

Two tests pin it from both sides — the bowtie under default repair must declare `MultiPolygon`, and the same bowtie with `repair_geometry=False` must declare `Polygon`, so a hardcoded answer cannot satisfy both. Each reads the written column back with shapely and asserts the declaration matches what actually landed, rather than trusting the geo block to describe itself. Sabotage-verified: reverting the reorder fails exactly the new test and nothing else.

Metadata safety across the move was checked too — `repair_arrow_table_geometry` restores the schema metadata after its DuckDB round trip and returns the input unchanged on exception, on a missing geometry column, on `repair=False` and on zero rows. `covering`/`edges`/`orientation`/`epoch` are added downstream, past both positions.

### Filed as follow-up, not fixed here

**#962** — `if computed:` cannot distinguish "geoarrow could not read this" from "there was nothing to describe", so both fall through to the mapping table's guess. An all-NULL geometry column (reachable from FeatureServer records with `"geometry": null`, or a `--where` selecting only those) declares `["MultiPolygon"]` over a column containing zero geometries — the declared-but-wrong value this PR exists to stop producing. Low impact, since arcgis geometry always comes from DuckDB `ST_AsWKB` which does not emit unreadable WKB, but it contradicts the PR's own thesis and deserves its own change.

### Gates

Fast suite **6723 passed, 75 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100% on 9 changed lines** · `pytest -m meta` 205 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push hooks clean.

Reviews performed by Claude Code agents; every finding independently reproduced before acting on it.

Issue #795, opened 2026-09-02
gpio check calls 10,000-200,000 rows per group optimal while advising 10,000-50,000


cholmes:
`gpio check` carries two different row-group numbers, and they disagree with each
other and with the sort default.

## The disagreement

`assess_row_count` (`core/check_parquet_structure.py:134`) calls anything between
**10,000 and 200,000** rows per group "optimal":

```python
elif 10000 <= avg_rows <= 200000:
    return "optimal", "Row count per group is optimal", "green"
```

Immediately below, the guidelines the same command prints
(`core/check_parquet_structure.py:255-258`) advise a *narrower* band for the
workload the tool is mostly used for:

```
- Optimal rows: 10,000-200,000 rows per group
- Spatial queries: 10,000-50,000 rows per group with Hilbert sorting
  and GeoParquet v2.0 enables optimal row group skipping
```

So `gpio check row-group` can print "Row count per group is optimal" for a
150,000-row-group file and, in the next breath, advise 10,000-50,000. The same
10k-50k figure is what `docs/guide/check.md` and the optimization score's "Row
Group Size" factor quote.

`core/check_fixes.py` is a third number again: `fix_compression`,
`fix_bbox_column`, `fix_bbox_removal`, `fix_spatial_order` and `fix_row_groups`
all hard-code `row_group_rows=100000` (lines 93, 151, 232, 305, 352), so
`gpio check --fix` rewrites files to 100,352-row groups regardless of what the
check just advised.

## Why this is filed separately

#775 lowered the **`gpio sort`** default to 50,000 rows per group, the top of
the 10k-50k band. That fixed the sort side of the inconsistency and left this
one alone deliberately: narrowing `assess_row_count` is not a one-line honest
fix, it is a product decision about `gpio check`.

- It would flip existing published files from "optimal" to "suboptimal" — the
  pass/fail of `check row-group` is `row_status == "optimal"`.
- It would change the `check optimization` score for files nobody has touched.
- The wide band is arguably right for non-spatially-sorted files, where a query
  is not pruning on a bounding box at all. If so, the fix is a *profile-aware*
  band (`CheckProfile` is already threaded into `assess_row_group_size` but not
  into `assess_row_count`), not a single narrower range.
- `fix_spatial_order` calls `hilbert_order(..., row_group_rows=100000)`
  explicitly, so it now writes a different layout than `gpio sort hilbert` does.
  Dropping that argument would make the fix inherit the sort default; that is a
[...]

Issue #775, opened 2026-08-30
gpio sort defaults to 100k rows per group, against gpio's own 10k-50k guidance for spatial queries


cholmes:
`gpio sort hilbert` (and `sort str`, `sort quadkey`) default to **100,000 rows per
row group**:

```
--row-group-size INTEGER   Exact number of rows per row group (default: 100000 if
                           --row-group-size-mb not set)
```

gpio's own guidance says 10,000–50,000 for exactly this case, in three places:

- `docs/guide/sort.md:102` — "For GeoParquet 2.0 or parquet-geo-only files with
  Hilbert sorting, use **10,000-50,000 rows per group**. Smaller row groups create
  tighter bounding boxes that enable more row group skipping during spatial
  queries. Benchmarks show 10k rows + Hilbert + v2.0 enables ~67% row group
  skipping vs 0% with large row groups."
- `docs/guide/check.md:165` — the same figure.
- `core/check_parquet_structure.py:256` — what `gpio check` *prints as advice*:
  "Spatial queries: 10,000-50,000 rows per group with Hilbert sorting".

So the default a sort command picks is 2–10x the number the same tool recommends
for the workload sorting exists to serve, and `gpio check` will then advise the
user to change what `gpio sort` just chose.

## Measured cost

Re-sorting nine real published catalog files, default vs `--row-group-size 50000`.
"Efficiency" is the estimated row-group skip rate as a fraction of what the file's
row-group count allows (see #755); "% read" is the mean share of the file a query
window covering 10% of each dimension must read.

| file | rows | default: groups / eff | 50k: groups / eff | % read, default -> 50k |
|---|---|---|---|---|
| United_Arab_Emirates | 267,429 | 3 / **0.66** | 6 / **0.87** | 43.6% -> 27.5% |
| Albania | 493,729 | 5 / 0.90 | 10 / **0.95** | 43.0% -> 19.2% |
| Belarus | 451,664 | 4 / 0.93 | 9 / 0.89 | 53.1% -> 21.9% |
| postcode6 | 464,964 | 4 / 0.96 | 10 / 0.93 | 49.0% -> 21.2% |
| brp_gewaspercelen_2020 | 773,139 | 7 / 0.93 | 16 / **0.95** | 53.2% -> 14.8% |
| nwb_wegen | 1,626,321 | 14 / 0.96 | 32 / **0.97** | 100% -> 9.6% |

The efficiency figure is *relative to the row-group count*, so it partly hides the
[...]

PR #958, opened 2026-09-09
fix(check): say which row-count band is a verdict and which is advice, and make unsorted.parquet actually unsorted

### LINKED ISSUE #940: test(fixtures): unsorted.parquet reads as perfectly ordered
Found while writing coverage for #929 (AI fix agent).

`tests/data/unsorted.parquet` reports `ratio 0.0, fix_available False` from the spatial-order check — i.e. the fixture named "unsorted" reads as **perfectly ordered**, so it cannot drive the `check spatial --fix` branch it looks like it exists for. The agent had to build a local fixture instead (places shuffled into 16 row groups) to exercise both the spatial and row-group fix paths.

Pre-existing and unrelated to #929, but worth fixing because the name actively misleads: anyone reaching for a genuinely-unsorted input will pick this file and get a test that passes for the wrong reason. Either regenerate it so it is actually unsorted, or rename it to say what it is.

Related context: #774 (open) reworks how spatial order is judged — worth coordinating, since a regenerated fixture would want to be unsorted under the new pruning-based definition, not just the old consecutive-pair one.

Refs #929.

cholmes:
Two independent defects, one commit each.

### 1. `gpio check row-group` no longer contradicts itself (#795)

The check carried three row-group numbers from three sources. It could print
*"Row count per group is optimal"* for a 150,000-row group and, immediately
under it, advise 10,000-50,000.

The two bands are now named constants in `core/check_parquet_structure.py` with
the reasoning attached, and every message is built from them:

| Constant | Value | Role |
|---|---|---|
| `GENERAL_ROW_COUNT_RANGE` | 10,000-200,000 | the pass/fail **verdict** |
| `SPATIAL_ROW_COUNT_RANGE` | 10,000-50,000 | **advice** for spatial pruning |

- A file inside the general band but above the spatial one is still `optimal`,
  and now says what it is optimal *for*: "Row count per group is optimal for
  general use (10,000-200,000); spatial queries prune more with 10,000-50,000
  rows per group".
- The printed guidelines label each band with the workload it serves and name
  the sort default, so the advice reads as advice instead of as a correction of
  the verdict just printed.
- `docs/guide/check.md` documents both bands and cites the measurement.

**No status or score changes for any existing file** *from the band change*: the verdict band is
unchanged, so `passed`, `row_status` and the `check optimization` score are all
byte-identical to before. Only wording changed.

### 2. `tests/data/unsorted.parquet` is now actually unsorted (#940)

Note the scope of the claim above: it covers section 1. The fixture regenerated
in section 2 *does* change its own status and score, deliberately — 15 groups of
~96 rows reads `poor` from `check row-group` and scores 0/5, where the
single-group version read `optimal` and scored 2/5. It is a fixture for bad
spatial order, not a file that is otherwise healthy; `conftest.py` now says so.

**Regenerated, not renamed** — the rows were never the problem. They are
Overture places in UUID order, which measures indistinguishable from a shuffle
(mean neighbour distance 0.316 in file order vs 0.314 shuffled, 0.028 xy-sorted).
The file was written as a **single row group**, and the spatial-order check
compares row-group bounding boxes: with one group there are no pairs, so it
short-circuits to "well ordered" whatever the rows contain. Renaming would have
described the symptom and thrown away a fixture whose data is exactly right.

Same 1,445 rows, same row order, same schema, same 1.0.0 `geo` metadata; now
15 row groups of 100 instead of one of 1,445. The check goes from
`ratio 0.0, passed True, fix_available False` to
`ratio 1.0, passed False, fix_available True`, so it can finally drive the
`check spatial --fix` branch it appears to exist for.

`tests/data/generate_test_fixtures.py` grows `generate_unsorted()`, which
rebuilds the file deterministically from the committed
`country_partition/El_Salvador.parquet` (its actual provenance: same 1,445 ids,
superset schema), plus a verification pass that fails loudly if the fixture ever
reads as ordered again. There was no generator before — the file arrived as a
binary blob in #123 with no provenance.

### Which band, and on what evidence

Per the repo's own rule, thresholds get derived from benchmarks or first
principles rather than back-filled from what the code happens to do. The three
sources were not equally supported:

- **10,000-50,000** has measurements behind it. #775 re-sorted nine published
  catalog files from 100,000 to 50,000 rows per group and cut the share of the
  file a query window covering 10% of each dimension must read from 43-100% down
  to 10-28%. It is also what `docs/guide/check.md`, `docs/guide/sort.md`, the
  `check optimization` score factor and the `gpio sort` default (50,000, the top
  of the band) already say.
- **10,000-200,000** has no documentation and no benchmark anywhere in the repo.

So the narrow band wins on evidence — but narrowing the *verdict* to it would
have been wrong, and that is the decision worth spelling out. `gpio convert`'s
own general write default lands at 100,352 rows per group (100,000 rounded up to
a multiple of 2,048) and DuckDB writes 122,880. A 10,000-50,000 verdict would
make `gpio check` report `suboptimal` for a file `gpio convert` had just written
at its default, and fail most of the GeoParquet in the world for having been
written by a mainstream writer. That is a worse contradiction than the one being
fixed.

The honest answer is that both numbers are right and they answer different
questions: one is a verdict about layout sanity, the other is advice about
spatial pruning. Neither number moved. What changed is that the check now says
which is which — issue option 3, with the constants shared so the numbers cannot
drift apart again.

**Not in this PR:** `core/check_fixes.py` still hard-codes `row_group_rows=100000`
in five places, and `fix_spatial_order` should probably inherit the sort default
so `check --fix` and `gpio sort hilbert` write the same layout. That changes the
bytes `--fix` produces, which deserves its own PR and its own before/after — it
is out of this one's scope.

### Two tests were passing for the wrong reason

Both are called out because the fixture change makes them assert for real:

- `test_poorly_ordered_file` wrapped every assertion in
  `if result["ratio"] >= 0.3` / `if not result["passed"]`. With `ratio 0.0` and
  `passed True`, neither guard ever fired: the test only checked that the call
  returned a dict with a `"method"` key.
- `test_issues_and_recommendations_when_failed` asserted **nothing at all** —
  its entire body sat under `if not result["passed"]`, and the check always
  passed. It could only fail by raising.

Both now assert unconditionally (`passed is False`, `ratio >= 0.3`,
`fix_available is True`, issues and recommendations non-empty). A third test is
new: the fixture must have at least 3 row groups, which is the property that was
silently missing.

The other two users of the fixture are unaffected and were checked:
`tests/test_convert.py` asserts `row_count == 1445` (unchanged; its docstring's
stale size note was corrected), and `tests/e2e/test_journeys.py` never touches
this file — its `unsorted.parquet` is a 200,000-point temp file it builds itself,
already written with `--row-group-size 10000` for exactly this reason.

## Verification

  → **6475 passed, 75 skipped, 9 xfailed** in 172s.
- Meta lane: `uv run pytest -m meta -q` → **203 passed**, 7289 deselected.
- `uv run pre-commit run --all-files` → clean;
  `uv run pre-commit run --all-files --hook-stage pre-push` → clean, all 27 hooks
  including `menard-check`, `xenon`, `import-linter`, `mypy` and `vulture`.
- Diff coverage: `uv run diff-cover coverage.xml --compare-branch=origin/main
  --fail-under=90` → **100%** (13 changed lines, 0 missing).
- Fixture, before → after:
  `{'method': 'bbox_stats', 'passed': True, 'ratio': 0.0, 'total_pairs': 0, 'fix_available': False}`
  → `{'method': 'bbox_stats', 'passed': False, 'ratio': 1.0, 'total_pairs': 14, 'fix_available': True}`
- Regeneration is reproducible: re-running `generate_unsorted()` on the committed
  source reproduces the fixture's rows, row order, geometry and `geo` metadata
  exactly (verified against the previous file before replacing it).

Under `--cov` instrumentation, 4 subprocess/pipe tests
(`test_pipe_integration.py`, `test_journey_01`) timed out on a machine shared
with other agents; all 4 pass serially and passed in the uninstrumented full run.

Refs #795 (see below)
Closes #940

## What this does *not* close — #795 stays open

Adversarial review found the headline instance of #795 still live, so this is
`Refs` rather than `Closes`. `gpio sort` asks for 50,000 rows per group; DuckDB
rounds the request up to a multiple of its 2,048-row vector size, so **51,200**
lands on disk — outside the 10,000–50,000 band that `gpio check optimization`
scores:

```
$ gpio sort hilbert n102400.parquet n102400_sorted.parquet
$ gpio check optimization n102400_sorted.parquet
Score: 2/5
  [fail] Row Group Size: Average 51,200 rows per group (optimal 10k-50k for spatial queries)
```

That is exactly what #775 predicted and #795 was filed about — gpio advising a
change to what gpio just chose. Fixing it needs a decision (change the sort
request, move the band, or score against the verdict band instead), so it is
**#961**, along with `check_fixes.py`'s five hardcoded `row_group_rows=100000`
that #795 also names.


cholmes:
## Multi-agent adversarial review — one blocker found and fixed in 0fde52e

Worth stating up front: **CodeRabbit hit its OSS rate limit on this PR and posted no review**, unlike #955 and #956 which got real ones. Its green check was not a signal here, so this was the only review the PR got. Reviewed in an isolated worktree, rebased onto current `main` (`4af17e8`) first.

### The blocker: a PR about self-contradiction shipped a self-contradiction

`docs/guide/check.md` line 158 — **added by this PR** — says `gpio sort` writes 51,200. Line 162, four lines below and retained, said the sort commands *"default to 50,000 rows per group, the top of that band, so a freshly sorted file already sits inside it."*

51,200 is not inside 10,000–50,000. I verified the underlying fact rather than the prose:

```
$ gpio sort hilbert n102400.parquet n102400_sorted.parquet
$ python -c "...ParquetFile(...).metadata" -> 51200.0 rows per group
$ gpio check optimization n102400_sorted.parquet
Score: 2/5
  [fail] Row Group Size: Average 51,200 rows per group (optimal 10k-50k for spatial queries)
```

Both the guide sentence and the matching claim on `parquet_writer.py`'s `DEFAULT_SORT_ROW_GROUP_ROWS` now say what actually happens: the sort commands *ask for* 50,000, DuckDB rounds a request up to a multiple of its 2,048-row vector size, and 51,200 lands just outside the band.

### `Closes #795` → `Refs #795`

That reproduction *is* #795's headline symptom — gpio advising a change to what gpio just chose — and it is still live. Merging with `Closes` would have retired the issue with its most visible instance intact. Filed as **#961**, together with `check_fixes.py`'s five hardcoded `row_group_rows=100000` that #795 also names, and the small-file leniency inconsistency (a 56 KB file in 3 row groups is told "Row count per group is very low" directly under a guideline saying small files are fine as one group). Fixing the rounding needs a product decision, so it did not ride along.

### Also fixed: a constant that claimed authority it did not have

`SPATIAL_ROW_COUNT_RANGE`'s docstring named `gpio check optimization` as one of the things it governs — but `check_optimization.py:120` hardcoded `10000 <= avg_rows <= 50000` independently, so that was true only by coincidence. **This is the same drift hazard #948 removed from the kdtree options earlier today.** That function already imported from the same module, so wiring it was free. Verified inert: `passed` and the detail string are byte-identical at 9,999 / 10,000 / 49,999 / 50,000 / 50,001 / 51,200 / 120,000.

`GENERAL_ROW_COUNT_RANGE`'s docstring now admits its upper bound is **inherited, not derived**. The argument in it justifies a bound somewhere above DuckDB's 122,880, not 200,000 specifically; 200,000 is what the check has always used, kept so no published file changes status. Only the spatial band has a measurement behind it and the comment should not read as if both do.

### What held up under attack

The strongest claim in the body survived a hard look. "No status or score changes for any existing file" was checked two ways: a domain sweep of `assess_row_count` over **879 inputs** (every band boundary ±1, a 997-step scan to 250k, each under three call shapes) — **0 status differences, 0 colour differences**, with all 334 message differences confined to the intended `(50,000, 200,000]` window; and a fixture sweep of `check_row_groups` + `check_optimization` over all 25 repo fixtures, where the only file that differs is the one deliberately regenerated.

The fixture regeneration is exactly as described: 1,445 rows, `table.equals()` **True**, ids identical pairwise in file order, schema identical, `geo` metadata byte-identical; only `created_by`, format version, row-group count and size differ. `generate_unsorted()` reproduces the committed file **byte-identically** (sha256 `6c8fe094…`), so the provenance claim is real rather than decorative.

And the two rewritten tests **strengthen** rather than paper over: `test_poorly_ordered_file` went from `if ratio >= 0.3: assert ...` — a dead guard that passed whatever the fixture contained, which is *how* a fixture named "unsorted" reported ratio 0.0 — to unconditional assertions.

### Scoping corrections to the body

The "no status or score changes" claim is true for section 1 but not at PR scope: the regenerated fixture moves from `optimal`/2-of-5 to `poor`/0-of-5. That is intended, but it means the file is no longer a "one bad property" fixture. Body and the `unsorted_test_file` docstring now say so, and the title now covers both halves.

### Gates

Fast suite **6722 passed, 75 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100% on 13 changed lines** · `pytest -m meta` 205 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push (xenon, mypy, vulture, menard-check) all clean · `cz check` passes on the new title.

Reviews performed by Claude Code agents; every finding independently reproduced before acting on it.

Issue #680, opened 2026-08-21
gpio add quadkey fails on non-identifier --quadkey-name values


cholmes:
`_add_quadkey_file_based` builds its SQL with raw string interpolation instead of `quote_identifier()`, so `gpio add quadkey` fails on any column name that isn't a bare SQL identifier.

## Affected code

`geoparquet_io/core/add/quadkey.py`:

- Line 640 interpolates the output column name directly:

  ```python
  lat_lon_to_quadkey({lat_expr}, {lon_expr}, {resolution}) AS {quadkey_column_name}
  ```

- Lines 630-631 interpolate the bbox column name directly:

  ```python
  lat_expr = f"(({bbox_col}.ymin + {bbox_col}.ymax) / 2.0)"
  lon_expr = f"(({bbox_col}.xmin + {bbox_col}.xmax) / 2.0)"
  ```

This is the only unquoted `AS` clause in the module. The sibling copies of the same query — lines 298, 304, and 484 — all wrap the name in `quote_identifier()`.

## Steps to reproduce

```bash
gpio add quadkey input.parquet output.parquet --quadkey-name 'weird name'
```

Result:

```
ParserException: Parser Error: syntax error at or near "name"
```

A name containing a double quote fails differently:

```bash
gpio add quadkey input.parquet output.parquet --quadkey-name 'has"quote'
# ParserException: Parser Error: unterminated quoted identifier at or near ""quote
```

Both reproduce through the Python API as well:

```python
add_quadkey_column(src, out, quadkey_column_name="weird name", resolution=12)
```

## Expected behavior

`--quadkey-name` accepts any string that Parquet accepts as a column name, matching the behavior of the streaming and table code paths in the same module.

## Suggested fix

Wrap both names in `quote_identifier()` from `core/duckdb_utils.py`, as the sibling call sites already do. Pass the raw value — escaping is not idempotent.

`CLAUDE.md` documents this rule under **DuckDB Patterns**: never hand-roll SQL escaping; `quote_identifier()` is for identifiers. Column names arrive from user flags and from a file's own `geo.primary_column`, so this is an injection surface rather than a style preference.

PR #662 set the pattern for this class of fix.

## Notes

The `bbox_col` name comes from a fixed candidate list (`bbox`, `bounds`, `bounding_box`), so today it can't carry a hostile value. Quote it anyway to keep the call sites uniform and to keep the fix from being partial.

[...]

PR #959, opened 2026-09-10
fix(cli): guard empty --*-name values and let 'check spatial --fix' work in place

### LINKED ISSUE #933: fix(cli): an empty --*-name option raises an unhandled traceback
Found during the #925 review (AI review agent; both behaviors verified).

`gpio partition kdtree ... --kdtree-name ''` produces an unhandled traceback:
- before #925: `_duckdb.ParserException: syntax error at or near "iter_2"`
- after #925: `ValueError: cannot quote an empty SQL identifier` (from `duckdb_utils.py:234`)

Strictly lateral — not a regression introduced by #925 — but the post-#925 failure now happens in a path with a *nameable* error, making it a one-liner to convert into a clean `InvalidParameterError`/`ClickException` with a message naming the offending option.

Worth doing as a shared guard rather than a kdtree-only fix: since #680, every `--*-name` option (`--bbox-name`, `--h3-name`, `--a5-name`, `--s2-name`, `--quadkey-name`, `--kdtree-name`, `--column`) has the same shape, and an empty or whitespace-only value is meaningless for all of them. A small validator (a Click callback, or a check inside `quote_identifier`'s callers) would cover the family at once.

Refs #925, #680.

### LINKED ISSUE #941: fix(check): 'check spatial --fix' cannot fix in place, unlike its sibling checks
Found while writing coverage for #929 (AI fix agent); real behavior, verified, not touched by that PR.

`gpio check spatial --fix` without `--fix-output` refuses with "Cannot overwrite input file" from `handle_fix_common`, because the command exposes no `--overwrite` option — while `check row-group`, `check bbox` and `check compression` all do and can therefore fix in place (with a backup).

So the four fix-capable check subcommands are inconsistent: three support in-place repair, one silently cannot, and the error message doesn't explain that the option is missing rather than the operation being unsafe. A user following the pattern from the other three hits a wall.

Either add `--overwrite` to `check spatial --fix` for parity, or make the error name the actual constraint ("check spatial cannot fix in place; pass --fix-output <path>").

Note #929's coverage tests work around this by using `--fix-output`, with the in-place backup message pinned via the row-group test instead — so whichever way this is resolved, those tests already cover both shapes.

Refs #929.

cholmes:
## What changed

**#933 — an empty `--*-name` value is now a usage error, not a traceback.**
`cli/decorators.py` gains `_validate_column_name`, a Click callback, and
`column_name_option()`, a thin `click.option` wrapper that wires it up. Twelve
option declarations switch from `@click.option` to `@column_name_option`:
`--bbox-name`, `--h3-name`, `--a5-name`, `--s2-name`, `--kdtree-name`,
`--quadkey-name` (across `gpio add`, `gpio partition` and `gpio sort`) plus
`gpio partition string --column`. No default, help text or type changes.

**#941 — `gpio check spatial --fix` can repair a file in place.**
`fix_spatial_ordering` now routes an in-place rewrite through a temp file and
moves it back, and `check spatial` gains `--overwrite` through the same
`overwrite_option` decorator its three fix-capable siblings already use.

## Why

**#933:** since #680 every `--*-name` option has had the same exposure. The
value ends up as a delimited SQL identifier, so `gpio add bbox in.parquet
out.parquet --bbox-name ''` read the whole input file and *then* died with a raw
`ValueError: cannot quote an empty SQL identifier` traceback (exit 1), while
`--bbox-name '   '` was accepted outright and wrote a column literally named
`'   '`.

The option definitions turned out **not** to be shared — each command declares
its own `@click.option` with its own default and help — so there was no single
existing decorator to hang a callback on. The next-best choke point is one
shared callback plus a named wrapper that every site now uses, so the rule lives
in one place. `tests/test_cli_column_name_guard.py` walks the live command tree
and fails if a new `--*-name` option appears without the guard, which is what
makes it free for the next one.

**Usage error (exit 2), not a domain error (exit 1)**, following what the repo
already does with bad option values: `_validate_write_memory` in the same module
raises `click.BadParameter` for a malformed `--write-memory`, and every
`InvalidParameterError` from core (`sort quadkey --quadkey-name nope`,
`partition string --column nope`) is mapped to exit 2 by
`handle_core_exception`. `click.BadParameter` also names the offending flag
itself, so the message doesn't have to repeat it.

**#941: `--overwrite` was not omitted for a good reason, but it was also not the
cause.** `handle_fix_common` only consults `overwrite` for *remote* files; a
local in-place fix needs no flag at all (it just writes a `.bak`). The real
blocker is that `fix_spatial_ordering` was the only `fix_*` function that handed
the input path straight through as `hilbert_order`'s output. `hilbert_order`
reads the whole input while writing, so `handle_output_overwrite` refuses an
output that resolves to its own input — and `overwrite=True`, which
`fix_spatial_ordering` already passed, does not lift that check. Worse, the
refusal landed *after* `handle_fix_common` had written the `.bak`, leaving a
stray backup behind.

`fix_compression` (`tempfile.mkstemp`) and `fix_bbox_all` (`output + ".tmp"`)
already route in-place rewrites through a temp file, which is why
`check compression`, `check bbox`, `check row-group` and even `check all --fix`
(via `_apply_spatial_ordering_fix`) all work in place. `fix_spatial_ordering`
simply never got that treatment. So: a plain oversight, fixed the same way —
and `--overwrite` added on top, since remote in-place fixes genuinely did need
it for parity.

## Verification

Before, on a shuffled copy of `places_test.parquet`:

```
$ gpio partition kdtree in.parquet out/ --kdtree-name '' --partitions 2
Processing 766 features with 2 partitions ...
Error: Failed to add KD-tree column: cannot quote an empty SQL identifier ...

$ gpio add bbox in.parquet out.parquet --bbox-name ''
Processing 766 features...
ValueError: cannot quote an empty SQL identifier          # exit 1, traceback

$ gpio add bbox in.parquet out.parquet --bbox-name '   '
Successfully added bbox column '   ' to: out.parquet      # exit 0

$ gpio check spatial poor.parquet --fix
Applying Hilbert spatial ordering...
✓ Created backup: poor.parquet.bak
Error: Cannot overwrite input file: poor.parquet          # exit 1, .bak orphaned
```

After:

```
$ gpio add bbox in.parquet out.parquet --bbox-name ''
Error: Invalid value for '--bbox-name': column name cannot be empty or
whitespace-only                                           # exit 2, nothing read

$ gpio add bbox in.parquet out.parquet --bbox-name 'my bbox'
Successfully added bbox column 'my bbox' to: out.parquet  # non-blank still fine

$ gpio check spatial poor.parquet --fix
✓ Spatial ordering applied successfully!
Optimized file: poor.parquet
Backup: poor.parquet.bak                                  # exit 0
```

- Fast suite: `pytest -n auto -m "not slow and not network and not meta"` —
  **6761 passed, 75 skipped, 9 xfailed**. (A first `-n auto` run reported 7
  failures in `test_journeys` / `test_pipe_integration` /
  `test_geometry_repair_integration`; all 40 pass serially — machine saturation,
  not this change.)
- Meta lane: `pytest -m meta` — **205 passed**.
- `pre-commit run --all-files` clean; `--hook-stage pre-push` clean (xenon,
  import-linter, deptry, vulture, mypy, menard-check, fast-tests).
  **43 changed lines, 0 missing, 100%**.

### `tests/data/cli_surface.json`

Re-recorded with `GPIO_UPDATE_SNAPSHOT=1 uv run pytest tests/test_cli_surface.py`.
The diff is **16 added lines, 0 removed** — one option object appended to the
`params` list of `check.spatial` (the block starting at `"spatial": {` on line
3328):

```json
+      {
+       "default": "False",          // flag defaults to off
+       "hidden": false,             // shown in --help
+       "is_flag": true,             // boolean flag, no value
+       "metavar": null,             // no declared placeholder
+       "multiple": false,           // not repeatable
+       "name": "overwrite",         // the new parameter
+       "nargs": 1,
+       "opts": [
+        "--overwrite"               // the only spelling
+       ],
+       "param_type": "Option",
+       "required": false,
+       "secondary_opts": [],        // no --no-overwrite counterpart
+       "type": "boolean"
+      },
```

Every field is identical to the `overwrite` entry the three sibling check
subcommands already record, because it comes from the same `overwrite_option`
decorator. Nothing else in the file changed: #933 only adds a `callback=`, which
the snapshot does not record.

Closes #933
Closes #941

## Summary by CodeRabbit

* **New Features**
  * Added an `--overwrite` option to `check spatial` for controlling replacement of existing outputs during fixes.
  * Spatial fixes can now update input files in place while preserving backups.
  * Multi-file fixes now report rewritten files and backup locations.

* **Bug Fixes**
  * Column-name options reject empty or whitespace-only values with a clear command-line error.
  * In-place spatial repairs reliably handle local and remote outputs, including temporary-file cleanup.

### DIFF
--- a/docs/guide/check.md
+++ b/docs/guide/check.md
@@ -418,8 +418,26 @@ gpio check all partitions/ --all-files
 gpio check all partitions/ --sample-files 3
 ```
 
-!!! note "--fix not available for partitions"
-    The `--fix` option only works with single files. To fix issues in partitioned data, first consolidate with `gpio extract`, apply fixes, then re-partition if needed.
+!!! warning "--fix rewrites every file it checks"
+    `--fix` applies to whichever files the check ran on. With `--all-files` or
+    `--sample-files N` on a directory that means **every file it checked is
+    rewritten in place**, not just the first one:
+
+    <!-- doctest: skip="needs partitions/, which the harness does not seed" -->
+    ```bash
+    # Rewrites all 4 files in the partition, in place
+    gpio check spatial partitions/ --fix --all-files
+    ```
+
+    Each rewritten file gets a `.bak` alongside it (unless you pass
+    `--no-backup`), and the run ends with a `Fixed N files:` list naming what it
+    wrote. There is no confirmation prompt, so check the file count in the
+    `📁 Checking ...` notice before adding `--fix`.
+
+    To write the fixed files somewhere else instead, `gpio check all` takes a
+    directory as `--fix-output` (it refuses a single file path when more than
+    one file is being fixed). Alternatively, consolidate the partition with
+    `gpio extract` first, fix that, then re-partition.
 
 ## See Also
 
--- a/geoparquet_io/cli/commands/add.py
+++ b/geoparquet_io/cli/commands/add.py
@@ -15,6 +15,7 @@ from geoparquet_io.cli.decorators import (
     SingleFileCommand,
     any_extension_option,
     bbox_option,
+    column_name_option,
     dry_run_option,
     geoparquet_version_option,
     output_format_options,
@@ -370,7 +371,7 @@ def add_geometry_metrics_cmd(
 @add.command(name="bbox", cls=SingleFileCommand)
 @click.argument("input_parquet")
 @click.argument("output_parquet", required=False, default=None)
-@click.option("--bbox-name", default="bbox", help="Name for the bbox column (default: bbox)")
+@column_name_option("--bbox-name", default="bbox", help="Name for the bbox column (default: bbox)")
 @click.option(
     "--force",
     is_flag=True,
@@ -498,7 +499,9 @@ def add_bbox_metadata_cmd(ctx, parquet_file, verbose):
 @add.command(name="h3", cls=SingleFileCommand)
 @click.argument("input_parquet")
 @click.argument("output_parquet", required=False, default=None)
-@click.option("--h3-name", default="h3_cell", help="Name for the H3 column (default: h3_cell)")
+@column_name_option(
+    "--h3-name", default="h3_cell", help="Name for the H3 column (default: h3_cell)"
+)
 @click.option(
     "--resolution",
     default=9,
@@ -571,7 +574,9 @@ def add_h3(
 @add.command(name="a5", cls=SingleFileCommand)
 @click.argument("input_parquet")
 @click.argument("output_parquet", required=False, default=None)
-@click.option("--a5-name", default="a5_cell", help="Name for the A5 column (default: a5_cell)")
+@column_name_option(
+    "--a5-name", default="a5_cell", help="Name for the A5 column (default: a5_cell)"
+)
 @click.option(
     "--resolution",
     default=15,
@@ -644,7 +649,9 @@ def add_a5(
 @add.command(name="s2", cls=SingleFileCommand)
 @click.argument("input_parquet")
 @click.argument("output_parquet", required=False, default=None)
-@click.option("--s2-name", default="s2_cell", help="Name for the S2 column (default: s2_cell)")
+@column_name_option(
+    "--s2-name", default="s2_cell", help="Name for the S2 column (default: s2_cell)"
+)
 @click.option(
     "--level",
     default=13,
@@ -717,7 +724,7 @@ def add_s2(
 @add.command(name="kdtree", cls=SingleFileCommand)
 @click.argument("input_parquet")
 @click.argument("output_parquet")
-@click.option(
+@column_name_option(
     "--kdtree-name",
     default="kdtree_cell",
     help="Name for the KD-tree column (default: kdtree_cell)",
@@ -830,7 +837,7 @@ def add_kdtree(
 @add.command(name="quadkey", cls=SingleFileCommand)
 @click.argument("input_parquet")
 @click.argument("output_parquet", required=False, default=None)
-@click.option(
+@column_name_option(
     "--quadkey-name",
     default="quadkey",
     help="Name for the quadkey column (default: quadkey)",
--- a/geoparquet_io/cli/commands/check.py
+++ b/geoparquet_io/cli/commands/check.py
@@ -58,6 +58,7 @@ class MultiFileCheckRunner:
         self.warnings = 0
         self.failed = 0
         self.issues: list[tuple[str, str, str]] = []  # (file, level, message)
+        self.fixed: list[tuple[str, str | None]] = []  # (output path, backup path)
         self.current_index = 0
         self.is_multi_file = len(files) > 1
 
@@ -118,6 +119,30 @@ class MultiFileCheckRunner:
                     self._record_issue(file_path, "warning", issue)
         self._update_progress()
 
+    def record_fix(self, output_path: str, backup_path: str | None):
+        """Record a file this run rewrote, for the end-of-run summary."""
+        self.fixed.append((output_path, backup_path))
+
+    def _print_fix_summary(self):
+        """List what ``--fix`` rewrote.
+
+        In multi-file mode the per-file "Optimized file:" lines are suppressed,
+        so a run that rewrote every file in a partition in place used to end on
+        a summary that mentioned only the checks. Naming the files is the
+        minimum a mass in-place rewrite owes the user.
+        """
+        if not self.fixed:
+            return
+
+        plural = "" if len(self.fixed) == 1 else "s"
+        click.echo(click.style(f"Fixed {len(self.fixed)} file{plural}:", fg="green"))
+        for output_path, backup_path in self.fixed[: self.max_issues_shown]:
+            suffix = f" (backup: {backup_path})" if backup_path else ""
+            click.echo(f"  - {output_path}{suffix}")
+        extra_fixes = len(self.fixed) - self.max_issues_shown
+        if extra_fixes > 0:
+            click.echo(click.style(f"  ... and {extra_fixes} more", fg="cyan"))
+
     def print_summary(self):
         """Print final summary after all files are checked."""
         if not self.is_multi_file:
@@ -147,6 +172,7 @@ class MultiFileCheckRunner:
 
         summary = ", ".join(summary_parts) if summary_parts else "0 checked"
         click.echo(f"Summary: {summary} ({total} files checked)")
+        self._print_fix_summary()
 
 
 @check.command(name="all", cls=GlobAwareCommand)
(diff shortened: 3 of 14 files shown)


cholmes:
## Multi-agent adversarial review — three S2s found and fixed in 81d9562

Reviewed in an isolated worktree, rebased onto current `main`. **No CodeRabbit review exists on this PR**, so this was the only review it got. Every finding below I reproduced myself before acting on it.

### S2 — the new in-place path destroyed the user's data on a move failure

`_move_temp_output_into_place` unlinked the destination *then* moved, and the caller's `finally` deleted the temp unconditionally. Injecting `OSError(28, "No space left on device")` on the move:

```
before:   original survived : False
          temp survived     : False      <- both copies gone
after :   original survived : True  (136,488 bytes, unchanged)
```

Realistic triggers are exactly the conditions under which someone in-place-fixes a large file: ENOSPC, quota, read-only mount, EXDEV. `.bak` covers the CLI default path, but under `--fix --no-backup`, or for any direct caller of the core function, it was total unrecoverable loss. Notably this was *worse* than `fix_compression`, the function the PR cites as its model — that one leaks the temp, which is ugly but recoverable.

Fixed structurally rather than patched: the temp is now staged **beside the target** (`mkstemp(dir=dirname(output))`, dot-prefixed so it stays out of `*.parquet` globs) and moved with `os.replace()`, with no prior unlink. That is atomic on POSIX and Windows, and it closes three things at once — the data loss, the cross-filesystem case where `$TMPDIR` is a different mount (a multi-GB rewrite staging into a tmpfs), and the Windows `WinError 32` hazard permanently.

### S2 — `Closes #933` was overstated: the same bug reproduced one command away

The guard was applied by matching the flag *spelling* (`--*-name`, `--column`) rather than by asking which values become SQL identifiers:

```
gpio add bbox      --bbox-name ''        -> exit 2, clean usage error      (guarded)
gpio sort hilbert  --geometry-column ''  -> exit 1, full Python traceback  (not)
gpio partition quadkey --quadkey-column '   ' -> exit 0, creates a column named '   '
```

Both halves of #933 — the raw traceback and the silently-accepted whitespace name — survived. The sweep was redone **by data flow**, tracing every CLI option to `quote_identifier()`. Three more options were reachable and are now guarded (`sort hilbert -g`, `sort str -g`, `partition quadkey --quadkey-column`); all three now exit 2 with no traceback. Just as usefully, the options deliberately *left* are now recorded with their reasons in `REVIEWED_UNGUARDED` — those already validated against the file schema before quoting, and those that are string literals rather than identifiers. `Closes #933` is now accurate.

### S2 — the ratchet test had a hole, so "free for the next one" was not true

`_column_name_options()` keyed by command path and **overwrote**, so only one matching option per command reached the comparison. Whether the ratchet fired depended on decorator ordering, not on whether an option was guarded. I sabotage-verified the fix in the ordering that was the actual hole — a bare `@click.option("--zzz-name", ...)` inserted *above* a guarded option:

```
before fix: 2 passed          <- sabotage invisible
after  fix: 1 failed          <- caught
```

It now accumulates, matches `--*-column` as well as `--*-name`, and skips `is_flag` options.

### Two things the fix found beyond the brief

- **`handle_fix_common` carried the same string-comparison bug**, so an aliased `--fix-output` (a `./` prefix, a relative/absolute mix, a symlink) took the "different file" branch — meaning **no `.bak` and no `--no-backup` confirmation**, which is precisely the mitigation the data-loss fix relies on. The comparison now lives in one `is_same_file_path()` helper used at all four decision points, including `handle_output_overwrite`, so they cannot drift apart again.
- **`extract bigquery --include-cols/--exclude-cols` quote unvalidated names** — that path has no `validate_columns` equivalent at all. Correctly left out: it is a missing-schema-validation defect on a comma-separated list, not a blank-single-name defect, and needs a live BigQuery schema. Filed as **#969**.

### Also addressed

`docs/guide/check.md` said "`--fix` only works with single files", which the in-place path makes wrong — `check spatial <dir> --fix --all-files` now rewrites a whole partition. Doc corrected, and a per-run summary added (multi-file mode suppressed the per-file lines, so a run that rewrote three files used to end as if nothing had happened). A confirmation prompt was considered and declined with a reason I agree with: the looping behaviour predates this PR across four fix-capable checks, and `.bak` is on by default with `--no-backup` already prompting.

I flagged the Windows open-handle trap as the highest risk. **It does not reproduce** — `lsof` at the moment of the move showed zero open parquet handles, on both the direct and full CLI paths. The reviewer did find the latent version: `_hilbert_order_file_based` never closes its DuckDB connection on the success path, so the invariant holds only by CPython refcounting. The `os.replace` fix makes that moot, but it is worth knowing.

### Not fixed here

`fix_spatial_ordering` hardcodes `row_group_rows=100000`, so an in-place fix moves a file out of the 10k–50k band and `check optimization` marks it `[fail]`. Pre-existing (`--fix-output` does the same) and tracked in **#961**, which #967 now addresses.

### Gates

Fast suite **6828 passed, 76 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100% on 89 changed lines** · `pytest -m meta` **205 passed** (a parallel run showed 2 subprocess flakes under load; clean serially) · docs lane 21 passed, 3 skipped · `lint-imports` 4 kept / 0 broken · `ruff` and pre-commit clean · +29 test cases. `cli_surface.json` moved by exactly one hunk, the intended `--overwrite` on `check spatial`; the guard widening changed no help text, default, type or flag name.

cholmes:
### CI follow-up: all three Windows jobs failed, now fixed in 5decf25

```
tests/test_file_utils.py::TestIsSameFilePath::test_relative_and_absolute_spellings
ValueError: path is on mount 'C:', start on mount 'D:'
```

GitHub's Windows runners put pytest's `tmp_path` on `D:` and the process CWD on `C:`, and `os.path.relpath` with no `start` raises across drives. Invisible on macOS and Linux, where both live under one root.

**The production helper is fine** — `Path.resolve()` does not raise across drives — so this is the test, not the code under test.

There is an irony worth recording: the repo walks into this by design. `no-cwd-change-in-tests` forbids `chdir` and its error message explicitly points at `os.path.relpath` as the alternative, which is exactly what breaks here. Anyone writing relative-path coverage against `tmp_path` will hit it again.

A relative spelling can only name the same file when it is reachable from the CWD, and with `chdir` unavailable there is nothing meaningful to assert on a different drive — so the case is skipped with that reason rather than reworked into a weaker assertion that would still pass while testing less. I verified the skip fires by simulating the cross-drive `ValueError`, rather than assuming it.

`is_same_file_path` now also answers `False` on `ValueError` (a malformed drive spec or embedded NUL on Windows). Callers use it to decide whether they are about to overwrite their own input, so "I cannot tell" must take the not-in-place branch — the same safe direction the rest of this PR takes.

Note this is *not* the Windows open-handle hazard I flagged in the review brief. That one genuinely does not reproduce: `lsof` at the moment of the move showed zero open parquet handles, and the `os.replace` fix retires the concern permanently regardless.

Gates on the fix: fast suite **6828 passed, 76 skipped, 9 xfailed, 0 failed** · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit clean.

Issue #952, opened 2026-09-09
fix(metadata): a secondary geometry column's emptied geometry_types is never recomputed on a file path


cholmes:
## What

When a merge or partition write cannot carry a secondary geometry column's stats, gpio writes the spec's "not known" sentinel `geometry_types: []` rather than deleting the REQUIRED key (#934, #949). That keeps the output readable, which is the point — but the `[]` is then **sticky**: no gpio *file → file* command ever recomputes it.

`duckdb_kv._compute_missing_metadata` gates the recompute on `"geometry_types" not in col_meta`, and `[]` is present, so the gap is invisible to it. Verified: `gpio convert geoparquet`, `gpio add bbox` and `gpio extract` all preserve `[]` unchanged.

Only the Arrow/stdout path recovers it — `gpio extract geoparquet in.parquet -` correctly yields the real `["Polygon"]` and a tight bbox, because `backfill_derived_stats` recomputes every column from the rows it writes and reads `[]` as the gap it is.

## Why it matters

`[]` is spec-valid and strictly better than the unreadable output it replaced, so this is not a correctness bug. But a user whose partitioned output carries `boundary: []` has no documented file-based way to restore the real type list, and the asymmetry between the two write paths is not obvious.

## Options

1. Teach `_compute_missing_metadata` to treat an empty `geometry_types` list as a gap, the way `backfill_derived_stats` already does. Needs care: `[]` is also the legitimate answer for a genuinely empty result, so an empty output must not acquire a spurious recompute pass.
2. Leave the behaviour and document the round trip through `gpio extract … -` in the guide as the way to restore precise stats.

Option 1 is the real fix, and would make the two write paths agree.

Found by an adversarial review of #949.

cholmes:
## Correction: this is narrower than I filed it

I wrote that "no gpio *file → file* command ever recomputes `[]`". That is **wrong for the primary geometry column** and right only for secondary ones. Found during the adversarial review of #965; I reproduced it before correcting.

Two-geometry file with **both** columns emptied to `[]`, run through `gpio convert geoparquet` on `68d5f38`:

```
before:  geometry=[]        boundary=[]
after :  geometry=['Point']  boundary=[]
```

[...]

PR #960, opened 2026-09-10
fix(common,stac,stream): stop the last malformed-'geo' readers crashing the write

### LINKED ISSUE #947: fix: remaining unguarded raw geo-block readers outside #883/#887's fixed sites
Third instalment of the same class. #883 introduced `sanitize_geo_metadata()` and fixed four write-path readers; #887 (PR #945) fixed six more in `crs_utils`, `convert` and `geometry_detection`. This is the remainder, found by #945's sweep and left alone because each is larger than a one-liner and sat in files other agents held at the time.

A carried `geo` block is arbitrary JSON from someone else's file. Indexing it without a shape check turns a malformed input into a `TypeError`/`AttributeError` from three frames down instead of a warning and a sane fallback.

## Sites

| File | Lines |
|---|---|
| `core/common.py` | 2774-2794, 3435 |
| `core/add/bbox_metadata.py` | 222-253, 338-344 |
| `core/add/bbox.py` | 210-216 |
| `core/reproject.py` | 76-124, 289-295, 383, 825-831 |
| `core/stac.py` | 216-219 |
| `core/stream_io.py` | 281-286 |
| `api/table.py` | 2558 |

Spot-checked two before filing. `common.py:2780` reads `(geo_dict.get("columns") or {}).get(geometry_column)` — the `or {}` handles `null` but not a *list* `columns`, where `.get` raises `AttributeError`. `reproject.py:289-295` parses the block and hands it to `apply_target_crs_to_geo_meta` / `_drop_nonplanar_edges_from_geo_meta` with no shape check.

## The line to respect

#883 established, and #887 upheld, that **read-only/validation readers do not sanitize** — `gpio check` must see the file as it really is. `parse_geo_metadata`, `crs_utils.parse_geo_metadata_from_schema` and `duckdb_metadata.get_geo_metadata` are deliberately untouched, and #945 classed `geometry_detection` as validation-shared (`check spatial` and `check row-group` call it) so it only refuses a `primary_column` that cannot be a column name rather than sanitizing.

Whoever takes this should classify each site the same way before fixing it, and say which category it fell in. `stac.py` and `api/table.py` in particular need that judgement — neither is obviously a write path.

## Also left open by #887

A `primary_column` naming a column that does not exist still surfaces as a raw `BinderException` from DuckDB in `sort` and `add`, rather than a message naming the metadata as the cause.

Refs #883, #887, #945.

*Filed by Claude (AI) from PR #945's sweep; two sites spot-verified.*

cholmes:
## What changed

The third instalment of #883's class: the eight readers of a carried `geo`
block that indexed it without a shape check. Each site is classified the way
#883 and #887 classified theirs, and the classification decides the fix.

**Write path** — the block becomes an output file, a transform or a published
artifact, so it goes through the shared `sanitize_geo_metadata`:

| Site | Traceback on `origin/main` |
|---|---|
| `common._geo_block_to_carry_on_fast_path` | `AttributeError: 'list' object has no attribute 'get'` (and `'str'`) at `common.py:2786` |
| `common.write_geoparquet_table` | `TypeError: argument of type 'NoneType' is not iterable` at `common.py:3275`; `list indices must be integers`, `string indices must be integers` and `'str' object has no attribute 'get'` at `:3276` |
| `add_bbox_metadata` (file) | `TypeError: 'str' object does not support item assignment` at `bbox_metadata.py:344` |
| `add_bbox_metadata_table` | no crash, but a wrong-typed `crs` was carried into the returned table |
| `stac._add_projection_properties` | `TypeError: argument of type 'NoneType' is not iterable` at `stac.py:216`; `list/string indices must be integers` at `:219` |
| `stream_io._extract_crs_from_metadata` | `TypeError: argument of type 'NoneType' is not iterable` at `stream_io.py:290`; `list/string indices…` and `'str' object has no attribute 'get'` at `:291` |

`_geo_block_to_carry_on_fast_path` is the strongest case in the set: whatever it
returns is written to the output *verbatim*, so a carried `"crs": 42` reached the
file. `stac` is the judgement call the issue flagged — it reads metadata and
writes no Parquet, but the CRS and geometry types it lifts leave gpio inside a
**published** STAC Item, which is what makes it a write path.

**Validation-shared** — `gpio check bbox` and `check all` report through it, so it
must keep seeing the file as it really is and guards instead of sanitizing:

- `common._check_bbox_metadata_covering` — `AttributeError` from
  `columns.items()` at `common.py:3460`, reached by `gpio check bbox` *and* by
  `add bbox-metadata`, which is how the file path above crashed early. A
  `columns` that is not an object, or a `covering` that is not one, declares no
  covering; that is the truthful answer, not a crash.

**Read-only** — the other judgement call:

- `api.Table.metadata` — `AttributeError: 'list' object has no attribute 'get'`
  / `TypeError` at `table.py:2565-2567`. It is the Python API's
  `gpio inspect meta`, and its `geo_metadata` key is documented as *the file's
  full `geo` block*, so the block comes back exactly as the file holds it and
  only the derived keys beside it are guarded. Asserted by the tests: the block
  round-trips unchanged for every malformed shape.

**Already fine, no change:**

- `add/bbox.py:210-227` (`_passthrough_version`) type-checks both the block and
  the `version` string it reads.
- `reproject.py` 76-124, 289-295 and 825-831 were fixed by #945. Its one
  remaining raw read (`_sanitize_null_crs_file`) sits behind
  `geoparquet_crs_is_null`, which #945 sanitized, so no malformed shape reaches
  it — verified rather than assumed.

**Shared, not duplicated.** The three steps every one of these readers repeated —
find the key, decode it, check the shape — become `sanitized_carried_geo`, and
#945's "the named primary column, else the schema, never the literal
`geometry`" recovery becomes `carried_geometry_column`. `reproject`'s two local
helpers now delegate to both. `stream_io` also reads the CRS for the column it
is about to attach it to rather than for whatever `primary_column` survived
sanitizing — the silent half of #887 at this site.

## Why

A carried `geo` block is arbitrary JSON written by some other tool. Indexing it
without a shape check turns a malformed *input* into a Python type error several
frames from anything the user can act on. #883 fixed four such readers and
#887/#945 six more; these are the remainder, all pre-existing on `main` and all
reproduced before being fixed.

The line both predecessors upheld and this one upholds: **read-only and
validation readers do not sanitize**, because `gpio check` has to see the file as
it really is. `parse_geo_metadata`, `crs_utils.parse_geo_metadata_from_schema`
and `duckdb_metadata.get_geo_metadata` stay untouched.

## Verification

Reproduced every site on `origin/main` first with the shapes #883's tests already
use (`columns` null / list / string, an entry that is not an object, a non-string
`primary_column`), then re-ran the same script after the fix: no crashes, and the
only remaining failures are DuckDB's own reader refusing an unreadable file
(`Invalid Input Error: Geoparquet column 'geometry' is not an object`) — a domain
error naming the real cause, which is outside gpio's reach and is exactly what
#945 documented.

- `tests/test_malformed_geo_metadata.py`: +18 tests (329 lines), parametrized
  over both `geometry` and `geom` column names as #945 established, plus a
  positive test per site so the guards cannot cost a well-formed file its
  covering, its CRS, its `proj:*` properties or its derived metadata keys.
- `uv run pytest -n auto -m "not slow and not network and not meta" -q` —
  **6849 passed, 75 skipped, 9 xfailed**.
- `uv run pytest -m meta -q` — 205 passed (three `test_commitizen.py`
  subprocess timeouts on the first parallel run; all 29 pass serially — machine
  saturation, not the change).
- `uv run pre-commit run --all-files` clean; `--hook-stage pre-push` clean,
  `menard-check` included.
  **66 changed lines, 0 missing, 100%**.

One incidental finding worth recording: a fixture written by `pq.write_table`
carries an `ARROW:schema` key holding the *original* metadata, and pyarrow
prefers it over the Parquet key-value block on the way back in — so a test that
checks what `add bbox-metadata` wrote has to read the KV block, not
`schema_arrow.metadata`. Noted in the test.

**Not fixed, and deliberately so.** The issue's second item — a `primary_column`
naming a column the file does not have surfacing as a raw `BinderException` from
`sort` and `add` — is confirmed (`Binder Error: Referenced column "nope" not
found in FROM clause!` from `hilbert_order.py:520` and from `duckdb_kv.py:737`),
but it is not a one-liner. Naming the metadata as the cause means a schema probe
inside `find_primary_geometry_column`, which `check spatial` and `check
row-group` also call — new I/O in a validation-shared reader, on a file `#945`
drew a careful line around. It wants its own issue.

Closes #947

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved resilience when reading, writing, reprojecting, and validating files with malformed GeoParquet metadata.
  * Prevented crashes caused by invalid geometry columns, bounding-box information, CRS values, or other metadata structures.
  * Ensured CRS information is selected from the relevant geometry column.
  * Preserved raw metadata where appropriate while omitting invalid derived values.

* **Tests**
  * Added coverage for malformed and well-formed metadata across table, streaming, bounding-box, CRS, and projection workflows.

---

## Correction: this is not the last instalment

An adversarial review found **four more live members of the same class**, all
reproducing on `main` and all untouched by this PR, so none of them block it.
They are filed as **#968**. The lead one is higher severity than anything the
three instalments have fixed:

- `geo_metadata.py:826` (`_prune_geo_dict_to_columns`) reads `primary_column`
  raw, so a non-string value makes it drop the **entire `geo` block, CRS
  included**. An absent `crs` is spec-defined as OGC:CRS84, so `gpio extract`
  and `gpio sort hilbert` silently relabel a projected file as lon/lat.
  Measured: an EPSG:5070 file on column `geom` with `primary_column: 123` comes
  out with no CRS, where the well-formed control keeps 5070. That is the same
  corruption shape #945's review caught, still live.
- `validate.py:730` — `gpio check spec`, the command you run to be *told* a file
  is malformed, dies with `AttributeError: 'list' object has no attribute
  'items'`.
- `check_parquet_structure.py:463` — `gpio check bbox` crashes on a non-object
  block, one layer above the guard this PR added.
- `streaming.py:206-224` — a non-string `primary_column` still reaches
  `quote_identifier` (#887 item 3, verbatim).

Also inaccurate above: `streaming.py:584` carries the identical
`geo_meta.get("primary_column", "geometry")` pattern, so this is not "the one
place still open to it". It is dead today (both callers pass `geometry_column`
explicitly), but it is a second site.

#968 also argues for the mechanical fix: a `meta`-lane guard that fails on any
new raw `geo_meta["columns"]` / `.get("primary_column")` read outside the shared
helpers, so there is no fourth instalment.

### DIFF
--- a/geoparquet_io/api/table.py
+++ b/geoparquet_io/api/table.py
@@ -2558,12 +2558,23 @@ class Table:
         if schema_metadata and b"geo" in schema_metadata:
             try:
                 geo_meta = json.loads(schema_metadata[b"geo"].decode("utf-8"))
+                # A read-only reader, the Python API's `gpio inspect meta`:
+                # `geo_metadata` is documented as the file's full `geo` block,
+                # so it is handed back exactly as the file holds it and is NOT
+                # sanitized -- the same line #883 drew around `parse_geo_metadata`
+                # and `check`. Only the derived keys below need guarding: they
+                # indexed the raw block and crashed on a `columns` that is not
+                # an object, or an entry that is not one (#947).
                 result["geo_metadata"] = geo_meta
 
                 # Extract geometry_types from geo metadata
-                columns_meta = geo_meta.get("columns", {})
-                if self._geometry_column and self._geometry_column in columns_meta:
-                    col_meta = columns_meta[self._geometry_column]
+                columns_meta = geo_meta.get("columns") if isinstance(geo_meta, dict) else None
+                col_meta = (
+                    columns_meta.get(self._geometry_column)
+                    if isinstance(columns_meta, dict) and self._geometry_column
+                    else None
+                )
+                if isinstance(col_meta, dict):
                     result["geometry_types"] = col_meta.get("geometry_types")
                     result["edges"] = col_meta.get("edges")
                     result["orientation"] = col_meta.get("orientation")
--- a/geoparquet_io/core/add/bbox_metadata.py
+++ b/geoparquet_io/core/add/bbox_metadata.py
@@ -32,7 +32,12 @@ from geoparquet_io.core.duckdb_utils import (
 )
 from geoparquet_io.core.exceptions import GeoParquetError
 from geoparquet_io.core.file_utils import resolve_file_url
-from geoparquet_io.core.geo_metadata import covering_supported, parse_geo_metadata
+from geoparquet_io.core.geo_metadata import (
+    covering_supported,
+    parse_geo_metadata,
+    sanitize_geo_metadata,
+    sanitized_carried_geo,
+)
 from geoparquet_io.core.geometry_detection import find_primary_geometry_column
 from geoparquet_io.core.logging_config import debug, success
 
@@ -218,12 +223,12 @@ def add_bbox_metadata_table(
 
     schema_metadata = dict(table.schema.metadata) if table.schema.metadata else {}
 
-    geo_meta: object = None
-    if b"geo" in schema_metadata:
-        try:
-            geo_meta = json.loads(schema_metadata[b"geo"].decode("utf-8"))
-        except (json.JSONDecodeError, UnicodeDecodeError):
-            geo_meta = None
+    # A write path: this block is re-serialized onto the returned table, so a
+    # malformed carried one goes through the shared shape check first (#947).
+    # Without it a `columns` mapping to a non-object let the covering be
+    # assigned into a string, and a wrong-typed `crs` was carried straight into
+    # the output.
+    geo_meta: object = sanitized_carried_geo(schema_metadata)
 
     # Same refusal as the file path: a table with no usable `geo` block has
     # nothing describing the geometry column, and the skeleton this used to
@@ -311,7 +316,12 @@ def add_bbox_metadata(
 
     # Get existing metadata
     metadata, _ = get_parquet_metadata(parquet_file)
-    geo_meta = parse_geo_metadata(metadata, False)
+    # `parse_geo_metadata` is the read-only reader and hands the block back as
+    # the file holds it; this command rewrites the file from it, so the
+    # malformed parts are dropped here the way every other write path drops
+    # them (#947). A `columns` entry that is not an object used to fail with
+    # `'str' object does not support item assignment` at the assignment below.
+    geo_meta = sanitize_geo_metadata(parse_geo_metadata(metadata, False))
 
     geo_meta = require_geo_metadata_for_covering(geo_meta, parquet_file)
 
--- a/geoparquet_io/core/common.py
+++ b/geoparquet_io/core/common.py
@@ -44,11 +44,13 @@ from geoparquet_io.core.geo_metadata import (
     DEFAULT_GEOPARQUET_VERSION,
     GEOPARQUET_VERSIONS,
     build_bbox_covering,
+    carried_geometry_column,
     create_geo_metadata,
     detect_bbox_column_from_schema,
     geoarrow_wkb_codes,
     prune_geo_metadata_to_columns,
     sanitize_geo_metadata,
+    sanitized_carried_geo,
     strip_derived_stats,
     strip_nonplanar_edges,
 )
@@ -57,7 +59,6 @@ from geoparquet_io.core.geoarrow_encoding import (
     is_wkb_extension_field,
 )
 from geoparquet_io.core.geometry_detection import (
-    STANDARD_GEOMETRY_NAMES,
     _detect_geometry_from_query,
     find_primary_geometry_column,
 )
@@ -2759,7 +2760,7 @@ def _geo_block_to_carry_on_fast_path(
     would not write itself — the caller then keeps its existing behaviour.
     """
     from geoparquet_io.core.crs_utils import apply_output_crs
-    from geoparquet_io.core.geo_metadata import _decode_geo_value, declare_carried_bbox_column
+    from geoparquet_io.core.geo_metadata import declare_carried_bbox_column
 
     if effective_version != "2.0" or not geometry_column or not original_metadata:
         return None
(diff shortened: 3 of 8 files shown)


cholmes:
## Multi-agent adversarial review — approve, with two body corrections

Reviewed in an isolated worktree, rebased onto current `main` (`f28fe48`). The rebase was clean *and* semantically verified: `git diff --stat origin/main` is byte-identical to the pre-rebase diff, no function was added or removed from any touched module, and the package imports.

**No CodeRabbit review exists on this PR**, so this was the only review it got.

### My review brief was wrong, and the evidence is thorough

I briefed the reviewer to hunt for the #945 failure shape — that PR removed a crash and replaced it with silently wrong coordinates, and this is the next instalment of the same class. It built **22 fixtures** (`geometry` and `geom` × `columns: null`/list/string, entry-as-string, block-as-list/string, `primary_column: 123`, `encoding: 123`, `crs: 42`, plus EPSG:5070 controls) and ran every one of the eight sites against both this branch and `origin/main`.

**The shape does not fire anywhere.** No `geom` file resolves to the literal `"geometry"`; no projected CRS is silently replaced. Two sites are *strictly better* than main:

| `write_geoparquet_table`, `geom` table, `primary_column: 123` | main | this PR |
|---|---|---|
| `primary_column` written | `123` (spec-invalid) | `"geom"` |
| `bbox` computed | absent (geometry unrecognised) | `[-100.0, 40.0, -100.0, 40.0]` |

And `gpio add bbox-metadata` wrote `"primary_column": 123` **into the file** on main; here it writes `"geom"` via #945's single-column repair, keeping the NAD83/Conus Albers CRS.

My other hypotheses also failed, correctly: `check bbox` reports truthfully on all ten malformed fixtures with no false PASS; `common.py`'s net −61 is three inlined copies replaced by two shared helpers, with case-insensitive name detection riding along; and a dropped `geometry_types` makes the carry return `None` so DuckDB regenerates — the fail-safe direction, leaving #952's sticky `[]` untouched.

Both judgement calls survive scrutiny. `Table.metadata()` has no internal consumer, and the block round-trips verbatim for all 22 fixtures with `geometry_column` resolving to `geom` every time.

### Two claims in the body were overstated — corrected above, and filed as #968

**"The third and *last* instalment" is false.** Four live members of the same class remain, all on `main`, all untouched here. I reproduced the lead one myself:

```
control   (primary_column: "geom")  extract geoparquet -> crs: 5070
malformed (primary_column: 123)     extract geoparquet -> crs: ABSENT  (=> OGC:CRS84)
```

`geo_metadata.py:826` (`_prune_geo_dict_to_columns`) reads `primary_column` raw, so a non-string value drops the **entire geo block, CRS included** — and an absent CRS means OGC:CRS84, so `extract` and `sort hilbert` silently relabel a projected file as lon/lat. That is the same corruption shape #945's review caught, still live, and higher severity than anything the three instalments have fixed. The fix is to route that read through the helper this PR just shared out.

Also verified: `gpio check spec` — the command you run to be *told* a file is malformed — dies with `AttributeError: 'list' object has no attribute 'items'` on a malformed `columns`.

**"The one place still open to it" is false** — `streaming.py:584` carries the identical `get("primary_column", "geometry")` pattern. Dead today (both callers pass the column explicitly), but a second site.

Neither blocks this PR: both are pre-existing and this PR does not touch that code. They are **#968**, which also argues for the mechanical fix — a `meta`-lane guard failing on any new raw `geo_meta["columns"]` read outside the shared helpers, so there is no fourth instalment. Four PRs have now fixed this class per-instance and each declared its batch complete; that is the signal to build the guard, the same lesson as #948's patch-target ratchet.

### Smaller notes for later

`_geo_block_to_carry_on_fast_path` can emit a block with no `primary_column` when sanitize drops a malformed one and `columns` has 2+ entries — unit-provable, unreachable today only because finding 1 drops the whole block first. `add_bbox_metadata_table` can return a block missing REQUIRED keys whose file then fails five of its own spec checks (byte-identical on main). Both are in #968.

One test note: `test_add_bbox_metadata_table_survives_a_malformed_block` passes against the merge base — `crs: 42` is not in `MALFORMED_BLOCKS` and the `except: pass` swallows. The fix is real (main returns `"crs": 42` in the table, this PR does not); it just is not pinned.

### Gates

Fast suite **6887 passed, 75 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100% on 66 changed lines** · `pytest -m meta` 205 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push all clean · **61 of the new tests fail against the merge base**, `geom` is in the parametrization, and all 403 tests in the file collect into the fast lane.

Issue #936, opened 2026-09-09
fix(extract,carto): remaining bare-identifier SQL injection sites (file-borne)


cholmes:
Found by the security sweep during the #925 review (AI review agent; exploitation demonstrated end-to-end for the extract site). Both confirmed present on current main. #925's body claimed kdtree was "the last known sibling" — it was not.

## 1. `core/extract.py:560` — file-borne, S2 (possibly S1)

```python
if geometry_wkt:
    escaped_wkt = _escape_sql_string(geometry_wkt)
    conditions.append(f'ST_Intersects("{geometry_col}", ST_GeomFromText(\'{escaped_wkt}\'))')
```

`geometry_col` is interpolated bare with hand-rolled double quotes. **Five lines above, the same function does it correctly**: `f"ST_Intersects({quote_identifier(geometry_col)}, ST_MakeEnvelope(...))"`. The WKT beside it is properly escaped — only the identifier is not.

Source is attacker-controlled: `geometry_col` = `find_primary_geometry_column(input_parquet)` = the file's own `geo.primary_column`. Same threat model as #918/#923 — a published GeoParquet file carries the payload.

Triggered by any `gpio extract ... --geometry <WKT>`. Verified via `--show-sql` that a crafted `primary_column` injects a call into the generated `WHERE` clause. Rated S2 rather than S1 only because function injection was demonstrated but a fully balanced silent exfiltration was not; a `WHERE` clause admits scalar subqueries, so treat it as S1-adjacent.

Fix: `quote_identifier(geometry_col)`, matching the sibling five lines up.

## 2. `core/carto.py:467` — operator-supplied, S3

```python
if fmt == "GeoJSON":
    read_expr = f'ST_Read("{full_url}")'
else:
    read_expr = f"read_csv_auto({sql_path(full_url)})"
```

The CSV branch on the very next line uses `sql_path`; the GeoJSON branch hand-rolls double quotes. `full_url` derives from `--url` and `_validate_carto_url` does not reject `"`. Self-inflicted (operator types the URL), so primarily a correctness bug — but a latent injection if the URL ever arrives from config or automation.

Note: a #904 review flagged this and its fixer reported migrating it; the change is not on main, so it was lost somewhere. Worth confirming rather than assuming.

## Also worth revisiting

[...]

PR #964, opened 2026-09-10
test(sql): pin the manual-quoting guard so no module can be exempted by path again

### LINKED ISSUE #946: fix(extract): #939's guard exemption is file-scoped and shields three fixable DuckDB sites
Found by the security review of #935 (AI review agent). This is a defect in the guard change **#939 itself made** — worth fixing promptly because it silently disables the new detection over a whole file.

## What #939 did

To land the unanchored `\"{` arm of `duckdb-antipatterns`, #939 exempted `extract_bigquery.py` wholesale (`.pre-commit-config.yaml:130-135`):

```
| grep -v '^geoparquet_io/core/extract_bigquery.py:' \
```

justified as: *"BigQuery reads `"..."` as a STRING literal, so its sites need backticks rather than quote_identifier and are tracked in #932."*

## Why that rationale is only half right

It holds for the **server-side** BigQuery branch — `extract_bigquery.py:724, 728, 732` — which are genuine #932 deferrals.

It does **not** hold for the **DuckDB local-filter** branch in the same function, `:738, 742, 747`:

```python
bbox_filter = f"ST_Intersects(\"{geom_col}\", ST_GeomFromText('{wkt}'))"
```

That is DuckDB SQL, where `"` *is* the identifier quote — `quote_identifier()` applies today, exactly as it does everywhere else. But the file-wide `grep -v` means the arm #939 added can never flag them, and they sit mis-tracked under a deferral whose reason doesn't apply to them.

## Reachability

`_determine_bbox_strategy` (`:659-689`) selects local filtering whenever the table is below `--bbox-threshold`, on explicit `--bbox-mode local`, or as the fallback when the row count is unavailable — the default path for small tables, not an edge case.

Exploitability is limited (hence S3, not higher): `geom_col` from schema auto-detection cannot carry a quote, since BigQuery column names are `[A-Za-z_][A-Za-z0-9_]*`. The reachable vector is `--geography-column`, because `_resolve_column_name` **falls through and returns the raw user string when it matches no schema column** (`:486`). Self-inflicted, so mainly a correctness/robustness bug plus a latent injection — but it is a one-line fix currently held back by a mis-scoped exemption.

## Fix

Narrow the exemption from file-scoped to the three server-side lines (a trailing `# allow-manual-quote` with a #932 reference is the idiom the config already supports), and apply `quote_identifier` to the three DuckDB-dialect sites. That restores detection over the rest of the file and puts #932's deferral where it actually belongs.

## Class status, for the record

With #935 merged, the reviewer found **no remaining site where a value from a data file's own metadata or schema reaches `execute()` as a bare identifier** — apart from #932's three genuine BigQuery-dialect deferrals and the three mis-shielded DuckDB sites above. The file-content-controlled identifier class is otherwise closed.

Refs #939, #932, #935.

cholmes:
## What changed

**First, the correction that shapes this PR: #946's code defect is already fixed on `main`.** #944 (merged 2026-09-09, after #946 was filed) removed the file-scoped exclusion *and* fixed the three DuckDB-dialect sites. I verified both rather than taking the PR body's word for it:

- `.pre-commit-config.yaml` carries no `grep -v '^geoparquet_io/core/extract_bigquery.py:'`; the only exempt path is `core/duckdb_utils.py`.
- `_build_bbox_filters` uses `quote_identifier(geom_col)` for its three DuckDB branches and `_quote_bigquery_identifier(geom_col)` for its three GoogleSQL ones.
- `pre-commit run duckdb-antipatterns --all-files` passes with the module un-exempted, so the un-exempted hook flags **nothing** in `extract_bigquery.py` today.

What was still missing is the part that makes #946 *recur*: nothing in the suite would notice the exclusion going back in. The bare-`duckdb.connect(` arm has a full hook self-test; the hand-rolled-quoting arm — the one that had to be file-exempted — had none.

So this PR adds `tests/test_duckdb_antipatterns_hook.py` (12 tests). It pulls the hook script straight out of `.pre-commit-config.yaml`, runs it against a synthetic `geoparquet_io/` tree, and pins both halves:

**Behavioural.** Each spelling the arm must catch — `f'"{col}"'`, the mid-f-string `\"{col}\"` shape from #936, and `col.replace('"', '""')` — is asserted rejected **from `geoparquet_io/core/extract_bigquery.py` itself**. That path is the fixture, so any path exclusion for that module turns straight back into a red test.

**Structural.** `TestManualQuoteRuleHasNoPathExemptions` reads the script and asserts the set of path-exempted modules is exactly `{core/duckdb_utils.py}` — which is exempt because it *implements* `quote_identifier()` and `_escape_sql_string()` and so cannot use them. It matches both shapes the script uses, `grep -v '^<path>:'` and `awk -F: '$1 != "<path>"'`, so neither spelling of a whole-module exemption can slip in. A companion test asserts the real `extract_bigquery.py` needs no exemption and carries no `# allow-manual-quote`.

Controls keep the rejections honest and prevent a vacuous pass: the same hostile source in an ordinary module, a clean module, the factory exemption, a nested `duckdb_utils.py` (not exempt — the exemption names one path), and the per-line `# allow-manual-quote` hatch.

Plus seven lines of comment in `.pre-commit-config.yaml` pointing the next person at the per-line hatch instead of another `grep -v`, and naming the test that enforces it.

**No production code changed.** The three sites #946 names are already correct; re-touching them would have been churn.

## Why

The task in #946 was to replace a whole-file exclusion with the narrowest possible suppression. #944 did better than narrowing it — it removed the need for one entirely. But the failure mode #946 actually describes is *"a guard with a hole in it is worse than no guard, because it reads as protection"*, and that failure mode is a **process** one: a file-scoped `grep -v` is one line in a config, it silences an unbounded set of future sites, and it is invisible in every gate. #939 added one and it took a security review of a downstream PR to find it. Removing that particular instance does not stop the next one.

The narrowest-possible-suppression rule is now **enforced**, not just documented: the only supported way to silence a site is a trailing `# allow-manual-quote` on the one line, which is visible in the diff that introduces it and cannot spread.

On the residual concern #946 raises under "Reachability" — `_resolve_column_name` falling through to the raw user string — I traced it and it is **not reachable**. Its sole caller (`:982`) sits behind `_detect_geometry_column_from_schema` (`:973`), which raises `ValueError` when `--geography-column` names no schema column, so `_resolve_column_name` is only ever reached for a column that exists and returns the schema's own spelling. With `quote_identifier` now applied, an unmatched name would in any case produce a quoted identifier naming no column — a clean error, not an injection. No change needed.

## Verification

**Reintroduce-and-catch, run both ways.** Planted `local_col = f'"{geom_col}"'` at `extract_bigquery.py:776`, in place of `quote_identifier(geom_col)`:

```
# current main's config
duckdb-antipatterns......................................................Failed
geoparquet_io/core/extract_bigquery.py:776:        local_col = f'"{geom_col}"'
ERROR: Do not hand-roll SQL identifier quoting.

# #939-era config (git show 6752c40:.pre-commit-config.yaml), identical source
duckdb-antipatterns......................................................Passed
```

That is #946 demonstrated end to end: the same hand-rolled quote, caught today, silently accepted under the file-scoped exclusion. Both edits reverted; `git status` clean before committing.

**TDD.** The new tests were run against the defect state — #939's config *and* its `extract_bigquery.py` (`git show 6752c40:...`) — before being run against `main`:

```
FAILED ...::test_rejected_in_the_formerly_exempt_module[f_string_quotes]
FAILED ...::test_rejected_in_the_formerly_exempt_module[escaped_mid_string]
FAILED ...::test_rejected_in_the_formerly_exempt_module[replace_doubling]
FAILED ...::test_duckdb_utils_is_the_only_exempt_path
  AssertionError: Found: ['geoparquet_io/core/duckdb_utils.py',
                          'geoparquet_io/core/extract_bigquery.py']
FAILED ...::test_the_real_extract_bigquery_module_needs_no_exemption
5 failed, 7 passed
```

→ `12 passed` on `main`. The 7 that pass in both states are the controls, which by design do not depend on the exclusion — so the 5 that flip are exactly the #946 surface.

### Gates

```
pytest tests/test_duckdb_antipatterns_hook.py tests/test_duckdb_connection_factory_bypass.py \
       tests/test_geometry_identifier_quoting.py tests/test_extract_bigquery.py
                                              187 passed, 3 deselected
pytest -n auto -m "not slow and not network and not meta"
                                              6764 passed, 75 skipped, 9 xfailed
pytest -m meta                                205 passed
pre-commit run --all-files                    clean
pre-commit run --all-files --hook-stage pre-push
                                              clean (deptry, mypy, vulture, xenon,
                                                     import-linter, menard-check, fast-tests)
diff-cover vs origin/main                     no lines with coverage information
                                              (config comments + a new test module;
                                               zero production lines changed)
```

The one fast-suite failure, `test_geojson_stream.py::TestPipelineIntegration::test_streaming_handles_broken_pipe`, is the documented `-n auto` saturation flake on a shared machine: `64 passed` re-run serially, and this PR changes no production code it could touch.

Closes #946

## Summary by CodeRabbit

- **Tests**
  - Expanded validation for the SQL identifier-quoting rule across exempt and non-exempt modules.
  - Added coverage for permitted per-line exceptions and the designated factory-module exemption.
  - Added checks confirming other modules cannot bypass the rule.
  - Tests now handle environments where Bash is unavailable by skipping relevant checks.

- **Documentation**
  - Clarified the approved exception approach and warned against broad module-level exclusions.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -132,6 +132,13 @@ repos:
             # sites use _quote_bigquery_identifier -- GoogleSQL reads "..." as a
             # STRING literal, so identifiers there are backtick-quoted -- so the
             # rule applies to it like everywhere else.
+            # Do NOT silence a new site with another `grep -v '^<path>:'`: that
+            # switches this rule off over the whole module, which is how #939's
+            # exemption hid three unrelated DuckDB-dialect sites until #946.
+            # duckdb_utils.py below is the only path that may ever be exempt;
+            # tests/test_duckdb_antipatterns_hook.py enforces that, and covers
+            # each spelling this arm has to catch. Silence one line instead,
+            # with a trailing "# allow-manual-quote".
             if grep -rnE 'f'"'"'"\{|\\"\{|\.replace\('"'"'"'"'"', '"'"'""'"'"'\)' geoparquet_io/ --include="*.py" \
                 | grep -v '^geoparquet_io/core/duckdb_utils.py:' \
                 | grep -v 'allow-manual-quote' \
--- a/tests/test_duckdb_antipatterns_hook.py
+++ b/tests/test_duckdb_antipatterns_hook.py
@@ -1,0 +1,243 @@
+"""Self-tests for the ``duckdb-antipatterns`` hook's hand-rolled-quoting arm.
+
+The rule bans ``f'"{col}"'`` and ``col.replace('"', '""')`` in favour of
+:func:`duckdb_utils.quote_identifier` / :func:`duckdb_utils._escape_sql_string`,
+because a column name can arrive from a file's own ``geo.primary_column`` or
+from ``--column`` / ``--geography-column`` -- an injection surface, not a nit.
+
+Why these tests exist (#946). When #939 unanchored the ``\\"{`` arm, the newly
+visible sites in ``core/extract_bigquery.py`` were silenced with a **file-scoped**
+exclusion in ``.pre-commit-config.yaml``::
+
+    | grep -v '^geoparquet_io/core/extract_bigquery.py:' \\
+
+That switched the whole rule off over an entire module, not just the three
+BigQuery-dialect lines the rationale covered, and nothing in the suite noticed.
+#944 fixed the sites and dropped the exclusion; these tests keep it dropped:
+
+* the spelling tests run their hostile source **from that exact path**, so any
+  path-scoped exclusion for it turns straight back into a failure, and
+* :class:`TestManualQuoteRuleHasNoPathExemptions` reads the config and asserts
+  that ``core/duckdb_utils.py`` -- which implements the two helpers and so
+  cannot use them -- is the only path the hook exempts at all.
+
+The sibling arm that bans bare ``duckdb.connect(`` is self-tested in
+``tests/test_duckdb_connection_factory_bypass.py``.
+"""
+
+from __future__ import annotations
+
+import re
+import subprocess
+from pathlib import Path
+
+import pytest
+import yaml
+
+REPO_ROOT = Path(__file__).parent.parent
+
+# The one module the hook is allowed to exempt: quote_identifier() and
+# _escape_sql_string() live there, so their own implementations necessarily
+# spell out the quoting the rule forbids everywhere else.
+FACTORY_MODULE = "geoparquet_io/core/duckdb_utils.py"
+
+# The module #939 exempted wholesale. Every rejection test below plants its
+# source here, so re-adding an exclusion for it fails these tests.
+FORMERLY_EXEMPT_MODULE = "extract_bigquery.py"
+
+
+def _bash_can_run_scripts() -> bool:
+    """Whether ``bash`` on this platform can actually execute a script.
+
+    ``shutil.which("bash")`` is not enough on Windows runners: there ``bash``
+    resolves to WSL's ``bash.exe``, which exits non-zero with an "install a
+    distribution" notice when no WSL distro is present -- which would pass
+    every rejection test below for entirely the wrong reason.
+    """
+    try:
+        probe = subprocess.run(["bash", "-c", "exit 0"], capture_output=True, timeout=30)
+    except (OSError, subprocess.SubprocessError):
+        return False
+    return probe.returncode == 0
+
+
+_NEEDS_BASH = pytest.mark.skipif(
+    not _bash_can_run_scripts(),
+    reason="pre-commit hook scripts are POSIX shell; no usable bash on this platform",
+)
+
+
+def _hook_script() -> str:
+    """The duckdb-antipatterns script from .pre-commit-config.yaml, exactly as
+    pre-commit invokes it."""
+    config = yaml.safe_load((REPO_ROOT / ".pre-commit-config.yaml").read_text(encoding="utf-8"))
+    for repo in config["repos"]:
+        for hook in repo.get("hooks", []):
+            if hook["id"] == "duckdb-antipatterns":
+                return str(hook["args"][-1])
+    raise AssertionError("duckdb-antipatterns hook not found in .pre-commit-config.yaml")
+
+
+# Each spelling the rule has to catch. The escaped-quote one is #936's shape:
+# it sits mid-f-string, which is why #939 had to unanchor the pattern.
+MANUAL_QUOTE_SPELLINGS = {
+    "f_string_quotes": "def build(col):\n    return f'\"{col}\"'\n",
+    "escaped_mid_string": (
+        'def build(col, wkt):\n    return f"ST_Intersects(\\"{col}\\", \'{wkt}\')"\n'
+    ),
+    "replace_doubling": "def build(col):\n    return '\"' + col.replace('\"', '\"\"') + '\"'\n",
+}
+
+
+@_NEEDS_BASH
+class TestHookRejectsManualQuoting:
+    """The rule must fire on every spelling, in every module -- including the
+    one #939 exempted by path."""
+
+    def _run_hook(self, cwd: Path) -> subprocess.CompletedProcess[str]:
+        return subprocess.run(
+            ["bash", "-c", _hook_script()],
+            cwd=cwd,
+            capture_output=True,
+            text=True,
+            encoding="utf-8",
+            errors="replace",
+        )
+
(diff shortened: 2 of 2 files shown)


cholmes:
## Multi-agent adversarial review — the ratchet had a hole; fixed in 032649d

The core design is right and it works: the **behavioural** half uses `extract_bigquery.py` itself as the fixture, so any path-scoped exclusion for that module turns straight back into a red test. That is the strongest idea in the PR, and a 20-mutation sabotage matrix confirms it — every spelling of a whole-module exemption aimed at that module goes red, including four the structural regexes never see.

But the **structural** half did not do what its own docstring claimed, and the gap is the same shape as the defect the PR exists to prevent.

### Two problems, one fix

**It blacklisted two exemption spellings.** A blacklist is only as good as its list, and four functionally identical whole-module exemptions sailed through — verified, each leaving the suite at 12 passed:

```
grep -v 'geoparquet_io/core/convert.py:'     (no caret)
grep -v "^geoparquet_io/core/convert.py:"    (double quotes)
awk -F: '$1 !~ /convert/'                    (regex, not !=)
grep -v "^$EXEMPT:"                          (path in a shell variable)
```

**And it asserted over the whole script, not the arm under test.** The hook has six arms and two others legitimately exempt `duckdb_utils.py`, so the set could dedupe to `{duckdb_utils.py}` with **zero contribution from the manual-quote arm**. Re-quoting that arm's own exemption stage left the suite green — the thing being pinned could be reworded out from under the assertion.

Replaced with a **whitelist of the arm's own pipeline**, stage by stage, scoped to the arm by its error message. Any added or reworded filter now fails whatever spelling it uses, including shapes nobody has thought of. All four spellings above now fail, plus the `files:`-key hole (which does not narrow what the script scans — it does its own recursive grep — but turns the whole hook into a skip).

The module docstring said the test asserts `duckdb_utils.py` "is the only path the hook exempts **at all**". It never checked that. It now describes what it does and why a whitelist beats a blacklist here.

I flagged pre-commit's `exclude:` YAML key as the most important hole. **It does not bypass the hook** — `pass_filenames: false` means the script does its own `grep -r geoparquet_io/`, so `exclude:` only trims the trigger set. Reproduced with a live planted violation: the hook still failed. Only a `files:` pattern matching nothing silences it, and that is loud in every CI log (`(no files to check)Skipped`) rather than a quiet one-line `grep -v`. Now pinned anyway.

The reachability claim in the body also holds — I could not construct a path to `_resolve_column_name`'s raw fall-through. `_detect_geometry_column_from_schema` raises when `--geography-column` names no schema column, and both functions `DESCRIBE` the same table with the same case-folded comparison, so a different-case name returns the schema's own spelling. Every consumer quotes anyway, and both quoting helpers are correct. No injection surface.

### Filed separately: two sibling guards have the hole this one no longer has — #970

`no-click-echo` and `forbid-bespoke-schema-reconciliation` both filter with unanchored `grep -v`, so they match the whole `path:line:text` line. Verified against the real commands:

```
clean.py   def _align_schema(a, b):                       -> FLAGGED (correct)
evil.py    def _align_schema(a, b):  # mirrors common.py  -> SILENCED by a comment
```

And `grep -v "common.py"` is a substring match on the path too, so it silently exempts **four** modules rather than one — `core/common.py` plus `partition/common.py`, `process/aggregate/common.py` and `grid_common.py`. Three real modules are outside that guard today. Out of scope here; **#970** carries it, and suggests reusing this file's stage-list pattern (which would also let the ~60 duplicated lines shared with `test_duckdb_connection_factory_bypass.py` move to a common helper).

### Verified, not taken on trust

All four of the body's claims about `main` check out: no bigquery exemption in the config; `duckdb_utils.py` is the only exempt path; `_build_bbox_filters` uses `_quote_bigquery_identifier` for its three GoogleSQL branches and `quote_identifier` for its three DuckDB ones; the un-exempted hook flags nothing. The extraction is not vacuous — it raises `AssertionError` if the hook id moves and `KeyError` if `args` disappears — and the file carries no `meta` marker, so its 14 tests run in the blocking PR job.

### Gates

Fast suite **6901 passed, 75 skipped, 9 xfailed, 0 failed** · `pytest -m meta` 205 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit **0 failures** · `diff-cover` correctly reports nothing to measure — this PR changes zero production lines.

PR #965, opened 2026-09-10
fix(extract): stop arcgis guessing geometry_types when the data does not say

### LINKED ISSUE #962: fix(extract): arcgis geometry_types falls back to a guess when the compute finds nothing, not just when it fails
## What

`_resolve_geometry_types` (`geoparquet_io/core/arcgis.py:1216-1220`) is:

```python
computed = _compute_geometry_types(table, "geometry", verbose)
if computed:
    return computed
declared = ARCGIS_GEOM_TYPES.get(esri_geometry_type)
return [declared] if declared else []
```

`if computed:` cannot tell "geoarrow could not read this" from "there was nothing to describe". Both fall through to the mapping table's guess — which is exactly the declared-but-wrong value #928 was filed to stop producing.

Two reachable cases:

1. **One malformed row poisons the whole column.** `ga.unique_geometry_types` is all-or-nothing: a column of `[Polygon, b"junk"]` raises rather than returning the types it could read, so an `esriGeometryPolygon` layer declares `["MultiPolygon"]` over data whose only readable geometry is a single `Polygon`.
2. **An all-NULL geometry column.** Reachable in production from FeatureServer records with `"geometry": null`, or a `--where` that selects only those. `[NULL, NULL]` under `esriGeometryPolygon` declares `["MultiPolygon"]` over a column containing zero geometries. NULLs are correctly filtered by the compute, so this is the "nothing to describe" case landing in the "compute failed" branch.

Neither is caught by `validate_geoparquet` — the all-NULL file reports zero failed and zero warnings.

## Why it matters

The PR that introduced this (#956) argues, correctly, that *"a declared-but-wrong type is worse than an absent one"*. These two paths are where the new code still declares a wrong type. The spec's `[]` ("the types are not known") is the right answer for case 2, and for case 1 the honest answer is either `[]` or the types that could be read.

## Fix

Distinguish the two outcomes rather than testing truthiness — fall back to `ARCGIS_GEOM_TYPES` only when the compute actually raised, and return `[]` when it merely found nothing:

```python
try:
    computed = _compute_geometry_types(table, "geometry", verbose)
except Exception:
    declared = ARCGIS_GEOM_TYPES.get(esri_geometry_type)
    return [declared] if declared else []
return computed  # [] is a legitimate answer: nothing to describe
```

Worth checking whether `_compute_geometry_types` swallows the geoarrow exception internally first, in which case the distinction has to be made one level down.

## Severity

Low impact in practice: arcgis geometry is always produced by DuckDB `ST_AsWKB` (`arcgis.py:991`), which does not emit unreadable WKB, so `ARCGIS_GEOM_TYPES` is close to dead code outside tests. The all-NULL case is the more realistic of the two.

Note this also adds a producer of `[]`, which #952 documents as sticky — no file→file gpio command recomputes an empty `geometry_types`.

Found by an adversarial review of #956.

cholmes:
## What changed

`_resolve_geometry_types` no longer falls back to the five-entry `ARCGIS_GEOM_TYPES`
table when the computed types come back empty. The table is deleted; the layer's
declared type now appears only as context in a warning, never in the file.

The two ways of arriving at `[]` are told apart by counting the non-NULL
geometries before computing, because `_compute_geometry_types` swallows the
geoarrow error itself and returns `[]` — so catching cannot distinguish them one
level up, as #962 suspected. Both still declare `[]`; the distinction drives the
message, not the value:

- **nothing to describe** — no geometry column, or every value NULL →
  `Fetched data holds no geometries; declaring geometry_types as [] (types not known).`
- **could not read** — non-NULL geometry the compute could not parse →
  `Fetched geometry could not be read, so geometry_types is declared as [] rather than guessed from the layer's esriGeometryPolygon.`

## Why

#956 stopped the geo block inventing `"Geometry"`, but `if computed:` put the
guess straight back wherever the compute returned nothing. Three reachable paths,
all measured against `esriGeometryPolygon`:

| input | declared before | declared after | readable in the file |
|---|---|---|---|
| `[Polygon, b"junk"]` | `["MultiPolygon"]` | `[]` | `Polygon` |
| `[NULL, NULL]` | `["MultiPolygon"]` | `[]` | none |
| `--exclude-cols geometry` | `["MultiPolygon"]` | `[]` | no geometry column |

`ga.unique_geometry_types` is all-or-nothing, so one malformed row empties the
answer for a whole column of good geometry. An all-NULL column has nothing to
describe — the compute filters NULLs by design, so it lands in the "compute
failed" branch. And `repair_arrow_table_geometry` passes a table with no geometry
column straight through, so `--exclude-cols geometry` reaches the geo block too.
In each case the layer's advertisement named a type the file does not contain.

**What the mapping table is still for: nothing.** It is not merely wrong in the
three cases above, it has no remaining case to serve.

- The one scenario where the declared type is the only information available —
  an empty result set — never reaches the geo block: `arcgis_to_table` returns a
  bare table at `total_count == 0` and raises at `total_rows == 0`, so the block
  is only ever built over at least one row.
- The unreadable-WKB case it was kept for is unreachable in practice: this
  geometry is always DuckDB `ST_AsWKB` output (`arcgis.py:991`), which is
  readable by construction.
- Even where it did fire, it is coarser than what it describes
  (`esriGeometryPolyline` → `MultiLineString` over LineString data, no Z/M) and
  covers five of Esri's types.

**Why not the partial answer.** #962 allows "the types that could be read" for
the malformed-row case. Declaring those claims the unreadable rows are among
them, which is exactly the unverifiable claim this PR removes — and a conformant
reader skips rows whose type is not declared, so under-declaring is data loss,
not a cosmetic mismatch. `[]` makes no claim.

**All-NULL vs could-not-read.** Both declare `[]`, and deliberately so: a column
holding zero geometries has zero types present, which is the same empty array the
spec uses for "not known". Declaring the layer's type over it would be vacuously
true and unverifiable. What differs is the warning — one is a data-quality
problem worth chasing, the other is just an empty column.

This adds a producer of `[]`, which #952 documents as sticky.

**Noticed, not fixed:** `extract arcgis --exclude-cols geometry` writes a geo
block whose `primary_column` names a column that is not in the file, which
`gpio check spec` fails with `geometry column "geometry" not found in schema`.
`extract geoparquet` handles the same case by writing plain Parquet with no `geo`
metadata. That is a separate bug and wants its own change; this PR only stops the
`geometry_types` inside that block from being invented.

## Verification

  → **6758 passed, 75 skipped, 9 xfailed** in 171s.
- Meta lane: `uv run pytest -m meta -q` → **205 passed**, 7585 deselected.
- Diff coverage: `uv run diff-cover coverage.xml --compare-branch origin/main --fail-under=90`
  → **100%** (7/7 changed lines).
- `uv run pre-commit run --all-files` clean; `--hook-stage pre-push` clean
  (vulture included, which is what confirms the deleted table had no other user).
- Five tests written failing first, against the shipped code, then made to pass:
  the malformed row, the all-NULL column, the excluded column, the mapped Esri
  type with unreadable WKB, and one that pins the two warnings apart.

Closes #962

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ArcGIS geometry type detection for exported GeoParquet files.
  * Files now correctly indicate unknown geometry types when geometry data is missing, empty, malformed, or unreadable.
  * Added clearer warnings distinguishing unavailable geometry data from empty geometry columns.

### DIFF
--- a/geoparquet_io/core/arcgis.py
+++ b/geoparquet_io/core/arcgis.py
@@ -49,15 +49,6 @@ WKID_TO_EPSG = {
     102113: 3785,  # Legacy Web Mercator
 }
 
-# Map ArcGIS geometry types to GeoJSON types
-ARCGIS_GEOM_TYPES = {
-    "esriGeometryPoint": "Point",
-    "esriGeometryMultipoint": "MultiPoint",
-    "esriGeometryPolyline": "MultiLineString",
-    "esriGeometryPolygon": "MultiPolygon",
-    "esriGeometryEnvelope": "Polygon",
-}
-
 
 @dataclass
 class ArcGISAuth:
@@ -1199,25 +1190,48 @@ def _stream_features_to_parquet(
 def _resolve_geometry_types(table: pa.Table, esri_geometry_type: str, verbose: bool) -> list[str]:
     """Geometry types for the geo block, read from the fetched WKB (#928).
 
-    The layer's advertised ``geometryType`` is only a fallback. It is coarser
-    than the data (Esri's ``esriGeometryPolyline`` covers both LineString and
-    MultiLineString, and says nothing about Z/M), it does not cover every Esri
-    type, and it used to fall back to the literal ``"Geometry"`` -- which the
+    The layer's advertised ``geometryType`` does not describe the file. It is
+    coarser than the data (Esri's ``esriGeometryPolyline`` covers both
+    LineString and MultiLineString, and says nothing about Z/M), it does not
+    cover every Esri type, and it used to fall back to ``"Geometry"`` -- which the
     GeoParquet schema does not allow, so unmapped layers wrote files that
     failed validation on their own metadata.
 
     Computing from the data instead is what every gpio write path does, gives
     the spec's " Z"/" M"/" ZM" suffixes for free, and keeps the declaration
-    consistent with the statistics a reader checks it against. When the WKB
-    cannot be read the mapping table still serves as a fallback, and an Esri
-    type outside it yields ``[]`` -- the spec's way of saying the types are
-    not known.
+    consistent with the statistics a reader checks it against.
+
+    The declared type is no longer a fallback either (#962). Falling back
+    whenever the compute came back empty put the invented value straight back on
+    three reachable paths, because an empty result does not mean "the data was
+    fine and said nothing": ``unique_geometry_types`` is all-or-nothing, so one
+    malformed row empties the answer for a whole column of good geometry; an
+    all-NULL column has nothing to describe; and ``--exclude-cols geometry``
+    leaves no column to read. In each case the layer's coarse advertisement
+    (``esriGeometryPolygon`` -> MultiPolygon, over data holding a single
+    Polygon or no geometry at all) named a type the file does not contain.
+    ``[]`` -- the spec's way of saying the types are not known -- is the honest
+    answer wherever the data does not supply one, and gpio's own
+    geometry_types-vs-statistics check would flag the guess anyway.
+
+    The two ways of arriving at ``[]`` are still worth telling apart for the
+    user: unreadable geometry is a data-quality problem worth chasing, an empty
+    column is not. They cannot be told apart by catching, because
+    ``_compute_geometry_types`` swallows the geoarrow error itself, so the
+    non-NULL geometries are counted first.
     """
+    geometry = table.column("geometry") if "geometry" in table.column_names else None
+    if geometry is None or geometry.null_count == len(geometry):
+        warn("Fetched data holds no geometries; declaring geometry_types as [] (types not known).")
+        return []
+
     computed = _compute_geometry_types(table, "geometry", verbose)
-    if computed:
-        return computed
-    declared = ARCGIS_GEOM_TYPES.get(esri_geometry_type)
-    return [declared] if declared else []
+    if not computed:
+        warn(
+            "Fetched geometry could not be read, so geometry_types is declared as [] "
+            f"rather than guessed from the layer's {esri_geometry_type}."
+        )
+    return computed
 
 
 def arcgis_to_table(
--- a/geoparquet_io/core/arcgis.py
+++ b/geoparquet_io/core/arcgis.py
@@ -1221,7 +1221,16 @@ def _resolve_geometry_types(table: pa.Table, esri_geometry_type: str, verbose: b
     non-NULL geometries are counted first.
     """
     geometry = table.column("geometry") if "geometry" in table.column_names else None
-    if geometry is None or geometry.null_count == len(geometry):
+    if geometry is None:
+        # The user dropped the column with --exclude-cols; the service may well
+        # have returned geometry. Blaming the fetch here would point at the
+        # wrong thing and give the user nothing to act on.
+        warn(
+            "The geometry column was excluded, so there is nothing to describe; "
+            "declaring geometry_types as [] (types not known)."
+        )
+        return []
+    if geometry.null_count == len(geometry):
         warn("Fetched data holds no geometries; declaring geometry_types as [] (types not known).")
         return []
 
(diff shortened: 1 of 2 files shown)


cholmes:
## Multi-agent adversarial review — merge-ready, one S3 fixed in bd8fe6e

Reviewed in an isolated worktree, rebased onto current `main` (`68d5f38`).

### The central claim survived every falsification attempt

I briefed the reviewer to break "the mapping table has no remaining case to serve", because deleting a fallback is irreversible in effect. Every avenue closes, and **two of my own hypotheses turned out not to exist for `extract arcgis` at all**:

- **`--bbox-mode server` is not a thing here.** It exists only in `extract_bigquery.py` and `wfs.py`.
- **The zero-feature path never reaches the geo block.** `arcgis.py:1330-1342` returns a table with no `geo` key; the block is built downstream and computes `[]` because `num_rows == 0`. Identical before and after this PR.
- `--limit 0` raises `No features returned from service` before the block.
- A server omitting geometry yields NULL via `ST_AsWKB` (the only geometry producer, for both page converters), so it lands in the all-NULL branch — where `[]` is right and the *old* code silently declared `["MultiPolygon"]`.

Five new tests fail against merge-base behaviour, confirming all three rows of the before/after table independently. All 13 collect into the fast lane. The `-3` deletions are a rename plus parametrization — the unmapped case is retained verbatim and a mapped case added, so it is strengthened, not weakened. #956's two bowtie tests still pass unmodified.

### Fixed before merge: a warning that blamed the wrong thing

`--exclude-cols geometry` reported `Fetched data holds no geometries` — but the fetch held perfectly good polygons; the user dropped the column. That points at the service for something the user did, and gives them nothing to act on. It now has its own message, and a test pins it — nothing covered this third path, since the existing distinction test compares only the unreadable and all-NULL cases and the exclude-cols test asserts the value without looking at what the user is told. Sabotage-verified.

### A correction to #952, which I filed

The body's note that this "adds a producer of `[]`, which #952 documents as sticky" is **overstated, and #952 was too broad**. Verified on a two-geometry file with both columns emptied:

```
before:  geometry=[]         boundary=[]
after :  geometry=['Point']  boundary=[]     <- gpio convert geoparquet
```

`convert geoparquet` forces a metadata rewrite that replaces the **primary** column's `geometry_types` unconditionally, rather than going through `_compute_missing_metadata`'s key-absent gate. Since this PR only ever writes `[]` on the primary column, every file it can produce is repairable today with one command. The stickiness is real only for **secondary** columns. I have posted that correction on #952.

### `--exclude-cols geometry` output is invalid — before and after

Stating it plainly since it came up: that path writes a geo block whose `primary_column` names a column not in the file, and `gpio check spec` reports 6 failures. **Byte-for-byte identical on the merge base** — `geometry_types` plays no part in it. That is #966, and fixing it here would mean deciding whether to drop the whole `geo` key, in a different function, with the sibling `carto`/`wfs`/`bigquery` audit #966 asks for. Correctly deferred; the body's "Noticed, not fixed" paragraph is accurate.

### Checked and clean

`ARCGIS_GEOM_TYPES` deletion is total — zero hits across the tree including tests, docs and notebooks; not imported, not public surface; `vulture` and `deptry` clean. The perf concern I raised is unfounded with a measurement: the count uses `ChunkedArray.null_count`, which Arrow caches — **3 µs on 2M rows across 16 chunks**, not an O(n) scan. Both warnings fire in the right branch only, are `warn()` from `core/logging_config`, and do not leak into successful runs; chunked, all-NULL, mixed and no-column shapes all count correctly.

One note for #966: the `geometry is None` half of this guard becomes dead if #966 is fixed by dropping the block entirely, and vulture will not catch a dead *branch*.

### Gates

Fast suite **6893 passed, 75 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100%** · `pytest -m meta` 205 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push all clean.

PR #967, opened 2026-09-10
fix(sort)!: snap --row-group-size to the writer's 2,048-row vector; default 50,000 → 49,152

### LINKED ISSUE #961: fix(check): gpio check optimization fails a file gpio sort just wrote at its own default
## What

`gpio sort` asks for 50,000 rows per group (`DEFAULT_SORT_ROW_GROUP_ROWS`, the top of the spatial-advice band 10,000–50,000). DuckDB rounds a row-group request up to a multiple of its 2,048-row vector size, so what actually lands on disk is **51,200** — just outside the band. `gpio check optimization` then scores that factor `[fail]`:

```
$ gpio sort hilbert n102400.parquet n102400_sorted.parquet
$ python -c "import pyarrow.parquet as pq; m=pq.ParquetFile('n102400_sorted.parquet').metadata; print(m.num_rows/m.num_row_groups)"
51200.0

$ gpio check optimization n102400_sorted.parquet
Score: 2/5
  [fail] Row Group Size: Average 51,200 rows per group (optimal 10k-50k for spatial queries)
```

gpio tells the user to re-partition a file gpio just produced at its default, and docks a point for it. This is the shape #775 predicted and #795 was filed about — *"gpio check will then advise the user to change what gpio sort just chose"*.

Reproduces for any file with more than a couple of groups: 102,400 rows → 2 groups → avg exactly 51,200; 10M rows → 196 groups → avg 51,020.

## Why it is filed separately

#958 fixed the *self-contradiction in the row-group check's own messages* (a verdict band and an advice band that read as if one corrected the other) and documented both bands. It deliberately changed no verdict, status or score. This instance needs a decision rather than a wording fix, so it did not ride along — #958 is `Refs #795` rather than `Closes #795` for exactly this reason.

## Options (a decision, not a bug fix)

1. **Have `gpio sort` request a value that rounds to ≤50,000** — e.g. 49,152 (2,048 × 24). Keeps the band honest and changes only what gpio itself writes. Changes the sort default, so it is user-visible.
2. **Raise the top of `SPATIAL_ROW_COUNT_RANGE`** to something ≥51,200. Cheapest, but the band has measurement behind it (#775) and 50,000 is the measured point — moving it to accommodate a rounding artifact weakens the evidence claim.
3. **Have `check optimization` score against `GENERAL_ROW_COUNT_RANGE`** and keep 10k–50k purely as advice, matching what #958 did for `check row-group`. Most consistent with #958's verdict-vs-advice split, but it changes existing scores, which #958 was careful not to do.
4. **Round the request down to the vector multiple at the writer** so a request of 50,000 yields 49,152 rather than 51,200.

My weak preference is 1 or 4 — the band is the measured thing and should not move to accommodate a rounding artifact.

## Also still open from #795

`core/check_fixes.py` carries five hardcoded `row_group_rows=100000` constants (`:97`, `:155`, `:240`, `:313`, `:360`). #795 asks that these derive from the same decision rather than being hardcoded, and that `fix_spatial_order` inherit the sort default. Untouched by #958.

## Related

A third, adjacent inconsistency of the same class, found while reviewing #958: the small-file leniency at `core/check_parquet_structure.py:165` gates on `num_groups == 1`, so a 56 KB file written as 3 row groups is told "Row count per group is very low", directly under a printed guideline saying "Small files (<64 MB): single row group is fine". The same 56 KB as one group passes.

Found by an adversarial review of #958.

cholmes:
## What changed

`gpio sort` asked for **50,000** rows per row group — the top of the 10,000–50,000 band `gpio check` advises — and **51,200** landed on disk, because the Parquet writer emits row groups in whole 2,048-row vectors and rounds a request *up* to a multiple of that. `gpio check optimization` then scored the file gpio had just written `[fail]` on its row-group factor and told the user to re-partition it.

**Neither band moves.** `GENERAL_ROW_COUNT_RANGE` and `SPATIAL_ROW_COUNT_RANGE` are exactly as #958 merged them. What changed is the request: the sort commands now snap it to the nearest whole writer vector before handing it over, so the number gpio asks for is the number that lands.

- The default becomes **49,152** (24 × 2,048), the largest whole vector at or below the band's top.
- An **explicit `--row-group-size` gets the same treatment**, so a user who types the band's own top of 50,000 also lands inside it. The alignment lives in `resolve_sort_row_group_rows`, the one place all four sort subcommands already route through — 4 call sites, no new plumbing.
- The adjustment is **announced, not silent**. A value that is already a whole number of vectors passes through without a word; one that is not gets a line naming both numbers:

  ```
  $ gpio sort hilbert in.parquet out.parquet --row-group-size 50000
  Writing 49,152 rows per row group rather than the requested 50,000: the Parquet
  writer emits row groups in whole 2,048-row vectors, so 50,000 is not a size a
  row group can have.
  ```

  An `info()` line rather than a `--verbose` note, because it changes what gets written and the user should not have to opt in to learning that. `str_order` already uses `info()` for its tile-size fallback, so this matches the house style.

### The rounding, measured rather than taken from the issue

500,000 rows, one `COPY … (FORMAT PARQUET, ROW_GROUP_SIZE n)` per row, DuckDB 1.5.5, largest row group read back from the footer:

| requested | max rows/group | note |
|---|---|---|
| 3,000 | **4,096** | nearest multiple is 2,048 — so this is a *ceiling*, not a nearest |
| 5,000 | **6,144** | nearest multiple is 4,096 — same |
| 10,000 | 10,240 | |
| 49,152 | **49,152** | already 24 whole vectors: passed through |
| 50,000 | **51,200** | the bug |
| 51,200 | 51,200 | |
| 100,000 | 100,352 | matches the figure `check` already quotes for `convert` |
| 122,880 | 122,880 | DuckDB's own default |

The direction matters: because the writer rounds **up**, it can only push a request out through the **top** of a band, never through the bottom. That asymmetry is what picks the alignment rule. `TestWriterVectorAlignment::test_the_writer_rounds_a_request_up_to_a_whole_vector` pins this so the premise cannot rot silently.

### Nearest, not down — this is the part worth reviewing

The brief said round *down*. Rounding down fixes the top of the band and **breaks the bottom**: a user who types `--row-group-size 10000`, the band's own floor, would get `10000 → 8,192` and the identical `[fail]`, one end further along. That is the same defect, re-created.

Snapping to the **nearest** vector is monotonic and maps both endpoints of the band inside it (10,000 → 10,240 and 50,000 → 49,152), so by monotonicity every request between them lands inside it too. It also reaches the value the brief asked for: the default is still 49,152. `test_every_request_inside_the_spatial_band_stays_inside_it` walks the band and asserts the property; the docstring on `align_to_writer_vector` records the counterexample so nobody "simplifies" it back to a floor.

For realistic values, nearest and DuckDB's ceiling mostly agree (10,000 → 10,240, 20,000 → 20,480, 100,000 → 100,352 under both). They differ only where the fractional part is below a half.

### Why not the other candidates

- **Raise `SPATIAL_ROW_COUNT_RANGE` to 51,200.** Moves a band that has measurement behind it (#775) to accommodate a rounding artefact, and changes `check optimization` scores for published files. The band is the measured thing; it should not move.
- **Compare against the band rounded the writer's way.** Same effect as raising it, but the check would then print one band and enforce another — the exact sin #958 was fixing.
- **Score against `GENERAL_ROW_COUNT_RANGE`.** Would make the *optimization* score's row-group factor pass a 200,000-row-group file, deleting the only place the measured spatial evidence drives a score.
- **Lower only `DEFAULT_SORT_ROW_GROUP_ROWS` to a literal 49,152.** Fixes gpio's own default but leaves `--row-group-size 50000` at 51,200 and still `[fail]`ing — a user following gpio's own printed advice would hit the bug.

### Do the verdict and the score factor share a constant?

**They deliberately use different ones, and both are single-sourced.** `check row-group`'s pass/fail verdict is `GENERAL_ROW_COUNT_RANGE` (10,000–200,000); `check optimization`'s row-group score factor is `SPATIAL_ROW_COUNT_RANGE` (10,000–50,000), imported from `check_parquet_structure` rather than re-spelled. That split is #958's design and it is intact. No drift found.

One line did assume the two coincided: the printed guideline said `gpio sort defaults to {spatial_high}`, true only while the default equalled the band's top. It now names `DEFAULT_SORT_ROW_GROUP_ROWS`.

## Two consequences beyond the row-group layout

- **`sort str` strips and row groups now always line up.** `--row-group-size` does double duty there, and both uses take the aligned value. They previously lined up only when you passed a multiple of 2,048, which `docs/guide/sort.md` told users to arrange by hand. On realistic values the strip count is a rounded square root and does not move; only degenerate requests (`row_group_rows=20`) reorder, which is why `test_str_order_file_write_preserves_the_in_memory_ordering` needed updating.
- **`DEFAULT_STR_TILE_SIZE` tracks the sort default**, so `ops.sort_str` and `Table.sort_str` now default to `tile_size=49152`. The two `KNOWN_PARITY_GAPS` entries that recorded this default are derived from the constant instead of hard-coding it, so the next move updates them automatically.

Marked `!` because sorted output differs byte-for-byte from 1.4.0 for identical input, as #797 was.

## Verification

**The exact repro from the issue**, 102,400 random points:

```
# before (a 50,000 request, rounded up by the writer)
rows/group: 51200.0
Score: 2/5
  [fail] Row Group Size: Average 51,200 rows per group (optimal 10k-50k for spatial queries)

# after — bare default
$ gpio sort hilbert n102400.parquet out.parquet
rows/group: 34133.3        # 3 groups: 49,152 + 49,152 + 4,096
Score: 3/5
  [pass] Row Group Size: Average 34,133 rows per group (optimal 10k-50k for spatial queries)

# after — explicit --row-group-size 50000, the case the default fix alone would miss
Writing 49,152 rows per row group rather than the requested 50,000: …
Score: 3/5
  [pass] Row Group Size: Average 34,133 rows per group (optimal 10k-50k for spatial queries)
```

- Fast suite: `uv run pytest -n auto -m "not slow and not network and not meta" -q` → **6773 passed, 75 skipped, 9 xfailed** in 166s.
- Meta lane: `uv run pytest -m meta -q` → **205 passed**, 7600 deselected.
- `uv run pre-commit run --all-files` → clean; `--hook-stage pre-push` → clean, all 27 hooks including `menard-check`, `xenon`, `import-linter`, `mypy`, `vulture` and `deptry`.
- Diff coverage: `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=90` → **100%** (13 changed lines, 0 missing).

## Out of scope, reported not fixed

Two `cli/` docstrings are now stale, and `cli/` is outside this change's scope:

- `cli/commands/sort.py:153-154` — the `sort str` help still says strips and row groups "only line up when `--row-group-size` is itself a multiple of 2048". They now always line up.
- `cli/decorators.py:209` — a docstring illustration reads `@row_group_options(default_rows=50_000)`.

Also untouched, and still open from #795: `core/check_fixes.py` hard-codes `row_group_rows=100000` in five places, and `fix_spatial_order` does not inherit the sort default — so `check --fix` and `gpio sort hilbert` still write different layouts.

Closes #961

## Summary by CodeRabbit

- **New Features**
  - Sort operations now align row-group and tile sizes to Parquet’s 2,048-row writer vectors.
  - The default sort row-group size is now 49,152 rows.
  - Explicit row-group values are adjusted to valid vector boundaries, with informational notices when adjusted.
  - Invalid values below one row are rejected.

- **Documentation**
  - Updated CLI help, guides, API references, and examples to explain row-group alignment and command-specific behavior.

- **Tests**
  - Expanded coverage for alignment, defaults, validation, logging, and end-to-end behavior.


cholmes:
## Multi-agent adversarial review — two S2s found and fixed in 8226856

Reviewed in an isolated worktree rebased onto current `main`; the rebase was verified to introduce nothing (`git diff` of the rebased pair is byte-identical to `git diff` of the base pair). Every finding below I reproduced myself.

### S2 — "nearest" re-created #961 *below* the band, a regression against main

The writer's behaviour is a strict **ceiling** to a multiple of 2,048 — independently re-measured at 25 points, exact at every one, with `duckdb.__standard_vector_size__ == 2048`. `nearest` diverges from that whenever the fractional part is under 0.5, and in that window it rounds **down**. For `8,193 ≤ n ≤ 9,215` it returned 8,192 where the writer would have produced 10,240, moving the value from inside both bands to below the 10,000 floor:

```
--row-group-size 9000, 102,400 rows
  main:   groups 10, avg 10,240  ->  [pass]
  PR:     groups 13, avg  7,877  ->  [fail] "Re-partition with 10k-50k"
  fixed:  groups 10, avg 10,240  ->  [pass]
```

That is the exact defect this PR exists to remove, relocated from above the band to just under it.

**The fix rejected the rule I suggested, correctly.** I proposed "ceil unless the ceiling exceeds the band top, then floor". Unscoped, that sends 100,000 → 98,304 and 200,000 → 198,656 — dragging every above-band request back toward the band, and contradicting the very docs correction I asked for in the same brief. The exception is instead scoped to requests already at or below the band top:

| request | main | this PR (before) | fixed |
|---|---|---|---|
| 3,000 | 4,096 | 2,048 | **4,096** |
| 9,000 | 10,240 | **8,192** | **10,240** |
| 49,152 | 49,152 | 49,152 | **49,152** |
| 50,000 | 51,200 | 49,152 | **49,152** |
| 100,000 | 100,352 | 100,352 | **100,352** |
| 200,000 | 200,704 | 200,704 | **200,704** |

Zero regressions against main at any input I tested. **49,153–50,000 is now the only range where gpio writes fewer rows than asked**, and the docstring says so out loud rather than arguing the general case. `DEFAULT_SORT_ROW_GROUP_ROWS` is now *derived* as `align_to_writer_vector(SPATIAL_BAND_TOP_ROWS)` rather than typed, with a test asserting the local band top equals `SPATIAL_ROW_COUNT_RANGE[1]` — the import direction forbids importing it back (`check_parquet_structure` imports `parquet_writer`), so the constant is pinned by test instead.

A new test walks from 1 to the band floor asserting nothing is ever rounded down. That range was the structural blind spot: `test_every_request_inside_the_spatial_band_stays_inside_it` only walked `[10_000, 50_000]`.

### S2 — `--help` said the opposite of the doc this PR wrote

```
gpio sort str --help : "...only line up when --row-group-size is itself a multiple of 2048"
docs/cli/sort.md     : "...line up whatever you pass"
```

Two user-facing surfaces, one PR, flatly contradictory — the same failure #958 was a whole PR about. Both now agree, and the shared `--row-group-size` help no longer says "**Exact**" (still true for `convert`, false for `sort`); each branch states which behaviour applies.

Worth recording that **`docs/cli/*` is not auto-generated** — `doc-sync` covers only CLAUDE.md / SKILL.md / contributing.md — which is exactly why a green pre-commit run did not catch the contradiction. My brief was wrong to assume a hook would regenerate it.

### Also fixed

Stale `50,000` in `api/ops.py` and `api/table.py` docstrings (both now name `DEFAULT_STR_TILE_SIZE`, and `table.py`'s `Args:` no longer contradicts its own `Note:` four lines below); the `< 1` guard hoisted into `resolve_sort_row_group_rows` so all four subcommands reject `-5` and `0` identically — verified on each — where before `sort hilbert` silently clamped garbage to 2,048 while `sort str` raised; a docs regression in `docs/guide/sort.md` (strips are not tiles: `_str_layout(250_000, 100_352)` gives `strip_size=200_704`); and the four S4s, including `skills/geoparquet.md` still advising `--row-group-size 50000`.

### What held up

The measured rounding table in the body is correct at every point. All four sort subcommands genuinely route through one place. The `info()` is not `--verbose`-gated and goes to stderr when stdout is piped, so it cannot corrupt a stream. The `!` is right — sorted output changes byte-for-byte *and* the public API default moves. And the **#958 property survives**: an `assess_row_count` sweep over 4,528 combinations against `origin/main` found **0 differing `(status, message, colour)` triples**.

My "is snapping an explicit value over-reach?" concern was answered well: no request off a 2,048 boundary is achievable, so honouring intent cannot mean writing the number typed. Announcing the substitution is the right compromise.

### Filed as follow-ups

- **#971** — `Table.write(row_group_rows=50000)` still writes 51,200, so the same number means 49,152 for `gpio sort` and 51,200 for `gpio convert` and the Python API. Correctly out of scope for a `fix(sort)!` change; `Table.write` is the shared path for every command.
- **#972** — `gpio check --fix` writes 100,352-row groups and `check optimization` then fails the file it just fixed. The sharper instance of #961, since `--fix` is explicitly remedial. `hilbert_order.py` already imports `DEFAULT_SORT_ROW_GROUP_ROWS`, so the swap is a one-liner.

### Gates

Fast suite **6927 passed, 75 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100%** · `pytest -m meta` 205 passed · docs lane **236 passed, 221 skipped** · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push all clean · `cz check` passes on the title · tests 33 → 53.

Title set to name the number, since it becomes the changelog's Breaking entry.

Issue #731, opened 2026-08-23
extract silently accepts unknown names in columns=[...] and fails later in the writer with a KeyError naming the wrong column


cholmes:
`extract` accepts column names that do not exist in the input without complaint. The run fails much later, inside the writer, with a `KeyError` naming a column that *does* exist — which sends the reader looking in the wrong place.

## Reproduce

`tests/data/canonical/places.parquet` has columns `fsq_place_id, name, address, placemaker_url, geometry, bbox`. Neither `id` nor `population` exists.

```python
import geoparquet_io as gpio

gpio.read('places.parquet').extract(
    columns=['id', 'name', 'population'],
    exclude_columns=['geometry'],
).write('out.parquet')
```

```
File ".../geoparquet_io/core/write_strategies/duckdb_kv.py", line 622, in ...
    geom_type = table.schema.field(geometry_column).type
KeyError: 'Column geometry does not exist in schema'
```

The error names `geometry`, which the file has and the caller deliberately excluded. The two genuinely bogus names, `id` and `population`, are never mentioned.

Verified twice independently, from two different guide examples in `docs/guide/extract.md`.

## Expected

Validate `columns=[...]` against the input schema and fail early, naming the columns that were not found — the same treatment `--include-cols` already gives on the CLI:

```
Error: Invalid value for --include-cols: Columns not found in schema: invalid_column.
Available columns: fsq_place_id, name, address, ...
```

Two things are arguably in play: the missing up-front validation, and the writer requiring a geometry column the caller asked to drop (excluding geometry to produce an attribute table is a documented use case). The second may be a separate decision.

Found by the documentation-examples test harness (#667). Both blocks are opted out with a reason naming the writer `KeyError`.

Refs #667

PR #973, opened 2026-09-10
fix(extract): reject blank --include-cols/--exclude-cols entries and check BigQuery names against the table

### LINKED ISSUE #969: fix(extract): bigquery --include-cols/--exclude-cols quote unvalidated column names
## What

The BigQuery extract path has no equivalent of the parquet backend's `validate_columns`. Entries from `--include-cols` / `--exclude-cols` flow straight to quoting:

`extract_bigquery.py:890` → `_build_column_list:678` → `_build_select_clause:595`

so a blank entry survives to the quoting call and raises raw:

```
$ gpio extract bigquery ... --include-cols '   '
   -> ['']  ->  ValueError: cannot quote an empty SQL identifier
```

A name that is non-blank but simply absent from the table is likewise never checked against the schema — the parquet path rejects that with a clean `InvalidParameterError`, the BigQuery path does not.

## Why it was not fixed in #959

#959 completed the #933 sweep for options that carry **a single column name** and become a delimited identifier, by tracing data flow to `quote_identifier()` rather than matching flag spellings. This site is a different shape and needs a different fix:

- it is a **comma-separated list**, so a Click callback validating "the value" is the wrong tool — each entry needs checking;
- the real defect is **missing schema validation**, not a blank-value guard, and validating properly needs a live BigQuery schema, which the CLI layer does not have.

Per the repo's `separate-prs-for-cross-cutting-fixes` convention it was left out rather than bolted on.

## Suggested fix

Give the BigQuery path a `validate_columns` equivalent that runs after the table schema is known, rejecting blank entries and names absent from the schema with `InvalidParameterError` (which `handle_core_exception` already maps to exit 2, matching the parquet backend). Worth checking `--where` and `--geography-column` on the same path while there.

Found while completing the #933 sweep in #959.

cholmes:
## What changed

**The Click layer rejects blank entries; the backend rejects names the table
does not have.**

`cli/decorators.py` gains `_validate_column_list`, a Click callback, and
`column_list_option()`, a thin `click.option` wrapper — the list-valued sibling
of #959's `_validate_column_name` / `column_name_option()`. Nine option
declarations switch from `@click.option` to `@column_list_option`:
`--include-cols` and `--exclude-cols` on all four `gpio extract` subcommands
that have them (`geoparquet`, `arcgis`, `bigquery`, `carto`), plus
`gpio pmtiles create --include-cols`. No default, help text or type changes.

`core/extract_bigquery.py` gains `validate_bigquery_columns()`, the BigQuery
equivalent of the parquet backend's `validate_columns`, run in
`_execute_bigquery_extraction` once the schema is known and before anything is
quoted into the SELECT. `_schema_column_names()` factors out the `DESCRIBE`
that three call sites already duplicated, and `_build_column_list` takes the
schema the validator already read, so `--exclude-cols` costs no extra
round-trip against BigQuery.

`tests/test_cli_column_list_guard.py` walks the live command tree and fails if a
new `--*-cols` option appears without the guard, the way
`test_cli_column_name_guard.py` does for the single-name family.

## Why

**Which layer owns which check.** A blank entry is a *usage* error — nothing
about the data can make `--include-cols '   '` meaningful — and it is catchable
before a file is opened or a service contacted, so it belongs at the CLI.
Whether a non-blank name exists is a *schema* question, so it belongs to
whichever backend can see the schema. They are not redundant:

- `--dry-run` builds the same SELECT without ever opening a connection, so on
  that path there is no schema to check against and only the Click guard can
  fire;
- the Python API (`ops.read_bigquery`, `Table.from_bigquery`) never goes through
  Click, so `validate_bigquery_columns` rejects blanks too.

**Why #959 did not cover this.** #959 closed the #933 sweep for options carrying
*one* column name, with a callback that validates *the value*. Here the value is
`"a,b,c"` and each entry needs checking — `"id,,name"` is not blank but carries
a blank entry. So the fix is a second callback of the same shape, not an
extension of the first.

**Error type and exit code follow the parquet backend.**
`core/extract.py:validate_columns` has raised `InvalidParameterError` since
#731, and `handle_core_exception` maps it to `click.BadParameter` → exit 2; the
new BigQuery validator matches it, message shape included ("Columns not found in
schema: … Available columns: …"). The Click guard raises `click.BadParameter`
directly — also exit 2 — which is what `_validate_column_name` and
`_validate_write_memory` in the same module already do.

**The other extract subcommands shared the gap, so they are fixed here.**
Tracing each `--include-cols`/`--exclude-cols` to where its entries land:

| Subcommand | Where a blank entry went | Fixed by |
|---|---|---|
| `bigquery` | `quote_identifier()` via `_handle_dry_run` / `_build_select_with_wkb` — raw `ValueError`, exit 1 | Click guard + schema check |
| `carto` | `quote_identifier()` via `_build_carto_query` — same `ValueError`, one network round-trip later | Click guard |
| `arcgis` | `outFields` sent to the server as `"name,,pop"` | Click guard |
| `geoparquet` | already clean: `validate_columns` reports it as "not found in schema" | Click guard now catches it earlier, with a message that says what is wrong |
| `pmtiles create` | forwarded verbatim to a `gpio extract` subprocess, so it failed one process away from the user | Click guard |

`extract wfs` has no such option. `--where` on the BigQuery path is already
routed through `validate_where_clause` before any query is built (#612), and
`--geography-column` is already resolved against the schema first
(`Column '   ' not found in table`), which is why both stay in
`REVIEWED_UNGUARDED` in the sibling test; neither needed a change.

**One side effect worth naming.** Names are resolved to their schema spelling,
because DuckDB matches a quoted identifier case-insensitively but
`_build_column_list` filters `--exclude-cols` by string comparison — so
`--exclude-cols ID` against a column named `id` silently excluded nothing.
Resolving makes the two agree.

## Verification

Before, with no credentials needed:

```
$ gpio extract bigquery my-project-x.ds.tbl out.parquet --include-cols '   ' --dry-run
  File ".../extract_bigquery.py", line 611, in <genexpr>
    select_cols = ", ".join(quote_identifier(c) for c in include_list)
ValueError: cannot quote an empty SQL identifier: DuckDB rejects the
zero-length delimited identifier ""                        # exit 1, traceback
```

After:

```
$ gpio extract bigquery my-project-x.ds.tbl out.parquet --include-cols '   ' --dry-run
Error: Invalid value for '--include-cols': column list cannot contain an empty
or whitespace-only entry                                   # exit 2

$ gpio extract bigquery my-project-x.ds.tbl out.parquet --include-cols 'id,,name' --dry-run
Error: Invalid value for '--include-cols': column list cannot contain an empty
or whitespace-only entry                                   # exit 2

$ gpio extract bigquery my-project-x.ds.tbl out.parquet --include-cols 'id,name' --dry-run
SQL: SELECT "id", "name" FROM bigquery_scan('my-project-x.ds.tbl')
                                                           # exit 0, unchanged

$ gpio extract carto https://phl.carto.com tbl out.parquet --include-cols '  '
Error: Invalid value for '--include-cols': column list cannot contain an empty
or whitespace-only entry                     # exit 2, no network round-trip

$ gpio extract geoparquet buildings.parquet out.parquet --include-cols 'id, ,x'
Error: Invalid value for '--include-cols': column list cannot contain an empty
or whitespace-only entry                                   # exit 2

$ gpio extract geoparquet buildings.parquet out.parquet --include-cols id
Extracted 42 rows (out of 42 total) to out.parquet         # exit 0, unchanged
```

- Fast suite: `pytest -n auto -m "not slow and not network and not meta"` —
  **7095 passed, 76 skipped, 9 xfailed**. No reruns needed; green first time.
- Meta lane: `pytest -m meta` — **205 passed**.
- `pre-commit run --all-files` clean; `--hook-stage pre-push` clean (deptry,
  mypy, vulture, xenon, import-linter, menard-check, fast-tests).
  **40 changed lines, 0 missing, 100%**.

### A note on `menard-check`

Running `pre-commit --all-files --hook-stage pre-push` with everything staged
reports 10 stale doc targets. That is **not** caused by this branch: `menard
check` inspects *staged* files, and staging a one-line no-op comment in
`geoparquet_io/cli/commands/extract.py` on a **clean `origin/main` checkout**
(detached worktree at 5ce725e) reproduces it exactly — four stale targets
attributed to `5ce725e`, `725b1d1` and `d40eca2`, all doc drift those commits
left behind. Against the real commits the hook passes, which is why nothing is
marked reviewed and no hook was bypassed here.

### `tests/data/cli_surface.json`

Unchanged. This branch only adds a `callback=` to existing options, which the
snapshot does not record — the same reason #959 needed no snapshot entry for its
`--*-name` guard.

Closes #969

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for comma-separated column lists across extraction commands.
  * Blank, whitespace-only, and invalid column names now produce clear command-line errors instead of database-related tracebacks.
  * BigQuery column validation now supports case-insensitive matching while preserving schema spelling.
  * Invalid column selections are rejected before queries or external services are accessed.

### DIFF
--- a/geoparquet_io/cli/commands/extract.py
+++ b/geoparquet_io/cli/commands/extract.py
@@ -18,6 +18,7 @@ from geoparquet_io.cli.decorators import (
     SingleFileCommand,
     any_extension_option,
     aws_profile_option,
+    column_list_option,
     compression_options,
     dry_run_option,
     geoparquet_version_option,
@@ -69,11 +70,11 @@ def extract(ctx):
 @extract.command(name="geoparquet", cls=GlobAwareCommand)
 @click.argument("input_file")
 @click.argument("output_file", type=click.Path(), required=False, default=None)
-@click.option(
+@column_list_option(
     "--include-cols",
     help="Comma-separated columns to include (geometry and bbox auto-added unless in --exclude-cols)",
 )
-@click.option(
+@column_list_option(
     "--exclude-cols",
     help="Comma-separated columns to exclude (can be used with --include-cols to exclude geometry/bbox)",
 )
@@ -286,11 +287,11 @@ def extract_geoparquet(
     "--bbox",
     help="Bounding box filter: xmin,ymin,xmax,ymax in WGS84 (pushed to server)",
 )
-@click.option(
+@column_list_option(
     "--include-cols",
     help="Comma-separated columns to include (pushed to server for efficiency)",
 )
-@click.option(
+@column_list_option(
     "--exclude-cols",
     help="Comma-separated columns to exclude (applied after download)",
 )
@@ -508,11 +509,11 @@ def extract_arcgis(
     help="Path to GCP service account JSON file (otherwise uses gcloud auth or "
     "GOOGLE_APPLICATION_CREDENTIALS)",
 )
-@click.option(
+@column_list_option(
     "--include-cols",
     help="Comma-separated columns to include",
 )
-@click.option(
+@column_list_option(
     "--exclude-cols",
     help="Comma-separated columns to exclude",
 )
@@ -1052,11 +1053,11 @@ def extract_wfs_cmd(
     type=click.IntRange(0, None),
     help="Maximum number of rows to extract",
 )
-@click.option(
+@column_list_option(
     "--include-cols",
     help="Comma-separated columns to include (default: all)",
 )
-@click.option(
+@column_list_option(
     "--exclude-cols",
     help="Comma-separated columns to exclude",
 )
--- a/geoparquet_io/cli/commands/pmtiles.py
+++ b/geoparquet_io/cli/commands/pmtiles.py
@@ -10,6 +10,7 @@ import click
 from geoparquet_io.cli.decorators import (
     SingleFileCommand,
     aws_profile_option,
+    column_list_option,
     repair_geometry_option,
     verbose_option,
 )
@@ -42,7 +43,7 @@ def pmtiles(ctx):
 @click.option("--max-zoom", type=int, help="Maximum zoom level (auto-detected if not set)")
 @click.option("--bbox", help="Bounding box filter: minx,miny,maxx,maxy")
 @click.option("--where", help="SQL WHERE clause for filtering")
-@click.option("--include-cols", help="Comma-separated list of columns to include")
+@column_list_option("--include-cols", help="Comma-separated list of columns to include")
 @click.option(
     "--precision",
     type=int,
--- a/geoparquet_io/cli/decorators.py
+++ b/geoparquet_io/cli/decorators.py
@@ -354,8 +354,19 @@ def _validate_column_list(ctx, param, value):
     BigQuery), which reports it as ``InvalidParameterError`` -- also exit 2.
     Catching blanks here matters even so: ``--dry-run`` builds the same SELECT
     without ever opening a connection, so there is no schema to check against.
+
+    A **wholly empty** value is not a blank entry, it is an unset option, and is
+    normalised to ``None``. Every backend reads these options as
+    ``[c.strip() for c in v.split(",")] if v else None``
+    (``core/extract.py``, ``core/extract_bigquery.py``, ``core/carto.py``,
+    ``core/pmtiles.py``), so ``""`` has always meant "not given" -- which is what
+    ``--include-cols "$COLS"`` expands to when ``COLS`` is unset. Rejecting it
+    would break working invocations without catching anything: ``"".split(",")``
+    is ``[""]`` only because there is nothing there to name. ``"   "`` and
+    ``"id,,name"`` are a different case -- both are truthy, so both reach
+    ``quote_identifier()`` on main, and both stay rejected.
     """
-    if value is None:
+    if value is None or value == "":
         return None
     if any(not entry.strip() for entry in value.split(",")):
         raise click.BadParameter(
--- a/geoparquet_io/cli/decorators.py
+++ b/geoparquet_io/cli/decorators.py
@@ -333,6 +333,51 @@ def column_name_option(*param_decls, **kwargs):
     return click.option(*param_decls, **kwargs)
 
 
+def _validate_column_list(ctx, param, value):
+    """Reject a blank entry in an option carrying a comma-separated column list.
+
+    The list-valued sibling of :func:`_validate_column_name`. ``--include-cols``
+    and ``--exclude-cols`` are split on commas by each backend, and the entries
+    become delimited SQL identifiers: ``gpio extract bigquery ... --include-cols
+    '   '`` split to ``['']`` and raised a raw
+    ``ValueError: cannot quote an empty SQL identifier`` traceback out of
+    ``quote_identifier()`` (#969); ``gpio extract carto`` hit the same call one
+    network round-trip later.
+
+    #959 could not cover these with :func:`_validate_column_name` because that
+    callback validates *the value*, and here the value is a list -- ``"id,,name"``
+    is not blank but carries a blank entry.
+
+    This layer owns blank entries only. Whether a non-blank name exists is a
+    schema question, so it stays with whichever backend can see the schema
+    (``validate_columns`` for parquet, ``validate_bigquery_columns`` for
+    BigQuery), which reports it as ``InvalidParameterError`` -- also exit 2.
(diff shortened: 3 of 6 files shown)


cholmes:
## Multi-agent adversarial review — one S2 found and fixed in 1359f41

### S2 — a previously-working invocation started failing

Every backend reads these options as `[c.strip() for c in v.split(",")] if v else None`, so **an empty string meant "unset"**. The new callback rejected it, because `"".split(",") == [""]`:

```
gpio extract geoparquet places_test.parquet out.parquet --exclude-cols ""
  main:            Extracted 766 rows (out of 766 total)   exit 0
  as submitted:    Error: Invalid value for '--exclude-cols': ...   exit 2
```

That breaks `--include-cols "$COLS"` with an unset variable — a common shell idiom — across all nine declarations, and the body said "No default, help text or type changes".

Fixed by returning `None` for a wholly empty value, which restores main's own semantics. Only exactly `""` is falsy, so the actual #969 defects are untouched:

```
--include-cols ''            -> exit 0  Extracted 766 rows
--include-cols '   '         -> exit 2  rejected
--include-cols 'id,,name'    -> exit 2  rejected
--include-cols ','           -> exit 2  rejected   (newly covered)
```

### The body claimed a factoring it had not done — so the factoring was done

The body said `_schema_column_names()` "factors out the `DESCRIBE` that three call sites already duplicated". It had converted exactly one, and its body was a verbatim copy of two others — a fifth copy, not a removal. All five inline `DESCRIBE SELECT * FROM bigquery_scan(...)` sites now go through one `_schema_rows()` helper, and `_execute_bigquery_extraction` reads the schema once for both geometry detection and the column check. That turned a regression into an improvement:

| | no col option | `--include-cols` | `--exclude-cols` |
|---|---|---|---|
| `origin/main` | 3 | 2 | 3 |
| as submitted | 3 | **3** | 3 |
| now | 3 | **2** | **2** |

- **The ratchet does not have #959's hole.** I expected the decorator-ordering bug that #959's first attempt had. Sabotaged both ways — an unguarded `--zzz-cols` above *and* below a guarded option — and it failed identically both times. `_column_list_options()` already uses the `setdefault(...).extend(...)` accumulate fix.
- **The `--exclude-cols` round-trip claim held** under measurement (3 DESCRIBE either way).
- **The sweep is complete at the data-flow level.** I asked for a trace of every list-shaped value, not just `--*-cols`: `sort column`'s positional `COLUMNS` validates each entry against the schema on all three paths; `process aggregate --metric` skips blanks and rejects a missing column (probed live: `--metric 'sum:   '` → `Metric 'sum:' is missing a column name`); `--levels`/`--converters`/`--metric-nodata` are not identifiers.

What the ratchet's *self-description* overclaimed is now recorded: `LIST_VALUED_OUTSIDE_THE_SWEEP` names the traced values the `--*-cols` filter structurally cannot see, with a test pinning that those parameters still exist so a rename fails loudly rather than silently shrinking coverage.

### Also taken

The raw `{table_id}` interpolation now exists once, in `_schema_rows`, whose docstring states the `_normalize_table_id`/`_TABLE_PART_PATTERN` pre-validation contract and why `sql_path()`/`quote_identifier()` do not apply (`bigquery_scan` takes a string literal). `validate_bigquery_columns` prefers an exact match to a case-folded one, so a local table carrying both `id` and `ID` resolves each to itself.

### Filed separately — #980

Carto and ArcGIS keep the exact defect #969 names, in the Python API. The PR's own rationale for a core-level BigQuery validator ("the Python API never goes through Click") applies verbatim to both, and they got only the Click guard. Carto **crashes**: `ops.from_carto(..., include_cols="id,,name")` → `carto.py:609` → `ValueError: cannot quote an empty SQL identifier`, after a network round-trip. ArcGIS sends `outFields=name,,pop` to the server.

### Gates

Fast suite **7117 passed, 76 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100% on 66 lines** · `pytest -m meta` 205 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push clean · **`cli_surface.json` unchanged and `--help` byte-identical to `origin/main` for all nine declarations**, re-verified after the fix.

Tests: guard file 66 → 77 (`""` moved out of the reject set into a new `TestWhollyEmptyMeansUnset` covering all nine declarations plus an end-to-end round trip); BigQuery file 75 → 78. All fail against `origin/main`; both modules are fast-lane, so CI gates them.

PR #974, opened 2026-09-10
fix(check): write 49,152-row groups from --fix, and anchor two unanchored hook exemptions

### LINKED ISSUE #970: fix(hooks): two pre-commit guards carry unanchored exemptions that silence more than intended
## What

Two guards in `.pre-commit-config.yaml` filter their own output with an **unanchored substring** match, so they exempt far more than the one module they name:

```yaml
:179  grep -rn "click\.echo" geoparquet_io/core/ --include="*.py" | grep -v "logging_config.py"
:198  grep -rnE 'def _(align|unify|reconcile)[a-zA-Z_]*schema' geoparquet_io/core/ --include="*.py" | grep -v "common.py"
```

`grep -v` sees the whole `path:line:text` line, not just the path field. Two consequences, both reproduced:

**1. A comment silences the guard.** Verified against a synthetic tree with the real command:

```
clean.py   def _align_schema(a, b):                        -> FLAGGED  (correct)
evil.py    def _align_schema(a, b):  # mirrors common.py   -> SILENCED
```

The same holds for `no-click-echo`: `click.echo("hi")  # migrate this to logging_config.py later` is silenced.

**2. `grep -v "common.py"` exempts four modules, not one.** Because it is also a substring match on the path:

```
geoparquet_io/core/common.py                        (intended)
geoparquet_io/core/partition/common.py
geoparquet_io/core/process/aggregate/common.py
geoparquet_io/core/process/aggregate/grid_common.py
```

Three real modules are silently outside the `forbid-bespoke-schema-reconciliation` guard.

## Why it matters

This is the same shape as #946, where a file-scoped `grep -v` in `duckdb-antipatterns` silently shielded three unrelated DuckDB sites and took a security review of a downstream PR to find. A guard with a hole reads as protection while providing none.

#964 pinned the `duckdb-antipatterns` guard against exactly this and demonstrated the technique — assert the arm's pipeline stage by stage, so any added or reworded filter fails whatever spelling it uses. These two siblings are unpinned and, unlike that one, currently *have* the hole.

## Fix

1. **Anchor both to the path field**: `grep -v '^geoparquet_io/core/logging_config.py:'` and `grep -v '^geoparquet_io/core/common.py:'`. That fixes the comment bypass and narrows the over-exemption in one change.
2. Check whether the three newly-un-exempted modules (`partition/common.py`, `process/aggregate/common.py`, `grid_common.py`) actually contain matches — if they do, decide per site whether it is a real violation or wants a per-line hatch, rather than restoring a broad exemption.
3. Give both guards the same self-test treatment as `tests/test_duckdb_antipatterns_hook.py`, so a future exemption cannot go in unnoticed. That file's `_manual_quote_stages()` / `_EXPECTED_STAGES` pattern is directly reusable; the duplication between it and `tests/test_duckdb_connection_factory_bypass.py` suggests a shared `tests/_precommit_hook.py` helper is worth extracting at the same time.

Worth sweeping the rest of `.pre-commit-config.yaml` for other `grep -v` filters while in there.

Found by an adversarial review of #964.

### LINKED ISSUE #972: fix(check): gpio check --fix writes row groups its own check optimization fails
## What

`core/check_fixes.py` hardcodes `row_group_rows=100000` at five sites (`:97`, `:155`, `:240`, `:313`, `:360`). The Parquet writer rounds that up to **100,352**, which is inside `GENERAL_ROW_COUNT_RANGE` (10,000–200,000) but well outside the 10,000–50,000 `SPATIAL_ROW_COUNT_RANGE` that `check_optimization` scores as one of its five factors.

So a `--fix` run produces a file whose optimization score still flags the row-group factor and advises re-partitioning:

```
$ gpio check all bigfix.parquet --fix
Optimized file: bigfix.parquet
$ gpio check optimization bigfix.parquet
Score: 2/5
  [fail] Row Group Size: Average 100,000 rows per group (optimal 10k-50k for spatial queries)
```

`gpio check --fix`, whose entire job is to make a file pass gpio's checks, writes a file gpio then fails. Same self-contradiction as #961, one command over — and arguably the sharper instance, since `--fix` is explicitly remedial.

## Scope

Pre-existing; #967 fixed the `sort` half of #961 and correctly left this out of a `fix(sort)!` change. #959 also touched `fix_spatial_ordering` (making the in-place path work) without changing the constant, on the same reasoning.

## Decide what `--fix` is for

- If it is a **general-purpose repair**, it should say so, and `check optimization` should stop scoring its output against the spatial band.
- If it is meant to leave a file **`check`-clean**, it should use `DEFAULT_SORT_ROW_GROUP_ROWS` (49,152) and route through `align_to_writer_vector` like the sort commands now do.

The second is what #795 asked for: *"fix_spatial_order should almost certainly inherit the sort default"*. Note `core/hilbert_order.py` already imports `DEFAULT_SORT_ROW_GROUP_ROWS`, so the swap is a one-liner with no new import edge — verified while reviewing #959.

Related: #961, #795, #967.

cholmes:
Two independent defects, one commit each.

## What changed

**#972 — `check --fix` wrote row groups its own check failed.**
`core/check_fixes.py` hardcoded `row_group_rows=100000` at all five of its write
sites. The Parquet writer emits whole 2,048-row vectors and rounds a request
*up*, so that landed at **100,352**: inside `GENERAL_ROW_COUNT_RANGE`
(10k–200k), which `check row-group` passes a file on, but outside
`SPATIAL_ROW_COUNT_RANGE` (10k–50k), which `check optimization` scores as one of
its five factors. All five sites now ask for `DEFAULT_SORT_ROW_GROUP_ROWS`
(**49,152**).

**Which band, and why the spatial one.** `--fix` is remedial, so the file it
leaves behind has to clear the *stricter* of the two bands — writing into the
general band leaves the score exactly as `--fix` found it. `--fix` is also
already committed to that band: it Hilbert-sorts the file, and sorting exists to
make spatial predicates prune row groups, so groups too large to prune would
undo what the sort just bought. And of the two bands, only the spatial one has a
measurement behind it (#775); `GENERAL_ROW_COUNT_RANGE`'s top is inherited, as
its own comment says.

**No second helper, and nothing typed.** `DEFAULT_SORT_ROW_GROUP_ROWS` *is*
`align_to_writer_vector(SPATIAL_BAND_TOP_ROWS)` — the constant and the snapping
helper #967 gave `gpio sort` — so `check --fix` and `gpio sort` cannot drift
apart, and neither can drift from the band. That also settles the open half of
#795: `fix_spatial_ordering` asked `hilbert_order` for 100,000 while
`gpio sort hilbert` wrote 49,152, so the two spellings of one operation
disagreed. They now agree, and the request no longer trips `hilbert_order`'s own
"consider a smaller `--row-group-size`" advice. The fix summary line said
`Optimized row groups (100k rows/group)`, a size no file ever had; it now names
what was written.

**#970 — two pre-commit guards exempted more than they named.** `no-click-echo`
and `forbid-bespoke-schema-reconciliation` both filtered their output with a
bare substring `grep -v`. Both now use the anchored form #964 settled on,
`grep -v '^<path>:'`, and both get self-tests that run the real hook script over
a synthetic tree. The three hook self-test modules now share
`tests/_precommit_hook.py` rather than a third copy of the same plumbing, which
#970 asked for.

## Why

`gpio check --fix`, whose entire job is to make a file pass gpio's checks, wrote
a file gpio then failed:

```
$ gpio check all bigfix.parquet --fix
Optimized file: bigfix.parquet
$ gpio check optimization bigfix.parquet
Score: 2/5
  [fail] Row Group Size: Average 100,000 rows per group (optimal 10k-50k for spatial queries)
```

And a guard with a hole reads as protection while providing none — the same
shape as #946, which took a security review of a downstream PR to find.
`grep -v` matches the whole `path:line:text` line, not just the path field, so
both guards had two holes:

1. **A comment silenced the guard.** Reproduced against the real hook scripts on
   a synthetic tree, before the change:

   ```
   some_module.py  def _align_schema(a, b):                        -> FLAGGED   (correct)
   some_module.py  def _align_schema(a, b):  # mirrors common.py   -> SILENCED
   some_module.py  click.echo("hello")                             -> FLAGGED   (correct)
   some_module.py  click.echo("hello")  # port to logging_config.py later -> SILENCED
   ```

2. **`grep -v "common.py"` exempted four modules, not one** — as a path
   substring it also covered `core/partition/common.py`,
   `core/process/aggregate/common.py` and `core/process/aggregate/grid_common.py`.
   Reproduced the same way: a violation planted at `core/partition/common.py`
   and at `core/grid_common.py` passed the guard before the change and fails it
   after.

Anchoring newly flags nothing and newly exempts nothing. Both guards report zero
hits over the real tree, before and after: the three un-exempted modules contain
no matching definitions, and `core/`'s only `click.echo` occurrences are prose
in `logging_config.py`'s own docstrings — which is why that module keeps its
(now path-scoped) exemption.

## Verification

- `uv run pytest -n auto -m "not slow and not network and not meta" -q` →
  **7053 passed, 76 skipped, 9 xfailed** in 316s.
- `uv run pytest -m meta -q` → **205 passed**, 7881 deselected.
- `uv run pre-commit run --all-files` → clean;
  `uv run pre-commit run --all-files --hook-stage pre-push` → clean
  (deptry, mypy, vulture, xenon, import-linter, menard-check, fast-tests all pass).
- `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=90`
  → **100%** (2 changed statements in `geoparquet_io/`, 0 missing).

New tests, both written failing first:

- `tests/test_check_fix_row_group_band.py` (11 tests). Every one of the five
  write sites is pinned to `DEFAULT_SORT_ROW_GROUP_ROWS`; a source scan fails if
  any site names a literal row count again; and the issue's own reproduction
  runs end to end through the real writer on a 110,000-row file — under the old
  constant it lands in two 100,352-row groups and `_check_row_group_size`
  returns `Average 55,000 rows per group … [fail]`, and after the fix three
  49,152-row groups average 36,667 and pass. That control test is kept, forcing
  the old value explicitly, so the passing test cannot be an artefact of the
  fixture.
- `tests/test_precommit_guard_exemptions.py` (20 tests). Both reproductions
  above, both controls (a violation is flagged; a clean module is not), the
  exempt module stays exempt, and the nested/`grid_`-prefixed shapes are not.
  Ten of these failed before the anchor. The structural half pins each guard's
  pipeline stage by stage rather than blacklisting known bad spellings — a
  blacklist is only as good as its list, and a whole-module exemption has many
  shapes (`grep -v` without the `^`, a double-quoted path, an `awk` regex match,
  an `--exclude=` on the search, the path in a shell variable). A separate test
  states the property directly, so a failure names it: every filter stage that
  mentions a module must anchor it as `'^<path>:'`.

**Noted, not fixed:** the sweep #970 suggested found the same class in three
arms of `duckdb-antipatterns` (config lines 67/72/77). `grep -v "^\s*#"` there
can never match — the line starts with the path, so it exempts nothing despite
reading as a comment filter — while `grep -v "DuckDB 1.5"` is a content match
that any line mentioning DuckDB 1.5 satisfies, comment or not. Those arms belong
to the hook #964 pinned and are worth their own change rather than being bolted
on here; happy to open an issue.

Closes #972
Closes #970

## Summary by CodeRabbit

* **Bug Fixes**
  * `gpio check --fix` now writes row groups sized to meet spatial optimization checks.
  * Optimization fix messages report the actual row-group size written.
  * Pre-commit guards now apply exemptions only to the intended files, preventing comments or similarly named modules from bypassing checks.

* **Documentation**
  * Updated the check guide to explain row-group sizing behavior and optimization results.

* **Tests**
  * Added coverage for row-group fixes and pre-commit guard exemptions.
  * Consolidated shared setup for pre-commit hook tests.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -183,7 +183,13 @@ repos:
         args:
           - -c
           - |
-            if grep -rn "click\.echo" geoparquet_io/core/ --include="*.py" | grep -v "logging_config.py"; then
+            # The exemption is anchored to the path field. `grep -v` sees the
+            # whole `path:line:text` line, so the bare `grep -v
+            # "logging_config.py"` this used to be was switched off by a comment
+            # that merely mentioned the module -- `click.echo("hi")  # port to
+            # logging_config.py later` was not flagged (#970). Same shape as
+            # #946, one hook over. Keep the `^...:` form.
+            if grep -rn "click\.echo" geoparquet_io/core/ --include="*.py" | grep -v '^geoparquet_io/core/logging_config.py:'; then
               echo "ERROR: Use logger instead of click.echo in core/. See CLAUDE.md Logging section."
               exit 1
             fi
@@ -202,7 +208,12 @@ repos:
         args:
           - -c
           - |
-            if grep -rnE 'def _(align|unify|reconcile)[a-zA-Z_]*schema' geoparquet_io/core/ --include="*.py" | grep -v "common.py"; then
+            # Anchored to the path field, for the reason above and for one
+            # more: as a substring, "common.py" also matched
+            # core/partition/common.py, core/process/aggregate/common.py and
+            # core/process/aggregate/grid_common.py, so three modules nobody
+            # meant to exempt were outside the guard (#970).
+            if grep -rnE 'def _(align|unify|reconcile)[a-zA-Z_]*schema' geoparquet_io/core/ --include="*.py" | grep -v '^geoparquet_io/core/common.py:'; then
               echo "ERROR: Reuse _compute_unified_schema / _cast_table_to_schema from core/common.py"
               echo "       instead of hand-rolling schema reconciliation. See the WFS/ArcGIS history"
               echo "       and docs/contributing.md (External extractors)."
--- a/docs/guide/check.md
+++ b/docs/guide/check.md
@@ -163,6 +163,8 @@ The check reports two numbers, and they answer different questions:
 
     Neither end of that band is a size a row group can actually have. The Parquet writer emits row groups in whole 2,048-row vectors and rounds a request *up* to a multiple of that, so a request for the band's top of 50,000 lands at 51,200 — just outside it. That used to be exactly what `gpio sort` asked for, so `gpio check optimization`, which scores this band as one of its five factors, marked a freshly sorted file `[fail]` and advised re-partitioning it ([#961](https://github.com/geoparquet/geoparquet-io/issues/961)). The band did not move; `gpio sort` now snaps its request to a whole vector itself, which maps the band's endpoints to 10,240 and 49,152, and writes 49,152 rows per group by default. It snaps up, as the writer does, except where that would push a request out of the top of the band — 50,000 becomes 49,152 rather than 51,200. Other write paths do not snap — if you pass `--row-group-size 50000` to `gpio convert`, the writer still rounds it up to 51,200.
 
+    `--fix` writes those same 49,152-row groups. It is a repair, so the file it leaves behind has to pass the checks gpio runs next, and this is the narrower of the two bands. Until [#972](https://github.com/geoparquet/geoparquet-io/issues/972) it asked for 100,000 rows per group — landing at 100,352, inside the general 10,000-200,000 band that `gpio check row-group` passes a file on but outside this one — so `gpio check optimization` scored a freshly fixed file `[fail]` on its row-group factor and advised re-partitioning it.
+
 ### Optimization Check
 
 === "CLI"
--- a/docs/guide/check.md
+++ b/docs/guide/check.md
@@ -165,6 +165,8 @@ The check reports two numbers, and they answer different questions:
 
     `--fix` writes those same 49,152-row groups. It is a repair, so the file it leaves behind has to pass the checks gpio runs next, and this is the narrower of the two bands. Until [#972](https://github.com/geoparquet/geoparquet-io/issues/972) it asked for 100,000 rows per group — landing at 100,352, inside the general 10,000-200,000 band that `gpio check row-group` passes a file on but outside this one — so `gpio check optimization` scored a freshly fixed file `[fail]` on its row-group factor and advised re-partitioning it.
 
+    **`--fix` follows this band, not the verdict band.** A file inside 10,000-200,000 but outside 10,000-50,000 still passes `gpio check row-group` — it is laid out the way mainstream writers lay files out — and `--fix` will still offer to rewrite it, because 49,152 rows per group is what it writes. Offering the fix off the verdict band instead meant `gpio check row-group --fix` printed "No fix needed" for a 100,352-row-group file while `gpio check optimization` on the same file printed `[fail] Row Group Size`; it also made the result depend on which subcommand you ran, since `check spatial --fix`, `check compression --fix` and `check bbox --fix` each rewrote the file for their own reasons and fixed the row groups in passing. The report says which band it is talking about, so a green "optimal for general use" line is followed by a note that `--fix` has something to do.
+
 ### Optimization Check
 
 === "CLI"
--- a/geoparquet_io/core/check_fixes.py
+++ b/geoparquet_io/core/check_fixes.py
@@ -22,6 +22,7 @@ from geoparquet_io.core.exceptions import GeoParquetError, RemoteAccessError
 from geoparquet_io.core.file_utils import is_same_file_path, resolve_file_url
 from geoparquet_io.core.hilbert_order import hilbert_order
 from geoparquet_io.core.logging_config import debug, info, progress
+from geoparquet_io.core.parquet_writer import DEFAULT_SORT_ROW_GROUP_ROWS
 from geoparquet_io.core.remote import (
     get_remote_error_hint,
     is_remote_url,
@@ -29,6 +30,30 @@ from geoparquet_io.core.remote import (
     remote_write_context,
     setup_aws_profile_if_needed,
 )
+
+# Every fix below writes DEFAULT_SORT_ROW_GROUP_ROWS rows per row group.
+#
+# `check --fix` is remedial, so the file it leaves behind has to pass the checks
+# gpio runs next -- including the stricter of the two row-count bands, the
+# 10,000-50,000 rows per group of SPATIAL_ROW_COUNT_RANGE that
+# `check optimization` scores as one of its five factors. Until #972 all five
+# write sites asked for a hardcoded 100,000, which the writer rounds up to
+# 100,352: inside GENERAL_ROW_COUNT_RANGE, outside the spatial one. So
+# `check all --fix` produced a file `check optimization` scored [fail] on its
+# row-group factor and advised re-partitioning -- a repair that fails the check
+# it exists to satisfy.
+#
+# Targeting the spatial band is the choice #972 asked for, and `--fix` is
+# already committed to it: it Hilbert-sorts the file (fix_spatial_ordering), and
+# sorting exists to make spatial predicates prune row groups. Groups too large
+# to prune well would undo what the sort just bought. Of the two bands it is
+# also the one with a measurement behind it (#775).
+#
+# The number is not typed here. DEFAULT_SORT_ROW_GROUP_ROWS is
+# align_to_writer_vector(SPATIAL_BAND_TOP_ROWS) -- derived from the band, and
+# already snapped to a whole 2,048-row writer vector, so what gpio asks for is
+# what lands. It is also exactly what `gpio sort` writes since #967, which is
+# what #795 meant by fix_spatial_order inheriting the sort default.
 
 
 def fix_compression(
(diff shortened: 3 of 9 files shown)


cholmes:
## Multi-agent adversarial review — the #970 half is exemplary; the #972 half was half-closed, fixed in 0d72613

### The #970 half: merge as-is, and it sets the bar

Ten sabotage spellings run against `.pre-commit-config.yaml`, restoring between each — **all ten caught**, including the four that defeated #964's first attempt:

| spelling | result |
|---|---|
| bare `grep -v "logging_config.py"` / `"common.py"` | caught |
| no caret, colon kept | caught |
| double-quoted but anchored | caught |
| `awk '!/common\.py/'` instead of `grep -v` | caught |
| `--exclude="common.py"` on the search | caught |
| exempt path in a shell variable | caught |
| extra `grep -v 'partition/'` stage | caught |
| `files:` / `exclude:` YAML keys | caught |

The anchoring provably flags nothing new — both guards return zero matches tree-wide, so the three modules that were silently exempt (`partition/common.py`, `process/aggregate/common.py`, `grid_common.py`) contain nothing to report. The shared-helper extraction is behaviour-preserving down to the assert text: identical collected test IDs in both refactored modules, every assert byte-identical after normalising indentation, and #964's arm-scoping discipline preserved.

### S2 — #972 was only half closed; its own transcript still reproduced

The PR changed what `--fix` **writes**. It did not change what **triggers** it. `fix_available` derived from `assess_row_count`, whose verdict is deliberately the *general* band — so a 100,352-row-group file was "optimal" and no fix ran. Reproduced on a 400k-row / 4-group file:

```
$ gpio check row-group f.parquet --fix
✓ No fix needed - row groups are optimal!
$ gpio check optimization f.parquet
[fail] Row Group Size: Average 100,000 rows per group (optimal 10k-50k for spatial queries)
```

Verbatim #972's transcript, **after** the fix. Worse, behaviour had become less predictable than before — whether `check all --fix` left a clean file depended on whether some *unrelated* repair happened to force a rewrite:

| `--fix` subcommand, same source | groups |
|---|---|
| `check spatial` / `compression` / `bbox` | 9 × 49,152 ✅ |
| `check row-group` | **4 × 100,352, "no fix needed"** ❌ |

**Fixed by moving the trigger, not the verdict.** A new `row_count_fix_available()` derives from `SPATIAL_ROW_COUNT_RANGE` while `assess_row_count`'s verdict, message and `passed` stay exactly as #795 and #958 settled them. The sub-64 MB single-group exemption moved into one shared `_is_small_single_group()` so verdict and trigger cannot drift apart. Verified:

```
$ gpio check row-group f.parquet --fix
Row count per group is optimal for general use (10,000-200,000); spatial queries prune more with 10,000-50,000
Row groups are larger than the spatial band (10,000-50,000); --fix rewrites at 49,152 rows per group
✓ Row groups optimized successfully!
$ gpio check optimization f.parquet     ->  [pass] Average 44,444 rows per group   (9 groups)
```

The issue's literal `check all --fix` transcript now passes, and it is idempotent — re-running says `✓ No fix needed`. The wording is coherent: nothing says "optimal" and then rewrites in silence.

I suspected 49,152 was a spatial number misapplied to the compression and bbox sites. **Measurement says otherwise** — 400k rows × 60 float columns, ZSTD-15:

| request | groups | file size |
|---|---|---|
| 100,000 | 4 | 226,907,468 B |
| 49,152 | 9 | **224,854,588 B** |

The smaller groups produce a file **0.9% smaller**; the doubled footer costs 34 KB, 0.015%. No measurable cost. I also looked for an edited pre-existing assertion of `100000` — the shape a regression usually hides in — and there is none: `test_check_fixes_core.py:636` still writes it as a *fixture* and still passes.

### Also fixed

`_pipeline` identified the arm with `script.count("if ") == 1`, which also matches `elif ` and any future comment containing "if ". It now matches `^[ \t]*if[ \t]` at a statement position — a spelling-sensitivity worth removing from a module whose whole thesis is not to depend on spellings.

New CLI tests close the gap the reviewer found: the old tests called `check_fixes.fix_*()` directly, bypassing the `fix_available` gate and the CLI — which is exactly where the residual defect lived. `TestTheIssuesOwnReproductionThroughTheCli` now runs the issue's commands through `CliRunner`, and `TestTheTriggerFollowsTheSpatialBand` pins that the verdict stays `optimal`/`passed` while `fix_available` is True. File: 12 → 18 tests.

### Filed separately — #981

`core/convert.py:1968` advertises `gpio convert geoparquet` as applying "100k row groups", but passes `row_group_rows=None`, so DuckDB picks 122,880 — and even 100k would land at 100,352 and fail `check optimization`. The same self-contradiction one command over, in a command whose docstring claims to apply best practices including Hilbert ordering.

### Gates

Fast suite **7059 passed, 76 skipped, 9 xfailed** — one `test_pipe_integration` failure on the known cold-run flake list, **26 passed serially** · `diff-cover --fail-under=90` → **100% on 21 lines** · `pytest -m meta` 205 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push clean.

Issue #761, opened 2026-08-29
Table.write() has no memory_limit: needs a write-settings object, not an 11th parameter


cholmes:
Split out of #673 (item 5).

`geoparquet_io/api/` has no `memory_limit` anywhere. `api/ops.py` is
`pa.Table → pa.Table`, so it genuinely does not apply there — but `Table.write()`
does write files, and already takes `compression`, `compression_level`,
`row_group_size_mb`, `row_group_rows`, `geoparquet_version`, `write_strategy`,
`profile` and more.

`docs/guide/write-strategies.md` currently documents the memory override as
CLI-only, which is accurate and also an admission that the Python API cannot do
something the CLI can.

## The decision

Adding `memory_limit=` as one more keyword is the obvious move and the wrong one —
`Table.write()`'s signature is already at the point where callers pass things
positionally by accident. The shape worth deciding on is a **write-settings object**:

```python
tbl.write("out.parquet", settings=WriteSettings(memory_limit="2GB", compression="ZSTD"))
```

which would also give `row_group_*`, `compression*` and `write_strategy` somewhere to
live, and gives the CLI a single thing to build. That is a public-API design change,
so it needs deciding rather than defaulting.

Also worth settling in the same pass: `--write-memory` only applies to the
`duckdb-kv` strategy (see the warning in `docs/guide/write-strategies.md`), so the
API surface has to express that constraint too.

Related: #663, #673.

cholmes:
## Ruling (maintainer-approved, 2026-09-05)

**Yes to a write-settings object; it is built by the first write-facade PR (Deep Review 3.3), not before.**

Shape:

```python
@dataclass(frozen=True)
class WriteOptions:
    geoparquet_version: str | None = None   # None = preserve / auto
    write_strategy: str = "auto"
    compression: str = "ZSTD"
    compression_level: int | None = None
    row_group_rows: int | None = None
    row_group_size_mb: int | None = None
    memory_limit: str | None = None         # DuckDB memory cap for the operation
    temp_directory: Path | None = None      # spill location (Deep Review 2.1)
    threads: int | None = None              # Deep Review 2.6
```

Rules:

- `Table.write(path, options: WriteOptions | None = None, **kwargs)`. Existing keywords keep working and merge into the object; naming the same field both in `options` and as a keyword is an error, not a silent override. No deprecation of the keywords — they stay as the one-off convenience.
[...]

PR #977, opened 2026-09-10
fix(duckdb): spill to a private temp directory instead of ".tmp" under the working directory

cholmes:
This is **Deep Review item 2.1 (central spill strategy)**. There is no tracking
issue for it — the item lives only in the Deep Review — so there is nothing to
reference here beyond #761, which this unblocks.

## What changed

`get_duckdb_connection` now gives every connection its own spill directory
under the OS temp directory when the caller does not name one. That is the
whole mechanism; the rest is fallout.

- `spill_directory(base_dir=None)` in `core/duckdb_utils.py` returns a fresh,
  **not-yet-created** path under `base_dir` (default: the OS temp directory,
  i.e. `TMPDIR`). DuckDB creates the directory the first time a query spills
  and removes it when the connection closes, so a run that never spills writes
  nothing and leaves nothing behind.
- The four call sites that already passed a `temp_directory` all passed the
  *same* well-known path — the admin cache dir — so they now take a private
  leaf under that same volume (`spill_directory(cache_dir)`). See the hazard
  under "Why" below.
- `temp_directory=` keeps its exact meaning: the path is used verbatim.
- `docs/guide/write-strategies.md` gains a "Where Spilled Data Goes" section,
  and its `extract bigquery` note is corrected — that connection is built
  through `get_duckdb_connection`, so it now has a spill path.

Four source files, 12 changed lines. No call site had to name the setting.

## Why

DuckDB's default `temp_directory` is the **relative** path `.tmp`, and 98 of
gpio's 102 `get_duckdb_connection` call sites left it there, so every spill
landed under whatever directory the user happened to run gpio from. Reproduced
below: a read-only working directory turns a sort into a hard failure even
though the input and output volumes are both writable.

### Why the default lives in the connection factory, not the write path

"Beside the output" was the review's stated intent, and it is not reachable.
DuckDB refuses to relocate a temp directory once it has been used:

```
NotImplementedException: Cannot switch temporary directory after the current one has been used
```

So a per-write override cannot be applied to a connection gpio shares across
writes — a partition loop finalizes N files on one connection, and file 2 would
throw. The location has to be fixed when the connection is opened, which is
exactly one place. `TMPDIR` is then the user's knob for a machine whose root
volume is too small to hold a large sort's spill, and it is also the answer for
an output path on a network or object-store mount, where spilling beside the
output would be pathological.

### Why the directory must be private per connection

This is correctness, not tidiness. DuckDB names its spill files after the block
size alone — `duckdb_temp_storage_S192K-0.tmp` — with nothing in the name
identifying the connection, the database or the process. Three connections
pointed at one directory, each sorting:

```
IOException: Could not read enough bytes from file ".../duckdb_temp_storage_S192K-0.tmp"
InvalidInputException: Invalid unicode (byte sequence mismatch) detected in segment statistics update
```

They overwrite each other's blocks. That is why the admin callers, which all
shared the cache directory, changed: two concurrent `gpio add admin-divisions`
runs (or one plus a `gpio partition admin`) could corrupt each other's spill.

### What did *not* get a default

- **`memory_limit` stays opt-in.** A limit only helps once spilling works, and
  a default below DuckDB's own (~80% of RAM) would push queries to disk that
  fit in RAM today.
- **`threads=1` and `preserve_insertion_order=false` stay where they are.**
  The repo's operational experience is that reliable spilling on large joins
  needed all three together, and `duckdb-kv` already clamps these two for the
  duration of a write (`_configure_duckdb_memory`) — it was only missing the
  temp directory, which it now inherits. As *global* defaults they would
  serialise every read gpio does and change result ordering for callers that
  never write a file.

## Verification

**The reproduction.** From a read-only CWD, with input and output on writable
volumes, a 600MB sort through `duckdb-kv`:

```
cwd writable: False
RESULT: FAILED -> IOException IO Error: Failed to create directory ".tmp": Permission denied
```

After the change the same write succeeds. It is pinned as
`test_spilling_write_survives_a_read_only_working_directory`, which runs in a
subprocess (the `no-cwd-change-in-tests` hook forbids moving this suite's own
CWD) and asserts all three steps in order: stock DuckDB fails on `.tmp`; a gpio
connection spills anyway; and a `write_parquet_with_metadata` sort larger than
its memory limit lands its spill in gpio's private directory and writes every
row, leaving the working directory empty.

`tests/test_spill_strategy.py` — 12 tests, 3.7s — also covers: the helper is
unique per call and creates nothing; the default is absolute and under
`gettempdir()`; two connections never share; an explicit `temp_directory` is
used verbatim; spill files really land there and are cleaned up on close; three
concurrent spilling connections do not corrupt each other (the regression guard
for the hazard above); and `memory_limit`/`threads`/`preserve_insertion_order`
are left at DuckDB's defaults.

| Check | Result |
|---|---|
| `pytest -n auto -m "not slow and not network and not meta"` | 7034 passed, 76 skipped, 9 xfailed (179s) |
| `pytest -m meta` | 204 passed (one `cz check` subprocess timeout under `-n auto`; passes serially, 29/29) |
| `pre-commit run --all-files` | clean |
| `pre-commit run --all-files --hook-stage pre-push` | clean (incl. menard-check, xenon, import-linter, vulture, mypy) |
| `diff-cover --compare-branch=origin/main` | **100%** — 12 changed lines, 0 missing |

## Summary by CodeRabbit

* **New Features**
  * DuckDB temporary storage now uses private, per-run directories to improve concurrent reliability.
  * CLI errors caused by insufficient temporary disk space now include actionable guidance.
  * Cache cleanup now identifies and removes leftover temporary spill data, with deletion and storage totals reported.

* **Documentation**
  * Added guidance on temporary storage, cleanup, read-only workspaces, limited volumes, concurrent runs, and interrupted processes.

### DIFF
--- a/docs/guide/write-strategies.md
+++ b/docs/guide/write-strategies.md
@@ -117,10 +117,10 @@ Override the default strategy when needed:
 A memory limit only helps if DuckDB has somewhere to put the data that does not
 fit. gpio gives every DuckDB connection its own scratch directory under the
 system temp directory — `$TMPDIR` on macOS and Linux, `%TEMP%` on Windows — and
-DuckDB removes it again when the connection closes. A run that never exceeds its
+DuckDB removes it again when the query finishes. A run that never exceeds its
 memory limit writes nothing there.
 
-This matters in two situations:
+This matters in three situations:
 
 - **A read-only working directory.** DuckDB's own default spill location is the
   relative path `.tmp`, so out of the box a large sort fails with
@@ -135,6 +135,22 @@ This matters in two situations:
     TMPDIR=/mnt/scratch gpio sort hilbert huge.parquet sorted.parquet
     ```
 
+- **A RAM-backed `/tmp`.** systemd mounts `/tmp` as a tmpfs sized at half of RAM
+  by default on Fedora, Arch and openSUSE, and so do Kubernetes'
+  `emptyDir: {medium: Memory}` and `docker run --tmpfs /tmp`. Spilling onto a
+  tmpfs spends the very memory the spill was meant to save, so on those systems
+  `TMPDIR` should name real storage — `df -h /tmp` and `findmnt /tmp` say which
+  you have.
+
+!!! warning "\"Out of Memory Error\" can mean out of *disk*"
+    DuckDB caps its own spill at `max_temp_directory_size` (90% of the spill
+    volume by default), so a small or RAM-backed temp volume fails cleanly
+    rather than filling the disk. The message it raises, though, says
+    `Out of Memory Error: failed to offload data block of size 256.0 KiB` and
+    suggests only memory-side remedies. gpio appends a line naming `TMPDIR`
+    whenever DuckDB's error mentions `max_temp_directory_size`, because that is
+    the knob that actually fixes it.
+
 !!! warning "Do not point two runs at one spill directory"
     DuckDB names its spill files after the block size alone
     (`duckdb_temp_storage_S192K-0.tmp`), with nothing in the name identifying
@@ -143,6 +159,37 @@ This matters in two situations:
     `IO Error: Could not read enough bytes from file`. `TMPDIR` is safe because
     gpio still gives each connection a private subdirectory beneath it; a fixed
     `temp_directory` passed straight to DuckDB is not.
+
+### Cleaning Up After an Interrupted Run
+
+DuckDB removes its scratch directory when the query finishes — including when it
+fails. A run **killed** mid-spill (Ctrl-C, `SIGTERM`, the OOM killer) is the
+exception: the process is gone before that cleanup can run, and a directory
+holding gigabytes stays on disk.
+
+gpio names those directories `gpio-spill-<pid>-<random>` and sweeps them at the
+start of every run: any leftover whose owning process has exited is removed
+before a new one is created, so interrupted runs do not accumulate. A directory
+belonging to a process that is still running is never touched. (Pids are the
+test, so give each container its own `TMPDIR` rather than bind-mounting one
+spill volume into several — across pid namespaces the test cannot tell a live
+run from a dead one.)
+
+The one place an OS temp reaper would never help is the admin cache, because the
+admin-boundary commands spill onto the cache volume rather than into `/tmp`.
+`--clear-cache` prices and deletes those leftovers too:
+
+<!-- doctest: skip="deletes the user's real admin cache" -->
+```bash
+gpio add admin-divisions in.parquet out.parquet --clear-cache
+```
+
+```
+Cache directory: /home/you/.geoparquet-io/cache/admin
+Files to delete: 1
+Total size: 3.00 MB
+Leftover spill directories: 1 (250.00 MB)
+```
 
 ### Automatic Detection
 
--- a/docs/guide/write-strategies.md
+++ b/docs/guide/write-strategies.md
@@ -112,6 +112,38 @@ Override the default strategy when needed:
 
 ## Memory Configuration
 
+### Where Spilled Data Goes
+
+A memory limit only helps if DuckDB has somewhere to put the data that does not
+fit. gpio gives every DuckDB connection its own scratch directory under the
+system temp directory — `$TMPDIR` on macOS and Linux, `%TEMP%` on Windows — and
+DuckDB removes it again when the connection closes. A run that never exceeds its
+memory limit writes nothing there.
+
+This matters in two situations:
+
+- **A read-only working directory.** DuckDB's own default spill location is the
+  relative path `.tmp`, so out of the box a large sort fails with
+  `IO Error: Failed to create directory ".tmp"` even when the input and output
+  volumes are perfectly writable. gpio never spills into the working directory.
+- **A small root volume.** A sort of a very large file can spill more bytes than
+  the output itself. If `/tmp` is too small, point `TMPDIR` at a volume that is
+  not:
+
+    <!-- doctest: skip="names /mnt/scratch and huge.parquet, neither of which the harness seeds" -->
+    ```bash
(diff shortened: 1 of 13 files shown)


cholmes:
## Multi-agent adversarial review — two S2s found and fixed in ad46498

The core change is right, minimal and correctly escaped. My attacks on SQL escaping, uniqueness/races, Windows path length, the admin-cache lifecycle and test honesty all came back clean, with evidence:

- **Escaping is exactly once.** Round-tripped through `SET temp_directory` for `C:\Users\...\Temp\...`, `C:\Temp\o'brien\x` and an embedded newline — all OK; plus an end-to-end 400k-row spill into a directory named `o'brien dir` (apostrophe *and* space). No #718 shape.
- **Collisions are impossible** — pid plus 48 bits of randomness, with three concurrently spilling connections tested.
- **The large-join combination is intact** at all four migrated sites (`threads=1`, `preserve_insertion_order=false` untouched), and nothing re-downloads: the leaf is a *sibling* of the cached parquet and every cache lookup is exact-name or a `*.parquet` glob.
- **The tests force real spills** (`memory_limit='128MB'` + a 400k-row `ORDER BY hash(i)`) and assert actual `duckdb_temp_storage*` files, not a settings string.

### S2 — unique names turned a bounded leak into an unbounded one

DuckDB removes the spill directory only when the query **completes**. Normal error paths are fine, but an interrupted long-running spill is not — and the old fixed `.tmp` name was self-limiting because DuckDB reuses it verbatim. Unique names removed that self-healing, and nothing reaped them: `_SPILL_DIR_PREFIX` existed only to make orphans *recognisable*, which is the naming half of a mitigation whose other half was never written.

Three interrupted runs, measured:

```
sigint   old (fixed ".tmp"):  2428 -> 2566 -> 2914 MB    1 leaf
         new (no reaper):     2678 -> 4351 -> 5979 MB    3 leaves
         swept (this fix):    2529 -> 2377 -> 2426 MB    1 leaf
```

SIGKILL behaves the same. The fix plateaus at exactly one leaf — the last run's, which the *next* run reaps.

`spill_directory()` now sweeps its base before minting. Ownership is decided by **pid liveness alone, never by age**, so a six-hour run cannot have its scratch swept out from under it, and `_process_is_alive` answers True whenever it cannot tell. I verified the scoping independently:

```
orphans identified:  ['gpio-spill-999999-bbbbbbbbbbbb']
surviving:           live-pid leaf, a non-gpio dir, an unparseable pid, a FILE named like a leaf
```

Only the dead-pid directory is removed. Every failure path is swallowed — tidying up must not fail the run that was only trying to start.

### S2 — the worst instance was in `$HOME`, invisible to `--clear-cache`

The four migrated sites put the leaf *inside* `~/.geoparquet-io/cache/admin/`, and both `clear_cache` (`admin_datasets.py:280`) and the size report (`add.py:201`) glob **only** `*.parquet`. So Ctrl-C'ing `gpio add admin-divisions` on Overture left multi-GB in the user's home that `--clear-cache` reported as 0 files and no OS tmp-reaper would ever touch. Now:

```
before:  Cleared cache: 1 files, 3.00 MB freed        after: 250M gpio-spill-...  <- still there
after :  Leftover spill directories: 1 (250.00 MB)
         plus 1 leftover spill directory, 250.00 MB freed   after: 0B
```

### S3 — the TMPDIR hazard, which my brief got half right

`max_temp_directory_size` defaults to 90% of the temp volume, so DuckDB self-limits and the failure is **clean, not corrupting** — my "it will fill the disk" framing was wrong. But on a 64 MB volume the message says *memory* for a *disk* shortage and never mentions `TMPDIR`, the one knob that fixes it. `spill_space_hint()` now recognises the real error (provoked with a 16 MB cap, not a mocked string) and adds gpio's own text; the guide names the tmpfs case by name — systemd's RAM-backed `/tmp` on Fedora/Arch/openSUSE, `emptyDir: {medium: Memory}`, `--tmpfs /tmp` — which matters because gpio's own write default sets `memory_limit` to 50% of RAM, so spilling there converts memory pressure into more memory pressure.

Both false cleanup claims (`write-strategies.md:120`, `duckdb_utils.py:562`) are corrected, and the Windows-sensitive assertion on DuckDB's own directory removal no longer runs there.

### A hole the fixer named rather than papered over

**Pid liveness cannot decide ownership across pid namespaces.** Bind-mount one spill volume into two containers and each sees the other's pids as absent, so a sweep could delete a live foreign run's directory. An mtime grace period was considered and rejected with a reason I agree with: a grace long enough to protect a stalled foreign query is long enough to break the accumulation bound. A cross-namespace-safe answer needs an `flock`ed marker inside the leaf, which means creating it eagerly and losing the "never created unless DuckDB spills" property this branch's tests pin — a redesign, not a fix. Documented in the docstring and the guide (give each container its own `TMPDIR`). Same-machine siblings are fully protected.

Also worth noting the fixer **caught its own bad measurement**: its first accumulation run gated the interrupt on a byte threshold that counted earlier leftovers, so runs 2 and 3 were killed before spilling and every mode looked flat. The numbers above use a fixed wall-clock interrupt.

### Surface

`cli_surface.json` moves by exactly one line — `"class": "Group"` → `"SpillAwareGroup"` on the root group. No option, command or help text changed. `SpillAwareGroup` is deliberately the root group rather than `handle_geoparquet_errors`, which covers only four command modules and would miss `gpio sort hilbert`, the guide's own worked example.

### Gates

Fast suite **7059 passed, 76 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **98%** (2 lines, both `except OSError: continue`) · `pytest -m meta` 205 passed · docs lane 236 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push clean · 25 new tests, fast lane, written before the implementation.

PR #978, opened 2026-09-10
fix(geo-metadata): keep a projected CRS through extract and sort, and stop 'check' crashing on the file it diagnoses

### LINKED ISSUE #968: fix(geo-metadata): four live malformed-'geo' readers remain, one silently loses a projected CRS
## Summary

#883 → #887 (#945) → #947 (#960) hardened three batches of readers that index a carried `geo` block without a shape check. #960's body calls itself "the third and last instalment" — an adversarial review found **four more live members of the same class**, and the lead one is higher severity than anything the three PRs fixed.

All four reproduce on `main` (`f28fe48`) and are untouched by #960, so none of them blocked that merge.

---

## 1. LEAD — `extract` and `sort hilbert` silently drop a projected CRS

`geoparquet_io/core/geo_metadata.py:826` (`_prune_geo_dict_to_columns`) reads `primary_column` raw:

```python
primary = geo_dict.get("primary_column")
if primary is not None and primary not in col_entries:
    if not (repoint_primary and col_entries):
        return _DROP_GEO
```

`123` is not `None` and is not a key, so **the entire `geo` block is dropped — CRS included**. An absent `crs` is spec-defined as OGC:CRS84, so a projected file is silently relabelled lon/lat.

Reproduced on a file whose `geom` column carries EPSG:5070 PROJJSON with `primary_column: 123`:

```
control (primary_column: "geom")   extract geoparquet -> crs: 5070
malformed (primary_column: 123)    extract geoparquet -> crs: ABSENT  (=> OGC:CRS84)
```

Same for `gpio sort hilbert`. This is exactly the corruption shape #945's review caught (a projected file treated as lon/lat), still live — and the fix is to route that read through the `sanitize_geo_metadata` / `carried_column_name` helpers #945 and #960 already shared out. Note `_repair_primary_column` from #945 would resolve the single-column case correctly if it ran here.

**This is the one to fix first.** It is silent data corruption, not a crash.

## 2. `gpio check spec` crashes on the malformed file it exists to diagnose

`geoparquet_io/core/validate.py:730` (`_columns_declaring_covering`):

```
for name, col in (geo_meta.get("columns") or {}).items()
AttributeError: 'list' object has no attribute 'items'
```

Reproduced with `"columns": ["geom"]` — exit 1, raw traceback. Also `'NoneType'` for `columns: null` and `'str'` for `columns: "geom"`.

`check spec` is the command a user runs to be *told* their file is malformed. It dies instead. This is the identical code shape #960 fixed at `common.py:3460`, one module over.

## 3. `gpio check bbox` crashes on a non-object block

`geoparquet_io/core/check_parquet_structure.py:463` (`_check_geoparquet_v1`):

```
version = geo_meta.get("version", "0.0.0") if geo_meta else "0.0.0"
AttributeError: 'list' object has no attribute 'get'
```

Same command #960 fixed one layer down (`_check_bbox_metadata_covering`), so the file now gets past the inner guard and dies at the outer one.

## 4. Non-string `primary_column` still reaches `quote_identifier`

`geoparquet_io/core/streaming.py:206-224` (`find_geometry_column_from_metadata`) returns `geo_meta.get("primary_column", "geometry")` with no type check. Via `core/extract.py:964`:

```
$ gpio extract geoparquet <file> -
  ... geometry_repair.py:105 -> duckdb_utils.py:239
  TypeError: argument of type 'int' is not iterable
```

This is #887 item 3 verbatim, which #945's `carried_column_name` was meant to close. The same line also carries the literal `"geometry"` fallback that #945's review identified as the cause of its silently-wrong-coordinates bug — for a file whose column is `geom`, that default names a column that does not exist.

---

## Also worth folding in

- **A second site with the same `"geometry"` default.** `streaming.py:584` — `geom_col_name = geometry_column or geo_meta.get("primary_column", "geometry")` inside `extract_crs_from_table`. Dead today (both callers pass `geometry_column` explicitly) but it contradicts #960's "the one place still open to it".
- **`_geo_block_to_carry_on_fast_path` can emit a block with no `primary_column`** (REQUIRED by the spec) when sanitize drops a malformed one and `columns` has 2+ entries so the repair cannot fire. Unit-provable; not reachable through any command today because finding 1 drops the whole block first. Defence-in-depth.
- **`carried_geometry_column` misses non-standard column names.** A table whose only geometry column is `footprint` gets a block with no `primary_column`, because `detect_geometry_column_from_names` knows only the five standard names. Falling back to the single `columns` key when it appears in the table's columns would fix it. Identical on main, not a regression.
- **`add_bbox_metadata_table` can return a block missing REQUIRED keys.** `columns: {"geom": "WKB"}` → sanitize drops the entry → it is recreated as `{}` → the written file fails five of its own spec checks (`encoding_valid`, `geometry_types_list`, `encoding_matches_data`, `geometry_types_match_data`, `coordinates_valid_for_crs`). `require_geo_metadata_for_covering` guards a wholly-absent block but not a partly-unusable one. Byte-identical on main.
- **Raw DuckDB tracebacks at the command boundary.** `gpio add bbox-metadata` on shapes DuckDB itself refuses exits 1 with an unhandled `_duckdb.InvalidInputException`. Catching `duckdb.Error` and re-raising `GeoParquetError` would match how the block-level shapes already report.

## Why this keeps recurring

Four PRs have now fixed this class per-instance (#883, #887/#945, #947/#960), and each declared its batch complete. The reliable fix is mechanical: every read of a carried `geo` block should go through `sanitize_geo_metadata` (write paths), `carried_column_name` (identifier reads), or an explicit guard (validation paths). A grep-based `meta`-lane test that fails on any *new* raw `geo_meta["columns"]` / `.get("primary_column")` outside those helpers would stop the fifth instalment — the same argument as `tests/test_patch_target_shadowing.py` added in #948, and the same lesson as the `duckdb-antipatterns` guard blind spot: when a defect class repeats, build the guard rather than fixing the instance.

Found by an adversarial review of #960.

cholmes:
## What changed

The fourth instalment of #883's class, and the first one whose lead item is
**silent data corruption** rather than a crash. Each site is classified the way
#883, #945 and #960 classified theirs, and the classification decides the fix.

| # | Site | Class | Symptom on `origin/main` |
|---|---|---|---|
| 1 | `geo_metadata._prune_geo_dict_to_columns` | write path → **sanitize** | a projected file silently relabelled OGC:CRS84 |
| 2 | `validate._columns_declaring_covering` + both per-column loops in `_run_geoparquet_checks` | validation → **guard** | `AttributeError: 'list' object has no attribute 'items'` |
| 3 | `check_parquet_structure._check_geoparquet_v1` | validation → **guard** | `AttributeError: 'list' object has no attribute 'get'` |
| 4 | `streaming.find_geometry_column_from_metadata` | column name → **guard** | `TypeError: argument of type 'int' is not iterable` |

**1 — the lead.** `_prune_geo_dict_to_columns` compared the *raw* carried
`primary_column` against the surviving `columns` keys. `123` is neither `None`
nor a key, so the whole `geo` block was dropped — `crs` included. An absent
`crs` is spec-defined as OGC:CRS84, so a projected EPSG:5070 file came out of
`gpio extract geoparquet` and `gpio sort hilbert` relabelled lon/lat, silently.

It is a write path — whatever survives is re-encoded into the output's `geo` key
— so it now goes through the shared `sanitize_geo_metadata`. `_repair_primary_column`
**was** reusable and does the whole job for the single-column case; the one thing
it could not do is the case where `columns` still held two entries at sanitize
time and pruning then left one, so the repair is retried after pruning.

**2 and 3 are validation readers and do not sanitize** — `gpio check` has to see
the file as it really is. `columns_present` now FAILs a non-object *entry*,
naming it, and the per-column loops run over the entries that can be read;
`_check_geoparquet_v1` reads a version only from a block that is an object.
Nothing usable is hidden: a 1.0 file that really does declare a 1.1-only
`covering` still fails `version_features_match`, and a well-formed 1.0 file still
gets its outdated-version report.

**4** is a column-name reader shared with read-only callers
(`Table.geometry_column`), so it takes #945's line: guard, hand back nothing
rather than a name that cannot be one, and let the caller ask the file's schema —
which every caller already does. That also retires the literal `"geometry"`
fallback this line carried (on a `geom` file it names a column that does not
exist — #945's review's silently-wrong-coordinates bug), here and at the second
site `streaming.extract_crs_from_table` still had, which the issue flagged as
contradicting #960's "the one place still open to it".

### The lead, reproduced

One row, `crs` = EPSG:5070 PROJJSON, the only difference between the two inputs
being `primary_column`. The control run pins what the command does with a
well-formed block; the malformed run has to match it.

**Before (`origin/main`):**

```
input  control  (primary_column: "geometry") : crs EPSG:5070 (NAD83 / Conus Albers)
input  malformed (primary_column: 123)       : crs ABSENT on primary 123

gpio extract geoparquet   control   -> crs EPSG:5070 (NAD83 / Conus Albers)
gpio extract geoparquet   malformed -> crs ABSENT on primary 'geometry'   <-- OGC:CRS84
gpio sort hilbert         control   -> crs EPSG:5070 (NAD83 / Conus Albers)
gpio sort hilbert         malformed -> crs ABSENT on primary 'geometry'   <-- OGC:CRS84
```

**After:**

```
gpio extract geoparquet   control   -> crs EPSG:5070 (NAD83 / Conus Albers)
gpio extract geoparquet   malformed -> crs EPSG:5070 (NAD83 / Conus Albers)
gpio sort hilbert         control   -> crs EPSG:5070 (NAD83 / Conus Albers)
gpio sort hilbert         malformed -> crs EPSG:5070 (NAD83 / Conus Albers)
```

### The three tracebacks

```
# 2. gpio check spec, columns: ["geom"] — the command run to be *told* the file is malformed
  File "geoparquet_io/core/validate.py", line 754, in _check_version_features
    covering_cols = _columns_declaring_covering(geo_meta)
  File "geoparquet_io/core/validate.py", line 730, in _columns_declaring_covering
    for name, col in (geo_meta.get("columns") or {}).items()
AttributeError: 'list' object has no attribute 'items'

# and one layer on, for columns: null / a string / an entry that is not an object
  File "geoparquet_io/core/validate.py", line 3927, in _run_geoparquet_checks
    for col_name, col_meta in columns.items():
AttributeError: 'NoneType' object has no attribute 'items'
```

```
# 3. gpio check bbox, geo block = ["geometry"]
  File "geoparquet_io/core/check_parquet_structure.py", line 476, in _check_geoparquet_v1
    version = geo_meta.get("version", "0.0.0") if geo_meta else "0.0.0"
AttributeError: 'list' object has no attribute 'get'
```

```
# 4. gpio extract geoparquet <file> -, primary_column: 123
  File "geoparquet_io/core/duckdb_utils.py", line 239, in quote_identifier
    if "\x00" in name:
TypeError: argument of type 'int' is not iterable
```

All four are pre-existing on `origin/main` and untouched by #960, so none of them
blocked that merge.

## Verification

Every site was reproduced on `origin/main` first and re-run after the fix.

**A sweep, not a claim.** 8 malformed shapes (`columns` null/array/string, an
entry that is not an object, a non-string `primary_column`, a non-string
`version`, a block that is an array, a block that is a string) x 26 CLI commands,
run against a pristine `git archive origin/main` tree and against this branch
with the same script:

| | raw `TypeError`/`AttributeError` |
|---|---|
| `origin/main` | **12** |
| this branch | **6** |

The 6 that remain are all one shape — see below. They are pre-existing and not
members of the batch this PR fixes.

- `uv run pytest -n auto -m "not slow and not network and not meta" -q` — **7080 passed**, 76 skipped, 9 xfailed
- `uv run pytest -m meta -q` — **205 passed**
- `uv run pre-commit run --all-files` clean; `--hook-stage pre-push` clean, `menard-check` and `xenon` included
- `diff-cover coverage.xml --compare-branch=origin/main --fail-under=90` — **28 changed lines, 0 missing, 100%**
- 41 new tests in `tests/test_malformed_geo_metadata.py`, parametrized over both
  `geometry` and `geom` as #945 established, with a **control run** beside each
  end-to-end assertion so a fix that merely stopped writing a CRS could not pass,
  and a positive test per site so no guard costs a well-formed file its report.

One flake, named rather than hidden: `test_pipe_integration.py::test_add_bbox_to_sort_hilbert`
fails only in the full ~7,100-test `-n auto --cov` run on this shared machine. It
passes serially, in isolation under `-n auto --cov`, and in the uninstrumented
full suite — the saturation noise CLAUDE.md documents, not this change.

## There is a fifth batch — 2 sites, 6 live command paths

Three PRs have each called themselves the last one, so this one will not. The
sweep found a batch that all four PRs missed because they all read the same two
keys. It is a **third key**: a non-string `version`.

| Site | Reached by |
|---|---|
| `core/common.py:197` `detect_geoparquet_file_type` — `version.startswith("2.")` | `check spec`, `check bbox`, `check optimization`, `add bbox` |
| `core/streaming.py:645` `extract_version_from_metadata` — `version.split(".")` | `sort hilbert` (via `hilbert_order._resolve_output_version`), `add quadkey` |

```
$ gpio check spec <file with "version": 2>
  File "geoparquet_io/core/common.py", line 197, in detect_geoparquet_file_type
    if version and version.startswith("2."):
AttributeError: 'int' object has no attribute 'startswith'
```

Both are byte-identical on `origin/main`. `common.py` is outside this PR's scope
and accounts for 4 of the 6 paths, so fixing only the `streaming.py` half would
leave `check spec` still crashing on the same file — a worse state than a clean
report. Per the repo's separate-PRs-for-cross-cutting-fixes convention, this
belongs in its own change.

Two more findings from the sweep, both **pre-existing and verified identical on
`origin/main`** (checked against a `git archive` of it, not assumed):

- `sort hilbert` on a file with two *non-standard-named* geometry columns and no
  usable `primary_column` exits 1 with `BinderException: Referenced column
  "geometry" not found`. `find_primary_geometry_column` ends in
  `return detected if detected else "geometry"`, and `detect_geometry_column_from_names`
  knows only five standard names — the issue's `footprint` item. The
  single-column half of that item is already handled by `_repair_primary_column`;
  only the 2+-column case is live, where naming one would be a guess anyway.
- The remaining `_check_v2_*` raw `geo_meta.get("columns", {}).get(...)` reads in
  `validate.py` are **not** reachable: they run inside `for col_name in columns.keys()`,
  which this PR's normalization makes empty for every malformed `columns`.

The issue's own conclusion still stands and this PR does not deliver it: the
reliable fix is a `meta`-lane grep guard that fails on any *new* raw
`geo_meta["columns"]` / `.get("primary_column")` / `.get("version")` outside the
shared helpers. That would have caught this batch and the fifth one at once.

Closes #968

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed or incomplete GeoParquet metadata without unexpected errors.
  * Validation now reports invalid metadata structures clearly instead of failing during checks.
  * Preserved CRS information when repairing metadata during column extraction and sorting.
  * Geometry-column detection no longer assumes a default column when metadata is missing or unusable.
  * Empty or invalid primary-column values are safely corrected or ignored.

### DIFF
--- a/geoparquet_io/core/check_parquet_structure.py
+++ b/geoparquet_io/core/check_parquet_structure.py
@@ -473,7 +473,13 @@ def _check_geoparquet_v1(parquet_file, file_type_info, verbose, return_results, 
     from geoparquet_io.core.geo_metadata import covering_supported
 
     geo_meta = get_geo_metadata(parquet_file)
-    version = geo_meta.get("version", "0.0.0") if geo_meta else "0.0.0"
+    # `get_geo_metadata` is a read-only reader: it hands the block back exactly
+    # as the file holds it, so a block that is not a JSON object arrives here
+    # verbatim and used to crash with `'list' object has no attribute 'get'`
+    # (#968). A validation reader guards and reports the truth rather than
+    # sanitizing: a block that declares no version has none, which is what
+    # "0.0.0" says and what the outdated-version issue below reports.
+    version = geo_meta.get("version", "0.0.0") if isinstance(geo_meta, dict) else "0.0.0"
     bbox_info = check_bbox_structure(parquet_file, verbose)
 
     # Build results
--- a/geoparquet_io/core/geo_metadata.py
+++ b/geoparquet_io/core/geo_metadata.py
@@ -864,12 +864,11 @@ def _prune_geo_dict_to_columns(geo_dict: dict, columns: set[str], repoint_primar
     """
     # Sanitizing drops a non-string `primary_column` and repairs it from a lone
     # surviving column, which is the recovery this function needs and #883/#945
-    # already shared out. A block it cannot make usable is an absent one.
-    declared_primary = isinstance(geo_dict, dict) and "primary_column" in geo_dict
-    sanitized = sanitize_geo_metadata(geo_dict)
-    if not isinstance(sanitized, dict):
-        return _DROP_GEO
-    geo_dict = sanitized
+    # already shared out. `_rewrite_geo_metadata` only calls this with a decoded
+    # object -- a block that is not one never gets here -- so sanitizing a dict
+    # hands back a dict.
+    declared_primary = "primary_column" in geo_dict
+    geo_dict = sanitize_geo_metadata(geo_dict)
 
     col_entries = geo_dict.get("columns")
     if not isinstance(col_entries, dict):
--- a/geoparquet_io/core/geo_metadata.py
+++ b/geoparquet_io/core/geo_metadata.py
@@ -854,7 +854,23 @@ def _prune_geo_dict_to_columns(geo_dict: dict, columns: set[str], repoint_primar
     itself is gone but another declared geometry column remains,
     ``repoint_primary`` decides between naming that survivor as the new primary
     and dropping the whole block.
+
+    This is a write path -- whatever survives is re-encoded into the output's
+    ``geo`` key -- so the carried block goes through the shared shape check
+    first. It used to compare the *raw* ``primary_column`` against the surviving
+    keys, and ``123`` is neither ``None`` nor a key, so the whole block was
+    dropped: the file lost its ``crs`` too, and an absent ``crs`` is spec-defined
+    as OGC:CRS84, silently relabelling a projected file lon/lat (#968).
     """
+    # Sanitizing drops a non-string `primary_column` and repairs it from a lone
+    # surviving column, which is the recovery this function needs and #883/#945
+    # already shared out. A block it cannot make usable is an absent one.
+    declared_primary = isinstance(geo_dict, dict) and "primary_column" in geo_dict
+    sanitized = sanitize_geo_metadata(geo_dict)
+    if not isinstance(sanitized, dict):
+        return _DROP_GEO
+    geo_dict = sanitized
+
     col_entries = geo_dict.get("columns")
     if not isinstance(col_entries, dict):
         return geo_dict
@@ -863,6 +879,13 @@ def _prune_geo_dict_to_columns(geo_dict: dict, columns: set[str], repoint_primar
         del col_entries[name]
 
     primary = geo_dict.get("primary_column")
+    if primary is None and declared_primary:
+        # Sanitizing dropped a malformed `primary_column` and could not repair
+        # it, because `columns` held more than one entry at the time. Pruning
+        # may just have left exactly one, which is an unambiguous primary.
+        _repair_primary_column(geo_dict)
+        primary = geo_dict.get("primary_column")
+
     if primary is not None and primary not in col_entries:
         if not (repoint_primary and col_entries):
             return _DROP_GEO
--- a/geoparquet_io/core/geo_metadata.py
+++ b/geoparquet_io/core/geo_metadata.py
@@ -312,10 +312,17 @@ def sanitize_geo_metadata(geo_meta):
     cleaned = geo_meta
 
     primary = geo_meta.get("primary_column")
-    dropped_primary = "primary_column" in geo_meta and not isinstance(primary, str)
+    # An empty string is a string, but it is not a column name -- and it lands in
+    # the same place a non-string does: `"" not in col_entries`, so pruning drops
+    # the whole block and the CRS with it. `carried_column_name` and the spec
+    # check in `validate` both already reject it; this is the third reader and
+    # has to agree, or the write paths corrupt what the check correctly flags.
+    dropped_primary = "primary_column" in geo_meta and not (isinstance(primary, str) and primary)
     if dropped_primary:
         problems.append(
-            f"'primary_column' is {_article(_json_type_name(primary))}, expected a string"
+            "'primary_column' is an empty string, expected a column name"
+            if isinstance(primary, str)
+            else f"'primary_column' is {_article(_json_type_name(primary))}, expected a string"
         )
         cleaned = dict(cleaned)
         cleaned.pop("primary_column")
--- a/geoparquet_io/core/streaming.py
+++ b/geoparquet_io/core/streaming.py
@@ -207,21 +207,33 @@ def find_geometry_column_from_metadata(metadata: dict | None) -> str | None:
     """
     Find the primary geometry column name from metadata.
 
+    A column-name reader shared with read-only callers (``Table.geometry_column``),
+    so it takes the line #945 drew: it guards rather than sanitizes, and refuses
+    to hand back the one thing that cannot be a column name. A carried
+    ``primary_column: 123`` used to travel from here through ``extract`` into
+    ``quote_identifier`` and fail with ``TypeError: argument of type 'int' is
+    not iterable``, four frames from anything a user can act on (#968).
+
+    It also no longer falls back to the literal ``"geometry"``: on a file whose
+    column is called ``geom`` that default names a column that does not exist,
+    which is the silently-wrong-coordinates bug #945's review identified. Every
+    caller asks the file's schema when this reader says nothing, which is the
+    answer a block-less file already gets.
+
     Args:
         metadata: Schema metadata dict (with bytes keys)
 
     Returns:
-        Geometry column name or None if not found
+        Geometry column name or None if the block does not name a usable one
     """
+    from geoparquet_io.core.geo_metadata import carried_column_name, decode_carried_geo
+
     if not metadata or b"geo" not in metadata:
         return None
-    try:
-        geo_meta = json.loads(metadata[b"geo"].decode("utf-8"))
-        if isinstance(geo_meta, dict):
-            return geo_meta.get("primary_column", "geometry")
-    except (json.JSONDecodeError, UnicodeDecodeError):
-        pass
-    return None
+    geo_meta = decode_carried_geo(metadata[b"geo"])
+    if not isinstance(geo_meta, dict):
+        return None
+    return carried_column_name(geo_meta.get("primary_column"))
 
 
 def find_geometry_column_from_table(table: pa.Table) -> str | None:
@@ -572,7 +584,10 @@ def extract_crs_from_table(
     if table.schema.metadata and b"geo" in table.schema.metadata:
         try:
             from geoparquet_io.core.crs_utils import crs_is_explicitly_null, warn_null_crs_once
-            from geoparquet_io.core.geo_metadata import sanitize_geo_metadata
+            from geoparquet_io.core.geo_metadata import (
+                carried_geometry_column,
+                sanitize_geo_metadata,
+            )
 
             geo_bytes = table.schema.metadata[b"geo"]
             # `Table.write()` resolves the CRS through here before it builds any
@@ -581,7 +596,13 @@ def extract_crs_from_table(
             geo_meta = sanitize_geo_metadata(json.loads(geo_bytes.decode("utf-8")))
             if isinstance(geo_meta, dict):
                 columns = geo_meta.get("columns", {})
-                geom_col_name = geometry_column or geo_meta.get("primary_column", "geometry")
+                # The literal `"geometry"` default named a column that does not
+                # exist on a file whose column is `geom`, and reported its CRS
+                # as absent (#887 review). Ask the table's schema instead — the
+                # shared recovery #960 shared out (#968).
+                geom_col_name = geometry_column or carried_geometry_column(
+                    geo_meta, table.column_names
+                )
                 if geom_col_name in columns:
                     if crs_is_explicitly_null(columns[geom_col_name]):
                         warn_null_crs_once(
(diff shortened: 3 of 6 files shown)


cholmes:
## Multi-agent adversarial review — merge-ready; one gap in the lead item fixed in 691f9d8

Reviewed against a pristine `git archive origin/main` tree with its own venv, so every "main vs branch" claim below is a measured diff rather than a reading.

### The lead item is genuinely fixed, not merely non-crashing

That distinction matters here, because **#945 — the second instalment — removed a crash and replaced it with silently wrong coordinates.** So the question at each site was what the recovered path *does*, not whether the traceback is gone.

63 fixtures (columns named `geom`/`geometry`/`wkb_geometry`/`shape`/`footprint`, real EPSG:5070 PROJJSON) × four write commands, both trees:

```
geom__primary_int   extract  main: crs=<absent>  ->  PR: NAD83 / Conus Albers (EPSG:5070)
geom__primary_list  extract  main: TypeError: unhashable type: 'list'  ->  PR: EPSG:5070
geom__primary_dict  extract  main: TypeError: unhashable type: 'dict'  ->  PR: EPSG:5070
```

The PROJJSON is **byte-identical to the input** in all 20 malformed outputs; `primary_column` names a column that exists in every case; `validate_geoparquet(validate_data=True)` passes on all 24 with zero failed checks. And the #945-shaped probe: `add h3`/`add s2`/`add quadkey`/`sort quadkey` outputs for all five malformed shapes are **value-identical to the control** (`h3_cell='8926e461507ffff'`, `s2_cell='87bd0684'`) — no projected file grid-keyed as lon/lat.

**No new crash paths**: 29 commands × 15 fixtures on both trees gave 118 tracebacks on main and 97 here; only-in-PR is **empty**, 21 fixed.

The validation guards produce **no false pass** — every check that failed on main still fails, the five well-formed-but-broken files are byte-identical, and both named cases hold (a 1.0 file declaring a 1.1-only covering still fails `version_features_match`; a well-formed 1.0 file still gets its outdated-version report). Where a crash became a report, the real problems came with it rather than replacing it.

### Fixed before merge: `primary_column: ""` was still corrupting

The PR's own lead defect, from the shape most likely to survive a naive guard — `""` **is** a string, so it passed `isinstance(primary, str)`, then hit `"" not in col_entries` and took the whole block, CRS included:

```
primary_column: ""   extract geoparquet   ->  crs ABSENT (=> spec default OGC:CRS84)
```

Worse than the `123` case beside it: that one warns, while this one saw a valid string and said **nothing**. The three readers disagreed at exactly this value — `carried_column_name("")` returns `None`, `validate._check_primary_column_present` FAILs it, and only the sanitizer accepted it. So `gpio check spec` correctly reported the file as broken while the write paths corrupted it.

Now `crs=5070`, `primary` repointed to the real column, and a warning naming the cause. `primary_column_is_empty` joins the parametrized shape list (flowing into every reader test) plus an explicit end-to-end CRS-preservation case. Sabotage-verified: reverting the guard fails four tests including both new ones.

I flagged the removal of the literal `"geometry"` fallback as a risk — that a caller relied on it and would now get `None`. **It runs backwards: the removal fixes a live crash.**

```
geom__primary_absent  extract -> stdout   main: EXIT=1   PR: rows=3, geom:binary
geom__primary_null    extract -> stdout   main: primary=None (invalid)   PR: primary='geom'
```

All three callers plus `Table._detect_geometry_column` already ask the schema on `None`.

### Two body claims that do not hold

- *"The single-column half is already handled by `_repair_primary_column`"* — a single **non-standard** geometry column still fails: `gpio sort hilbert` on a one-column `footprint` file with `primary_column: 123` gives `BinderException: Referenced column "geometry" not found ... Candidate bindings: "footprint"`. `_repair_primary_column` runs only inside `sanitize_geo_metadata`; `find_primary_geometry_column` reads through the guard and falls through to five standard names. Identical on main — a body-accuracy defect, but the body is the changelog entry.
- *The remaining sweep is "all one shape"* — `core/add/quadkey.py:119` reads `geo_meta.get("columns", {})` raw and dies on **six** malformed shapes that are not the `version` shape named. So batch five is at least two sites.

### Follow-ups, not blockers

The two-survivor prune path emits a block with no `primary_column` (REQUIRED by spec); downstream gpio recovers it, so the blast radius is a non-gpio consumer. `check bbox` on a non-object block reports "Version 0.0.0, upgrade recommended" and suggests a command that then dies — a misleading diagnosis rather than a false pass. `check all` still dies on every non-object block.

**And there is still no mechanical guard.** The body concedes this honestly, and I agree `common.py` belongs in its own PR — but a seventh site turned up in a single sweep, so the `meta`-lane grep test #968 asked for is now overdue rather than optional. I will keep #968 open scoped to that.

### Gates

Fast suite **7122 passed, 76 skipped, 9 xfailed, 0 failed** · `diff-cover --fail-under=90` → **96%** (1 line, the prune loop) · `pytest -m meta` 205 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push clean · 27 of the new tests fail against `origin/main`, parametrized over both `geometry` and `geom` with a real projected CRS — the two conditions whose absence hid #945's bug. The 4 deleted lines in `test_streaming.py` removed an assertion that **encoded the old wrong behaviour**.

Issue #983, opened 2026-09-10
fix(cli): four commands answer a DuckDB-rejected input with a traceback, not an error line


cholmes:
Found while verifying #979. A GeoParquet file whose `geo` block DuckDB itself
refuses is not a gpio bug -- but four commands answer it with a raw Python
traceback instead of an error line.

Reproduced on a clean `origin/main` worktree (82146f0) with a one-row file whose
`geo` block is `{"version": "1.1.0"}` -- valid JSON, a string version, nothing
gpio's own guards object to, and nothing DuckDB will read:

```
$ gpio check spec  file.parquet     ->  Summary: 6 passed, 0 warnings, 2 failed   (clean)
$ gpio check bbox  file.parquet     ->  ❌ No bbox column found                    (clean)

$ gpio check optimization file.parquet
    result = connection.execute(f"""
_duckdb.InvalidInputException: Invalid Input Error: Geoparquet metadata does not have a columns object

$ gpio add bbox file.parquet out.parquet
  File ".../core/common.py", line 3879, in add_computed_column
    total_count = con.execute(f"SELECT COUNT(*) FROM {sql_path(input_path)}").fetchone()[0]
_duckdb.InvalidInputException: Invalid Input Error: Geoparquet metadata does not have a columns object

$ gpio sort hilbert file.parquet out.parquet
    empty_count_result = con.execute(f"""
_duckdb.InvalidInputException: Invalid Input Error: Geoparquet metadata does not have a columns object

$ gpio add quadkey file.parquet out.parquet
  File ".../core/write_strategies/duckdb_kv.py", line 737, in _write_plain_parquet_from_query
    con.execute(copy_query)
_duckdb.InvalidInputException: Invalid Input Error: Geoparquet metadata does not have a columns object
```

`gpio inspect head` and `gpio convert geoparquet` on the same file both answer
`Error: Invalid Input Error: ...` and exit 1, so the handling exists -- it is
just not on these paths. The message DuckDB gives is a good one; only the
traceback around it is wrong.

The trigger is not this particular block: it is any input DuckDB's Parquet
reader refuses, so it is reachable from a file written by any other tool. #979's
non-string `version` is one such input, which is how this surfaced -- with #979
fixed, gpio no longer raises its own `AttributeError` there and DuckDB's refusal
becomes visible underneath. Both behaviours were already on `origin/main`; #979
does not change either.

Refs #979.

[...]

Issue #982, opened 2026-09-10
fix(add): 'add quadkey' crashes on six malformed 'geo' shapes #978 fixed everywhere else


cholmes:
Found by the #979 sweep (12 malformed `geo` shapes x 28 CLI commands), which
widened #978's command list. `gpio add quadkey` reads the carried block raw:

```python
# geoparquet_io/core/add/quadkey.py, _validate_crs_from_geo_metadata
    if not geo_meta:
        ...
        return
    columns_meta = geo_meta.get("columns", {})     # line 119
    if geom_col not in columns_meta:
        ...
        return
    crs_info = columns_meta[geom_col].get("crs")   # line 125
```

Neither the block nor `columns` nor the per-column entry is shape-checked, so
every malformed shape #978 fixed elsewhere still crashes here. Six of them, all
on `gpio add quadkey`, verified on `origin/main` (8ea5f9c):

| carried `geo` | traceback |
|---|---|
| block is an array | `AttributeError: 'list' object has no attribute 'get'` (line 119) |
| block is a string | `AttributeError: 'str' object has no attribute 'get'` (line 119) |
| `columns: null` | `TypeError: argument of type 'NoneType' is not iterable` (line 120) |
| `columns: ["geometry"]` | `TypeError: list indices must be integers or slices, not str` (line 125) |
| `columns: "geometry"` | `TypeError: string indices must be integers, not 'str'` (line 125) |
| `columns: {"geometry": "WKB"}` | `AttributeError: 'str' object has no attribute 'get'` (line 125) |

They were invisible to #978's sweep only because its command list did not carry
`add quadkey`; on the shape #979 fixes, `add quadkey` crashed one frame earlier
in `common.py` and so never reached this function at all.

This is a write path -- the block it reads is the one `add quadkey` carries into
its output -- so it takes #978's line rather than #968's: come through
`sanitized_carried_geo` / `sanitize_geo_metadata` instead of reading the raw
block, and the shape check happens once, in the place that already owns it.

#979's sweep test pins these six as `xfail(strict=True)` with this issue's
number, so whoever fixes them will be told to unmark them.

Refs #979, #978, #968.

*Filed by Claude (AI); reproduced on a clean `origin/main` worktree before filing.*

PR #984, opened 2026-09-10
fix(core): report a non-string geo.version in check spec and check bbox instead of crashing

### LINKED ISSUE #979: fix(core): a non-string geo.version crashes six command paths with a raw traceback
Found by the malformed-metadata sweep in PR #978 (#968), which deliberately left it unfixed because it straddles a file that PR does not own. Verified in the current tree before filing.

## The gap

Four PRs (#890, #954, #968 and their siblings) have now hardened gpio against hostile `geo` metadata. A sweep of 8 malformed shapes × 26 CLI commands, run against a pristine `git archive origin/main` tree and against #978's branch:

| tree | raw `TypeError` / `AttributeError` |
|---|---|
| `origin/main` | 12 |
| #978's branch | 6 |

**All six survivors are one shape: a `version` that is not a string.** Two sites:

**`core/common.py:197`** (`detect_geoparquet_file_type`)
```python
version = result["geo_version"]
if version and version.startswith("2."):
```
**`core/streaming.py:624`** (`extract_version_from_metadata`)
```python
version = geo_meta.get("version")
if version:
    parts = version.split(".")
...
except (json.JSONDecodeError, UnicodeDecodeError):
```
Both guard only truthiness, so any truthy non-string — `2`, `2.0` as a float, `["2.0"]`, a dict — reaches `.startswith` / `.split` and raises. In `streaming.py` the surrounding `except` catches JSON and Unicode errors specifically, so the `AttributeError` passes straight through it.

Six live command paths:

- via `common.py`: `check spec`, `check bbox`, `check optimization`, `add bbox`
- via `streaming.py`: `sort hilbert`, `add quadkey`

## Why one issue and not two

The natural instinct is to fix the `streaming.py` half wherever that file is already open. Don't: `common.py` accounts for four of the six paths, so a half-fix leaves `check spec` still dumping a traceback on the very same file — a worse outcome than the current uniform failure, because it looks fixed. Fix both sites in one change.

## Shape of the fix

Follow what the four prior PRs settled on — a type guard that degrades to "unknown version" rather than raising, so a malformed file gets a *validation finding* instead of a traceback. This is the same family as #968's `primary_column` handling: `check`'s job is to report that a file is malformed, so crashing on the malformed input is the one thing it must not do.

Worth pairing with a test that walks the malformed shapes across the command surface, rather than one test per site — that sweep is what found this, and a per-site test would not have.

Refs #968, #978, #890, #954.

*Filed by Claude (AI); both call sites re-read in the current tree before filing.*

cholmes:
## What changed

The fifth instalment of #883's class, and the one all four earlier PRs walked
past because they all read the same two keys. This is a **third** key: a
`version` that is not a string.

Every reader of `geo.version` parses it — `version.startswith("2.")`,
`version.split(".")` — and each guarded truthiness only:

```python
version = geo_meta.get("version")
if version and version.startswith("2."):     # common.py
if version: parts = version.split(".")       # streaming.py
```

so `2`, `2.0`, `["2.0"]`, `{"major": 2}` and `true` all sail past the guard. In
`streaming.py` the `except` beside the call names `json.JSONDecodeError` and
`UnicodeDecodeError` specifically, so the `AttributeError` passed straight
through it.

One shared guard, `geo_metadata.carried_version`, sitting next to
`carried_column_name` and taking the same line #945 drew for it: hand back
nothing rather than a value that cannot be a version, name the ignored key once,
and let each caller fall through to the default it already has for a file that
declares no version. Version readers are shared between the write paths and
`gpio check`, so they **guard, they do not sanitize** — `gpio check` has to see
the file as it really is. Degrading to "unknown version" is the whole point:
`check` is the command a user runs to be *told* their file is malformed.

| # | Site | Reached by |
|---|---|---|
| 1 | `common.py` `detect_geoparquet_file_type` | `check spec`, `check bbox`, `check optimization`, `add bbox` |
| 2 | `streaming.py` `extract_version_from_metadata` | `sort hilbert`, `sort str`, `add quadkey` |
| 3 | `common.py` `_detect_version_from_table` | `write_geoparquet_table`, `gpio.read(f).write(out)` |
| 4 | `check_parquet_structure.py` `_check_geoparquet_v1` | `check bbox`, `check bbox --fix` |

**3 and 4 are not in the issue, and both are load-bearing.** #979 named two
sites; the sweep found two more, and skipping either would have reproduced
exactly the half-fix the issue warns about:

- **3** is the API-side twin of 1, and it is a *write* path — `write_geoparquet_table`
  resolves auto-mode through it. Adding the shape to the shared `MALFORMED_BLOCKS`
  list the earlier PRs built found it immediately, in an existing test.
- **4** was *masked*. `check bbox` crashed in `detect_geoparquet_file_type` one
  frame earlier, so it never reached `version < "1.1.0"` — a comparison a number
  cannot make. Fixing site 1 alone would have moved `check bbox`'s traceback one
  file along rather than removing it. That, one layer down, is the issue's own
  argument for why this is one change.

### The six paths, reproduced

Same one-row file throughout, `"version": 2`, everything else well formed. On
`origin/main`, 6 of 6 die:

```
$ gpio check spec / check bbox / check optimization / add bbox  <file>
  File ".../core/common.py", line 197, in detect_geoparquet_file_type
    if version and version.startswith("2."):
AttributeError: 'int' object has no attribute 'startswith'

$ gpio sort hilbert <file> out.parquet
  File ".../core/hilbert_order.py", line 83, in _resolve_output_version
    return extract_version_from_metadata(metadata) or DEFAULT_GEOPARQUET_VERSION
  File ".../core/streaming.py", line 645, in extract_version_from_metadata
    parts = version.split(".")
AttributeError: 'int' object has no attribute 'split'

$ gpio add quadkey <file> out.parquet
  File ".../core/common.py", line 3604, in get_bbox_advice
    file_info = detect_geoparquet_file_type(parquet_file, verbose)
AttributeError: 'int' object has no attribute 'startswith'
```

After: **0 of 6 `AttributeError`s** -- which is what this PR set out to remove,
but it is not the same as 0 of 6 tracebacks. Measured on the branch, the same
file, the same six commands:

| command | before | after |
|---|---|---|
| `check spec` | `AttributeError` traceback | **clean report, rc 1** |
| `check bbox` | `AttributeError` traceback | **clean, rc 0** |
| `add bbox` | `AttributeError` traceback | `_duckdb.InvalidInputException` traceback |
| `sort hilbert` | `AttributeError` traceback | `_duckdb.InvalidInputException` traceback |
| `add quadkey` | `AttributeError` traceback | `_duckdb.InvalidInputException` traceback |
| `check optimization` | `AttributeError` traceback | `_duckdb.InvalidInputException` traceback |

DuckDB's own Parquet reader rejects **any** non-string `geo.version`
(`Geoparquet metadata does not have a version`), so the four DuckDB-backed paths
cannot read the file at all whatever gpio's guards do. They trade one traceback
for another; the user-visible fix is `check spec` and `check bbox`. Turning those
four into an error line is **#983**, which this PR does not do -- land it with
this and the sentence becomes true for all six.

`gpio check spec` now answers with the finding the command exists to produce, and
names the key it ignored:

```
Ignoring malformed 'geo' metadata on the input of file.parquet: 'version' is a number, expected a string
  ✗ metadata must include a "version" string
  ✗ version must be a string (got int: 2)
      Known versions are 1.x and 2.x
  ○ version is not a string; feature check not applicable
```

and `gpio check bbox` reports `⚠️ Version 0.0.0 (upgrade to 1.1.0+ recommended)`
— a version nobody can read is no version, which is what `0.0.0` already says
here for a block that declares none.

### Reusing what #978 built

Checked before writing anything: `sanitize_geo_metadata` does **not** touch
`version`, and should not. It is the write-path sanitizer, and every write path
*overwrites* `version` from `_initialize_geo_metadata` — a carried version never
reaches an output file. The reusable thing was the shape one layer down:
`carried_column_name`'s guard-and-name-the-key idiom, `_json_type_name` and
`_article` for the message, and `_emit_malformed_geo_warning`'s warn-once-per-
`(detail, source)` cache. `carried_version` is 8 lines because all of that was
already there.

## Verification

**A sweep, not a claim.** The measure #978 reported, re-run on this tree: 12
malformed shapes x 28 CLI commands (its 26 plus `add quadkey` and `sort str`),
against a pristine `origin/main` worktree and against this branch with the same
script.

| | 8 shapes x 28 commands | 4 extra `version` shapes x 28 |
|---|---|---|
| `origin/main` | **13** | **28** |
| this branch | **6** | **0** |

Every one of the 41 that go is a `version` crash. **The 6 that remain are all
one command and none of them are `version`** — see below; they are pre-existing,
filed, and pinned by the new sweep test rather than left implicit.

The sweep is now **committed**, in `tests/test_malformed_geo_metadata.py`, which
is the part that matters: none of the ~40 per-site tests the four earlier PRs
wrote would have found #979, because the crash was in a key none of them was
written for. It asserts the *set* of commands that die per shape against a
`KNOWN_UNGUARDED` table, so it fails in both directions — a new unguarded reader
fails it, and so does fixing a known one without striking it off the list. The
non-string `version` shape also joins the shared `MALFORMED_BLOCKS` list, so
every existing per-site test now runs against it (that is what found site 3).

- `uv run pytest -n auto -m "not slow and not network and not meta"` — **7431 passed**, 76 skipped, 9 xfailed
- `uv run pytest -m meta` — **205 passed**
- `uv run pre-commit run --all-files` clean; `--hook-stage pre-push` clean, `xenon`, `mypy`, `import-linter` and `menard-check` included
- `diff-cover coverage.xml --compare-branch=origin/main` — **14 changed lines, 0 missing, 100%**
- 63 new tests

One flake, named rather than hidden: `tests/e2e/test_journeys.py::test_journey_01_geojson_to_geoparquet_and_back`
failed once in the ~7,500-test `-n auto --cov` run on this shared machine. It
passes serially, passes under `-n auto --cov` in isolation on `origin/main` too,
and the identical instrumented full run immediately after was 7431/7431 — the
saturation noise CLAUDE.md documents, not this change.

## What the sweep found that this PR does not fix

Reported rather than quietly bundled, per the repo's separate-PR convention.
Both were verified on a clean `origin/main` worktree before filing, and neither
is touched by this change.

- **#982** — the 6 remaining sweep crashes, all `gpio add quadkey`.
  `_validate_crs_from_geo_metadata` reads `geo_meta.get("columns", {})[col].get("crs")`
  with no shape check at all, so every shape #978 fixed elsewhere still crashes
  there. Invisible to #978's sweep only because its command list did not carry
  `add quadkey`; invisible on the `version` shape because `add quadkey` used to
  crash a frame earlier, in `common.py`. It is a write path, so its fix is
  `sanitized_carried_geo`, not another guard — a different batch. The sweep test
  pins all six against this issue number.
- **#983** — `check optimization`, `add bbox`, `sort hilbert` and `add quadkey`
  answer any input DuckDB's Parquet reader refuses with a raw
  `_duckdb.InvalidInputException` traceback instead of an error line, while
  `inspect head` and `convert geoparquet` on the same file print
  `Error: Invalid Input Error: ...` and exit 1. Reproduced on `origin/main` with
  a block DuckDB rejects for an unrelated reason (`{"version": "1.1.0"}`), so it
  is not caused by this change — but with #979's `AttributeError` gone, DuckDB's
  refusal is what shows underneath on those four paths.

Closes #979

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of malformed or non-string GeoParquet version metadata.
  * Prevented crashes during file inspection, validation, version detection, rewriting, and command execution.
  * Invalid version values now use existing default-version behavior and produce a clear metadata warning.

* **Tests**
  * Expanded coverage for malformed version values across GeoParquet operations and command-line workflows.

### DIFF
--- a/geoparquet_io/core/check_parquet_structure.py
+++ b/geoparquet_io/core/check_parquet_structure.py
@@ -556,7 +556,10 @@ def _check_geoparquet_v1(parquet_file, file_type_info, verbose, return_results, 
     # make (#979). That crash sat behind the `detect_geoparquet_file_type` one
     # `check bbox` hit a frame earlier, so it only becomes reachable once that
     # is guarded -- fix one without the other and the traceback just moves.
-    version = carried_version(geo_meta.get("version") if isinstance(geo_meta, dict) else None)
+    version = carried_version(
+        geo_meta.get("version") if isinstance(geo_meta, dict) else None,
+        source=str(parquet_file),
+    )
     version = version or "0.0.0"
     bbox_info = check_bbox_structure(parquet_file, verbose)
 
--- a/geoparquet_io/core/check_parquet_structure.py
+++ b/geoparquet_io/core/check_parquet_structure.py
@@ -541,7 +541,7 @@ def _check_geoparquet_v2(parquet_file, file_type_info, verbose, return_results, 
 def _check_geoparquet_v1(parquet_file, file_type_info, verbose, return_results, quiet=False):
     """Check GeoParquet 1.x file (existing logic, bbox IS recommended)."""
     from geoparquet_io.core.duckdb_metadata import get_geo_metadata
-    from geoparquet_io.core.geo_metadata import covering_supported
+    from geoparquet_io.core.geo_metadata import carried_version, covering_supported
 
     geo_meta = get_geo_metadata(parquet_file)
     # `get_geo_metadata` is a read-only reader: it hands the block back exactly
@@ -550,7 +550,14 @@ def _check_geoparquet_v1(parquet_file, file_type_info, verbose, return_results, 
     # (#968). A validation reader guards and reports the truth rather than
     # sanitizing: a block that declares no version has none, which is what
     # "0.0.0" says and what the outdated-version issue below reports.
-    version = geo_meta.get("version", "0.0.0") if isinstance(geo_meta, dict) else "0.0.0"
+    #
+    # A version that is not a *string* is the same answer, and has to be: the
+    # comparison two lines down is `version < "1.1.0"`, which a number cannot
+    # make (#979). That crash sat behind the `detect_geoparquet_file_type` one
+    # `check bbox` hit a frame earlier, so it only becomes reachable once that
+    # is guarded -- fix one without the other and the traceback just moves.
+    version = carried_version(geo_meta.get("version") if isinstance(geo_meta, dict) else None)
+    version = version or "0.0.0"
     bbox_info = check_bbox_structure(parquet_file, verbose)
 
     # Build results
--- a/geoparquet_io/core/common.py
+++ b/geoparquet_io/core/common.py
@@ -45,6 +45,7 @@ from geoparquet_io.core.geo_metadata import (
     GEOPARQUET_VERSIONS,
     build_bbox_covering,
     carried_geometry_column,
+    carried_version,
     create_geo_metadata,
     detect_bbox_column_from_schema,
     geoarrow_wkb_codes,
@@ -184,7 +185,12 @@ def detect_geoparquet_file_type(parquet_file, verbose=False, con=None):
     if geo_meta:
         result["has_geo_metadata"] = True
         if isinstance(geo_meta, dict) and "version" in geo_meta:
-            result["geo_version"] = geo_meta["version"]
+            # A version that is not a string is no version this can report:
+            # `2` reached `.startswith` below and took `check spec`,
+            # `check bbox`, `check optimization` and `add bbox` down with a bare
+            # `AttributeError` (#979). `geo_version` is documented as the
+            # version *or None*, and every consumer already handles None.
+            result["geo_version"] = carried_version(geo_meta["version"], source=str(parquet_file))
 
     # Check for native Parquet geo types using schema
     geo_columns = detect_geometry_columns(parquet_file, con=con)
@@ -1020,7 +1026,10 @@ def _detect_version_from_table(table, verbose: bool = False) -> str | None:
     try:
         geo_meta = json.loads(metadata[b"geo"].decode("utf-8"))
         if isinstance(geo_meta, dict):
-            version = geo_meta.get("version")
+            # Guarded, not truthiness-checked: a non-string version crashed
+            # `.split` here on the write path too -- `write_geoparquet_table`
+            # resolves auto-mode through this function (#979).
+            version = carried_version(geo_meta.get("version"))
             if version:
                 parts = version.split(".")
                 if len(parts) >= 2:
--- a/geoparquet_io/core/geo_metadata.py
+++ b/geoparquet_io/core/geo_metadata.py
@@ -168,6 +168,39 @@ def carried_column_name(
     return None
 
 
+def carried_version(value, source: str | None = None) -> str | None:
+    """A carried ``geo.version`` *string*, or ``None`` when the block does not give one.
+
+    Every reader of this key parses it -- ``version.startswith("2.")``,
+    ``version.split(".")`` -- and each of them guarded truthiness only, so a
+    ``version`` that is ``2``, ``2.0``, ``["2.0"]``, ``{"major": 2}`` or ``true``
+    walked past the guard and raised ``AttributeError`` several frames from
+    anything a user can act on (#979). In ``streaming`` the ``except`` beside
+    the call named ``json.JSONDecodeError`` and ``UnicodeDecodeError``
+    specifically, so it did not even catch it.
+
+    Version readers are shared between the write paths and ``gpio check``, so
+    this takes the same line as :func:`carried_column_name`: guard rather than
+    sanitize, hand back nothing rather than a value that cannot be a version,
+    name the ignored key once, and let each caller fall through to the default
+    it already has for a file that declares no version. Degrading to "unknown
+    version" is the whole point -- ``gpio check`` is the command a user runs to
+    be *told* their file is malformed, so crashing on the malformed input is the
+    one thing it must not do.
+
+    An *absent* version is not malformed and says nothing; an empty string is
+    not a version either, and ``validate`` already FAILs it as one.
+    """
+    if isinstance(value, str):
+        return value or None
+    if value is None:
+        return None
+    _emit_malformed_geo_warning(
+        f"'version' is {_article(_json_type_name(value))}, expected a string", source
+    )
+    return None
+
+
 def decode_carried_geo(raw):
     """Decode a raw carried ``geo`` value (bytes or str) to JSON, or None.
 
--- a/geoparquet_io/core/streaming.py
+++ b/geoparquet_io/core/streaming.py
@@ -637,10 +637,16 @@ def extract_version_from_metadata(metadata: dict | None) -> str | None:
     """
     if not metadata or b"geo" not in metadata:
         return None
+    from geoparquet_io.core.geo_metadata import carried_version
+
     try:
         geo_meta = json.loads(metadata[b"geo"].decode("utf-8"))
         if isinstance(geo_meta, dict):
-            version = geo_meta.get("version")
+            # A truthiness guard let `2`, `2.0`, `["2.0"]` and `true` through to
+            # `.split`, and the `except` below names the two decoder errors
+            # specifically, so the `AttributeError` passed straight through it
+            # and out of `gpio sort hilbert` as a traceback (#979).
+            version = carried_version(geo_meta.get("version"))
             if version:
                 parts = version.split(".")
                 if len(parts) >= 2:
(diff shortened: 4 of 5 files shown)


cholmes:
## Multi-agent adversarial review — approved, two items fixed in 6d720cb

The engineering is sound. **Both S1-shaped hypotheses I briefed against were disproven with evidence**, which is worth recording because they are the two shapes that have bitten this class before:

- **A recovered version does not produce a mislabelled file.** A `version: 2` input downgrades 2.0 → 1.1, but the output is self-consistent — genuinely 1.1 WKB, labelled 1.1, **CRS 5070 intact**, no covering silently dropped. Byte-for-byte the same treatment `version: "garbage"` and `version: ""` already got before this PR, and the downgrade is announced. Not the #945 shape.
- **Unmasking surfaces nothing worse.** A 297-run cross-tree diff (27 fixtures × 11 commands, against a pristine `origin/main` worktree) shows the *only* behaviour changes are `check bbox` and `check spec`, both strictly improving. `carried_version` is the identity on every string, so well-formed files are untouched — confirmed empirically on `ok`/`garbage`/`short`/`empty` fixtures across all 11 commands. The one non-identity case, `"" → None`, is inert: all seven `geo_version` consumers treat `""` and `None` the same.

`gpio check spec` also does not produce a false pass — it names the finding specifically (`✗ version must be a string (got int: 2)`) at exit 1.

### Fixed: `check bbox` printed the same warning twice

`_check_geoparquet_v1` called `carried_version` without a `source`, so one copy named the file and one did not, and the `(detail, source)` warn-once cache could not collapse them:

```
Ignoring malformed 'geo' metadata on the input of f.parquet: 'version' is a number, expected a string
Ignoring malformed 'geo' metadata on the input: 'version' is a number, expected a string
```

That defeats the cache in both directions — the duplicate above, and a second malformed file in the same process going silent, which is exactly what `_emit_malformed_geo_warning`'s own docstring warns about. `parquet_file` was already in scope one line up. Now one named warning.

### Corrected: "After: 0 of 6" measured the exception class, not the outcome

Measured on the branch, same file, same six commands:

| command | before | after |
|---|---|---|
| `check spec` | `AttributeError` traceback | **clean report, rc 1** |
| `check bbox` | `AttributeError` traceback | **clean, rc 0** |
| `add bbox` | `AttributeError` traceback | `_duckdb.InvalidInputException` traceback |
| `sort hilbert` | `AttributeError` traceback | `_duckdb.InvalidInputException` traceback |
| `add quadkey` | `AttributeError` traceback | `_duckdb.InvalidInputException` traceback |
| `check optimization` | `AttributeError` traceback | `_duckdb.InvalidInputException` traceback |

DuckDB's own Parquet reader rejects **any** non-string `geo.version` (`Geoparquet metadata does not have a version`), so the four DuckDB-backed paths cannot read the file at all whatever gpio's guards do. They trade one traceback for another.

"0 of 6" is literally true of `TypeError`/`AttributeError` and substantively misleading about what a user sees — and since the title is the changelog entry, it would have shipped as a claim about six commands that holds for two. Title narrowed to name the two genuinely fixed; the body now carries the measured table and points at **#983** for the other four. Land #983 with this and the original sentence becomes true.

### Not fixed here, and worth saying plainly

**There is still no static ratchet.** This is instalment five (#883 → #887/#945 → #947/#960 → #968/#978 → this), and #968 asked for one. What ships instead is `KNOWN_UNGUARDED` plus a runtime sweep — a good net that fails in both directions, but it covers **26 of 59 CLI commands**, and 236 raw `geo`-key reads remain in `geoparquet_io/`. A new raw read in `partition *`, `process *`, `pmtiles *`, `add a5/h3/s2/kdtree`, `convert flatgeobuf/geopackage/shapefile/reproject`, `publish upload`, `inspect layers/stats` or `benchmark *` passes it silently.

The repo already has the right template: `scripts/check_sql_path_literals.py` with its baseline file and `--update` ratchet, built for the #718 class. Applying that shape here is the thing that ends the series. **#982** (`add quadkey`, six shapes) is correctly scoped out and pinned in `KNOWN_UNGUARDED` — though note the issue tells a future fixer to look for `xfail(strict=True)` markers that do not exist in the committed test.

For the `version` key specifically, the sweep does look complete: every raw read in the tree was classified, and the one candidate sixth-instalment crash (`inspect_utils.py:1259`, unhashable `dict` version) is unreachable behind `extract_partition_summary`'s per-file `try/except … continue`.

### Gates

Fast suite **7429 passed, 76 skipped, 9 xfailed** — 2 failures on the known cold-run flake list, **102 passed serially** · `diff-cover --fail-under=90` → **100% on 14 lines** · `pytest -m meta` 205 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit clean · the branch test file run against `origin/main` source gives **67 failed, 575 passed**, so every new assertion family genuinely fails on main, and the file is fast-lane with no markers.

Issue #971, opened 2026-09-10
fix(api): Table.write(row_group_rows=50000) writes 51,200-row groups while gpio sort --row-group-size 50000 writes 49,152


cholmes:
## What

After #967, the sort **CLI** snaps `--row-group-size` to a whole 2,048-row writer vector and announces the adjustment, so the number gpio asks for is the number that lands. The **Python API** does not:

```python
gpio.read(f).sort_hilbert().write(out, row_group_rows=50000)
```

hands 50,000 straight to the writer, which rounds it up. Measured on 200,000 rows: row groups `[51200, 51200, 51200, 46400]` — outside the 10,000–50,000 spatial band that `gpio check optimization` scores, so `check` fails a file the API just wrote. That is exactly the #961 shape, one layer over.

So `row_group_size = 50000` now means **49,152** for `gpio sort` and **51,200** for `gpio convert` and for `Table.write`.

## Why it was not fixed in #967

#967 put the alignment in `resolve_sort_row_group_rows`, the one place all four sort subcommands route through. `Table.write()` is the shared general write path used by every command, so snapping there is a behaviour change for `convert`, `add`, `extract` and everything else — a different decision, deliberately not made under a `fix(sort)!` title. #967 documents the split honestly in `docs/guide/sort.md` and `docs/guide/check.md` rather than papering over it.

## Options

1. **Snap in `Table.write` / `ParquetWriteSettings`**, so every write path agrees and the requested number always lands. Smallest surprise for users; a behaviour change for every non-sort command.
2. **Snap nowhere** and let the writer round everywhere, reverting the sort commands to advertise the landed number instead. Consistent, but re-opens #961.
3. **Leave the split** and have `check` stop treating a writer-rounded 51,200 as out of band — i.e. compare against the aligned value rather than the raw average.

Option 1 is probably right, but it wants its own PR and its own `!` if it lands.

Note the API also gets no `info()` line, so a Python user is not told the value moved even where it does.

Found by an adversarial review of #967.

PR #987, opened 2026-09-11
fix(convert): name the row-group size the writer will use when --row-group-size is rounded up

### LINKED ISSUE #986: convert: --row-group-size is rounded up to a 2,048-row vector without a word
## Bug Description

`gpio convert geoparquet --row-group-size N` writes row groups of a different size than `N` and prints nothing about it. DuckDB's Parquet writer emits whole 2,048-row vectors and rounds a request up to a multiple of that. `gpio sort` snaps the value itself and prints one line that names the requested and the written size (#961). `convert` hands the value to the writer and stays silent, so a user reads the size they asked for in the command and a different one in the footer.

`docs/guide/convert.md` documents the rounding. The runtime says nothing.

## Steps to Reproduce

```
gpio convert geoparquet tl_2025_us_county.shp out.parquet --row-group-size 249
python -c "import pyarrow.parquet as pq; m=pq.ParquetFile('out.parquet').metadata; print([m.row_group(i).num_rows for i in range(m.num_row_groups)])"
```

Output: `[2048, 1187]`. The same happens for 100, 405 and 647. A request of 4,667 rows on a 23,334-row FileGDB writes `[6144, 6144, 6144, 4902]`.

`tl_2025_us_county.shp` is the US Census TIGER 2025 county shapefile, https://www2.census.gov/geo/tiger/TIGER2025/COUNTY/.

## Expected Behavior

One line at write time, like the one `gpio sort` prints:

```
Writing 2,048 rows per row group rather than the requested 249: the Parquet writer emits row groups in whole 2,048-row vectors, so 249 is not a size a row group can have.
```

A request that is already a multiple of 2,048 prints nothing.

## Actual Behavior

The file is written with 2,048-row groups. No message names the change.

## Environment

- gpio 1.5.0, `main` at 82146f0
- Linux

## Would you like to submit a PR?

Yes. The note goes where the DuckDB `COPY` options are built, so every DuckDB-backed write path prints it and the pyarrow paths, which write exact sizes, do not.

cayetanobv:
## PR Title
fix(convert): name the row-group size the writer will use when --row-group-size is rounded up

## Description
Rebased onto `main` after [#990](https://github.com/geoparquet/geoparquet-io/pull/990) landed the write facade. The defect is still live on `main` (`1473251`), just smaller than it was, and the fix is now re-derived against the facade instead of alongside it.

`resolve_row_group_rows` snaps any request **above** one 2,048-row vector and prints a line naming both numbers. It deliberately passes a request of **2,048 rows or fewer** straight through, because `pq.write_table` honours such a request exactly and snapping it would override the only writer that could obey.

That passthrough leaves a hole on the writers that *cannot* obey. Measured on `main`, 10,000-row input:

```
$ gpio convert geoparquet in.parquet out.parquet --row-group-size 249
row groups: [2048, 2048, 2048, 2048, 1808]      # note printed: none
```

The user asks for 249, gets 2,048, and is told nothing — the same shape as #961, in the window the facade left open.

## Technical Details
- `parquet_writer.py`: `note_duckdb_copy_rounding(row_group_rows)`. It reuses `align_to_writer_vector` for the arithmetic and a new `_rounding_note(requested, landed)` for the sentence `resolve_row_group_rows` already prints — **no second rounding rule, no second wording**. Since resolve has already announced anything above one vector, the funnel note fires only for a sub-vector request and can never repeat one (pinned by a test).
- Called from the three funnels that know the write reaches DuckDB's `COPY`: `write_parquet_with_metadata`'s plain-COPY fast path, its `duckdb-kv` strategy branch, and `Table._write`'s `duckdb-kv` branch — the last so the Python API and the CLI do not disagree about what a number means ([#971](https://github.com/geoparquet/geoparquet-io/issues/971)).
- `None` says nothing, so a `--row-group-size-mb` target never gets a row count the user did not type named back at them.

### What lands per write strategy — measured, `--row-group-size 249`, 10,000-row input

| path | landed | note |
|---|---|---|
| plain `COPY` fast path (2.0 / parquet-geo-only output) | 2,048 | ✅ now printed |
| `duckdb-kv` (the default, CLI and `Table.write`) | 2,048 | ✅ now printed |
| `disk-rewrite` | 249 | silent — correct |
| `in-memory` | 249 | silent — correct |
| `streaming` | 249 | silent — correct |

The original version of this PR also noted on `disk-rewrite`, on the reasoning that it writes through `COPY`. Measured, that is wrong: its `COPY` output is a temp file that its pyarrow pass then regroups at the requested size, so `disk-rewrite` really does write 249-row groups. Telling that user "2,048 will land" would have been a new false statement. Scoped to the two writers that round.

### #990's convert doc overclaimed
`docs/guide/convert.md` said `--row-group-size` "is snapped to the writer's 2,048-row vector the same way the sort commands snap it, so the number you ask for is the number that lands." Below one vector that last clause is false on the DuckDB path — and, because the passthrough is deliberate, *what lands genuinely depends on which writer runs*. The guide now states the exception, says why gpio cannot snap it away, and points at this issue.

## Breaking Changes
**Does this PR introduce breaking changes?** No. No written file changes; only the message is new.

## Related Issue(s)
- Closes #986
- Corrects a doc claim from #990

## Verification
After, same input and command:

```
$ gpio convert geoparquet in.parquet out.parquet --row-group-size 249
Writing 2,048 rows per row group rather than the requested 249: the Parquet writer emits row groups in whole 2,048-row vectors, so 249 is not a size a row group can have.
row groups: [2048, 2048, 2048, 2048, 1808]
```

Full sweep on this branch (`note` column = whether the line is printed):

| request | landed | note |
|---|---|---|
| 1 | 2,048 | yes |
| 249 | 2,048 | yes |
| 1,000 | 2,048 | yes |
| 2,048 | 2,048 | **no** |
| 3,000 | 4,096 | yes (from `resolve`, unchanged) |
| 4,096 | 4,096 | **no** |
| 50,000 | 49,152 | yes (from `resolve`, unchanged) |

```
uv run pytest -n auto -m "not slow and not network and not meta"   # 7592 passed, 76 skipped, 7 xfailed
uv run pytest -m meta                                              # 205 passed
uv run pre-commit run --all-files                                  # all passed
uv run pre-commit run --all-files --hook-stage pre-push            # all passed (incl. menard-check, xenon, import-linter, mypy)
uv run diff-cover coverage.xml --compare-branch=origin/main        # 100% (15/15 lines)
```

`tests/data/cli_surface.json` is unchanged.

## Checklist
- [x] Integration tests for changes that cross module/format/service boundaries (the per-strategy tests write real files and read the footers back)
- [x] All tests pass (`uv run pytest`) for the touched suites; the full suite runs in CI
- [ ] At least one adversarial review (AI counts, rubber stamps don't)

## Summary by CodeRabbit

- **Documentation**
  - Updated `convert` guidance with the default row-group size of 49,152 rows.
  - Documented row-group rounding behavior, including DuckDB’s 2,048-row minimum and related notifications.
  - Clarified curved-geometry handling when skipping Hilbert ordering, including the option to preserve prior errors.

- **Bug Fixes**
  - Row-group requests of 2,048 rows or fewer are preserved where supported.
  - DuckDB-backed conversions and Python API writes now accurately report when requested sizes are rounded.

### DIFF
--- a/geoparquet_io/core/common.py
+++ b/geoparquet_io/core/common.py
@@ -74,6 +74,7 @@ from geoparquet_io.core.logging_config import (
 )
 from geoparquet_io.core.parquet_writer import (
     ParquetWriteSettings,
+    note_duckdb_copy_rounding,
     resolve_output_geoparquet_version,
     resolve_row_group_rows,
 )
@@ -3038,6 +3039,11 @@ def write_parquet_with_metadata(
                 debug(f"Writing GeoParquet version: {effective_version}")
                 debug(f"No metadata rewrite needed for {effective_version} - using plain COPY TO")
 
+            # Nothing but a bare DuckDB COPY writes this file, so a sub-vector
+            # request lands at 2,048 whatever was asked for. Say so here, where
+            # that is certain, rather than let the footer be the first to tell
+            # the user (#986).
+            note_duckdb_copy_rounding(row_group_rows)
             _plain_copy_to(
                 con=con,
                 query=query,
@@ -3135,6 +3141,14 @@ def write_parquet_with_metadata(
             }
             if strategy_enum == WriteStrategy.DUCKDB_KV:
                 write_kwargs["memory_limit"] = memory_limit
+            # duckdb-kv alone lets DuckDB's COPY have the last word on row-group
+            # size, so it alone rounds a sub-vector request up to 2,048 (#986).
+            # Measured with --row-group-size 249 on a 10,000-row input:
+            # duckdb-kv writes 2,048-row groups; in-memory, streaming and
+            # disk-rewrite all write 249-row ones -- disk-rewrite because its
+            # pyarrow pass regroups the COPY output at the requested size.
+            if strategy_enum == WriteStrategy.DUCKDB_KV:
+                note_duckdb_copy_rounding(row_group_rows)
 
             strategy.write_from_query(**write_kwargs)
 
--- a/geoparquet_io/core/parquet_writer.py
+++ b/geoparquet_io/core/parquet_writer.py
@@ -167,6 +167,12 @@ def resolve_row_group_rows(
     ceiling-everywhere rule: it answers "what would the DuckDB writer do", and
     that answer is still 2,048.
 
+    Passing the request through is not the same as pretending it lands. A
+    DuckDB-backed write still produces 2,048-row groups for a request of 249,
+    and used to do it without a word, so the user read 249 in the command and
+    2,048 in the footer (#986). :func:`note_duckdb_copy_rounding` says it, at
+    the funnels that know the write reaches DuckDB's ``COPY``.
+
     A row count below 1 is rejected, not clamped: ``sort str`` raised on
     ``--row-group-size -5`` while the other three quietly wrote 2,048-row groups.
     ``param_name`` is what that rejection blames. One function now serves both
@@ -185,13 +191,55 @@ def resolve_row_group_rows(
 
     aligned = align_to_writer_vector(row_group_rows)
     if aligned != row_group_rows:
-        info(
-            f"Writing {aligned:,} rows per row group rather than the requested "
-            f"{row_group_rows:,}: the Parquet writer emits row groups in whole "
-            f"{WRITER_VECTOR_ROWS:,}-row vectors, so {row_group_rows:,} is not a "
-            "size a row group can have."
-        )
+        info(_rounding_note(row_group_rows, aligned))
     return aligned
+
+
+def _rounding_note(requested: int, landed: int) -> str:
+    """The one sentence gpio says when a row-group request does not land.
+
+    One string, so the note the sort commands print when *gpio* snaps a value
+    and the note the DuckDB funnels print when the *writer* does are the same
+    sentence. They are the same fact about the same file.
+    """
+    return (
+        f"Writing {landed:,} rows per row group rather than the requested "
+        f"{requested:,}: the Parquet writer emits row groups in whole "
+        f"{WRITER_VECTOR_ROWS:,}-row vectors, so {requested:,} is not a "
+        "size a row group can have."
+    )
+
+
+def note_duckdb_copy_rounding(row_group_rows: int | None) -> None:
+    """Say what DuckDB's ``COPY`` will make of an already-resolved row count.
+
+    :func:`resolve_row_group_rows` announces every value *it* moves, so by the
+    time a write funnel calls this the only request still able to surprise
+    anyone is a sub-vector one: resolve passes 1-2,047 through untouched
+    because ``pq.write_table`` honours it exactly, and DuckDB's ``COPY`` then
+    rounds it up to 2,048 regardless. The note therefore fires for exactly that
+    window, and calling it after resolve is silent for everything else --
+    including a resolve that already spoke, whose value is a whole vector by
+    construction.
+
+    Which writer a request reaches is not something the facade can see, so this
+    is deliberately a *funnel* decision rather than part of resolve. Measured on
+    a 10,000-row input with ``--row-group-size 249``: the plain ``COPY`` fast
+    path and the ``duckdb-kv`` strategy both write 2,048-row groups, while
+    ``in-memory``, ``streaming`` and ``disk-rewrite`` write 249-row groups --
+    ``disk-rewrite`` because its pyarrow pass regroups the ``COPY`` output at
+    the requested size afterwards. Call this only where DuckDB's ``COPY`` has
+    the last word on the file the user gets.
+
+    ``None`` -- no row count asked for, or only a ``--row-group-size-mb``
+    target -- says nothing: naming a row count the user never typed would be
+    the inverse of the fix (#986).
+    """
+    if row_group_rows is None:
+        return
+    landed = align_to_writer_vector(row_group_rows)
+    if landed != row_group_rows:
+        info(_rounding_note(row_group_rows, landed))
 
 
 def resolve_output_geoparquet_version(
--- a/tests/test_convert_row_group_note.py
+++ b/tests/test_convert_row_group_note.py
@@ -1,0 +1,134 @@
+"""A DuckDB-backed write says what it will do with a sub-vector request (#986).
+
+The write facade snaps any request above one 2,048-row vector and announces the
+change (``resolve_row_group_rows``, #961/#990). It deliberately passes a request
+of 2,048 rows *or fewer* through untouched, because ``pq.write_table`` honours
+such a request exactly and snapping it would override the only writer that could
+have obeyed.
+
+That passthrough left a hole. DuckDB's ``COPY`` rounds a sub-vector request up
+to 2,048 anyway, so ``gpio convert --row-group-size 249`` wrote 2,048-row groups
+and said nothing: the user read 249 in the command and 2,048 in the footer.
+``note_duckdb_copy_rounding`` closes it, at the funnels that know the write
+reaches ``COPY``, with the same sentence the sort commands print.
+
+The per-strategy divergence the tests below pin is measured, not assumed: on a
+10,000-row input a request of 249 produces 2,048-row groups under ``duckdb-kv``
+and 249-row groups under ``in-memory``, ``streaming`` and ``disk-rewrite``.
+"""
+
+from __future__ import annotations
+
+import pytest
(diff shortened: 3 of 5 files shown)


cayetanobv (commented):
**Recommendation: approve once two things are fixed** — the note repeats once per partition file, and a sub-1 request gets a nonsense note before DuckDB rejects it. Both are small and inline below.

What I checked: the ceiling math against the measured writer behaviour (the table in `test_ceiling_is_the_writers_own_rounding` matches the footers `test_sort_row_group_default.py` reads back), the two call sites against the strategy enum (only `duckdb-kv` and `disk-rewrite` pass `ROW_GROUP_SIZE` to a `COPY`; the arrow strategies write the requested size exactly, so silence there is right), the sort commands (they arrive with an aligned value, so no double note — confirmed), and `tests/test_convert_row_group_note.py` + `tests/test_sort_row_group_default.py` locally on this branch (all pass).

One observation, not a change request: the note now fires for every DuckDB-backed write, not just `convert` — `add`, `extract`, `reproject`, `partition` all route through `write_parquet_with_metadata`. That is the right place for it, but the PR title and the docs sentence only mention `convert`. Worth a line in the PR description so the changelog entry is not read as convert-only.

geoparquet_io/core/common.py
```diff
@@ -3111,6 +3115,11 @@ def write_parquet_with_metadata(
             }
             if strategy_enum == WriteStrategy.DUCKDB_KV:
                 write_kwargs["memory_limit"] = memory_limit
+            if strategy_enum in (WriteStrategy.DUCKDB_KV, WriteStrategy.DISK_REWRITE):
+                # Both write through DuckDB's COPY, which rounds an explicit row
+                # count up to a whole vector. The arrow strategies write the
+                # requested size exactly, so they get no note (#986).
+                note_writer_vector_rounding(row_group_rows)
```
cayetanobv: This fires once per output file, and the partition commands write every final partition through this call (`partition/staging.py::write_partition_file`, `original_metadata` set, so always the strategy path with `duckdb-kv`). Measured on this branch with a 3-partition input:

```
$ gpio partition string pts.parquet parts/ --column bucket --row-group-size 249
...
Writing 2,048 rows per row group rather than the requested 249: the Parquet writer emits row groups in whole 2,048-row vectors, so 249 is not a size a row group can have.
Writing 2,048 rows per row group rather than the requested 249: ...
Writing 2,048 rows per row group rather than the requested 249: ...
Successfully created 3 partition file(s)
```

An h3/quadkey/admin partition with hundreds of cells prints hundreds of identical lines. Two ways out:

1. Snap and announce once at the partition entry point, the way the sort commands do through `resolve_sort_row_group_rows`, and pass the aligned value down. Then this call sees an aligned value and stays silent, and the partition files land at the announced size.
2. Keep the note here but say it once per process (a module-level `set` of requested values already announced, cleared nowhere — the CLI is one process per command).

(1) is the pattern the repo already has. Either way, a test that runs a partition command with `--row-group-size 249` and asserts the line appears exactly once would pin it.

geoparquet_io/core/parquet_writer.py
```diff
@@ -125,15 +125,50 @@ def resolve_sort_row_group_rows(
...
+    )
+
+
+def writer_vector_ceiling(rows: int) -> int:
+    """The row count DuckDB's writer emits for a ``ROW_GROUP_SIZE`` request.
+
+    A strict ceiling to a whole vector, with no band exception: this is what
+    the writer does on its own, not what ``gpio sort`` chooses to ask for.
+    """
+    return -(-rows // WRITER_VECTOR_ROWS) * WRITER_VECTOR_ROWS
```
cayetanobv: `writer_vector_ceiling(-5)` is `0`, and `convert` accepts any `int` for `--row-group-size` (no `min=1` on the option). Measured on this branch:

```
$ gpio convert points.parquet out.parquet --skip-hilbert --row-group-size -5
Converting points.parquet...
Writing 0 rows per row group rather than the requested -5: the Parquet writer emits row groups in whole 2,048-row vectors, so -5 is not a size a row group can have.
Error: Conversion failed: Conversion Error: Type INT32 with value -5 can't be cast because the value is out of range for the destination type UINT64
```

The note promises a 0-row group, then DuckDB refuses the value. `align_to_writer_vector` floors at one vector (`max(1, …)`), and `resolve_sort_row_group_rows` rejects `< 1` with `InvalidParameterError("--row-group-size", "must be at least 1")` before it gets this far. Suggest `note_writer_vector_rounding` do the same `< 1` rejection — then `convert` gets the sort commands' error instead of DuckDB's cast message, and this function never sees a non-positive value. (`0` today writes one 10,000-row group silently, which is also not what the user asked for.)

tests/test_convert_row_group_note.py
```diff
@@ -0,0 +1,81 @@
...
+    result = CliRunner().invoke(
+        cli,
+        ["convert", points_file, str(output), "--skip-hilbert", "--row-group-size", "4096"],
+    )
+    assert result.exit_code == 0, result.output
+    assert "rather than the requested" not in result.output
+    assert _row_group_rows(output) == [4_096, 4_096, ROWS - 2 * 4_096]
+
+
+def test_convert_says_nothing_when_no_size_is_requested(points_file, tmp_path):
```
cayetanobv, line 109: Good coverage of the convert surface. Once the two fixes above land, two more cases belong here: a partition command with `--row-group-size 249` asserting the note appears exactly once, and `--row-group-size 0` / `-5` on `convert` asserting the `must be at least 1` error and no `Writing 0 rows` line.

cholmes:
I've got a refactor of 'write' stuff coming in - I'm going to land it, but it'll overlap with stuff here, so do a careful merge.

cayetanobv:
Status check against #990, which touches the same three files (`common.py`, `parquet_writer.py`, `docs/guide/convert.md`) and conflicts with this branch.

**#990 supersedes most of this PR.** It resolves `row_group_rows` through `resolve_row_group_rows` at the write funnel for every path, rejects a value below 1 with the sort commands' `must be at least 1` error, and prints the rounding note from that one place. That covers both fixes the earlier review asked for here: the note repeating once per partition file, and the nonsense `Writing 0 rows` line for a sub-1 request.

**One gap #990 leaves open, and it is exactly #986.** Its resolver passes any request of one vector or less through untouched:

```python
if row_group_rows <= WRITER_VECTOR_ROWS:
    return row_group_rows
```

so `--row-group-size 249` prints nothing and DuckDB's `COPY` still writes 2,048-row groups. The rationale given is that pyarrow honours a sub-vector request exactly, which is right, but the DuckDB-backed paths (plain fast path, `duckdb-kv`, `disk-rewrite`) do not, and the reworded convert guide in #990 says "the number you ask for is the number that lands". For 249 on a DuckDB write that is not true.

That is the piece of this PR that still earns its place after #990: `note_writer_vector_rounding` at the two COPY sites, which fires precisely when the resolver has passed a sub-vector value through and the writer is about to round it. It says nothing for an aligned value, so it cannot double up with the resolver's own note.

**Plan:** hold this until #990 merges, then rebase it down to the COPY-site note and its tests (`writer_vector_ceiling`, the silent-when-aligned check, the 249 → 2,048 CLI case, plus a partition run asserting the line appears once). The sub-1 rejection and the partition once-per-run fix drop out as already handled. Raising the sub-vector gap and the docs sentence on #990 itself as well.

cayetanobv:
> I've got a refactor of 'write' stuff coming in - I'm going to land it, but it'll overlap with stuff here, so do a careful merge.

added comment below regarding the new PR #990

cayetanobv:
Merge status: **green and mergeable, but holding it** until #990 lands.

Two of the earlier review's findings are still in this branch as-is: the rounding note prints once per partition file (an h3 or quadkey partition with hundreds of cells prints hundreds of identical lines), and a request below 1 prints a `Writing 0 rows` note before DuckDB rejects the value. #990 fixes both at the write funnel and conflicts with this branch in `common.py`, `parquet_writer.py` and `docs/guide/convert.md`. Merging this first would ship the two-defect note and hand the conflict to #990.

Plan, as in the comment above: once #990 merges, rebase this down to `note_writer_vector_rounding` at the two DuckDB `COPY` sites plus its tests. That is the one piece #990 leaves uncovered — its resolver passes a request of one vector or less through untouched, so `--row-group-size 249` still writes 2,048-row groups without a word, which is the command from #986.

PR #988, opened 2026-09-11
fix(convert): linearize curved geometry under --skip-hilbert instead of failing

### LINKED ISSUE #985: convert: --skip-hilbert fails on curved geometry that the default path linearizes
## Bug Description

`gpio convert geoparquet` with `--skip-hilbert` fails on a source that holds curved geometry, unless the source is a local `.gpkg`. The default path converts the same source. It linearizes the curves during the bounds pass. `--skip-hilbert` removes that pass, and no other step linearizes.

The Python API already handles this case. `_read_spatial_to_arrow` catches the DuckDB error `Unsupported geometry type in WKB` and reads the file again through the linearized path. The CLI query path has the same fallback only inside the bounds pass. The docs record the gap in `docs/guide/convert.md`, and `tests/test_linearize.py::TestCliCurveParityWithoutPrescan` pins the failure.

## Steps to Reproduce

Any FileGDB with a CurvePolygon shows it. This one comes from the test fixture:

```
ogr2ogr -f OpenFileGDB curved.gdb tests/data/curved_geometry_test.gpkg
gpio convert geoparquet curved.gdb out.parquet                 # converts
gpio convert geoparquet curved.gdb out.parquet --skip-hilbert  # fails
```

The CAL FIRE historic fire perimeters FileGDB (23,334 rows, https://data.ca.gov/) behaves the same way.

## Expected Behavior

`--skip-hilbert` converts the file. When the write hits curved geometry, gpio linearizes the source and writes again, and it says so. `--no-linearize-curves` keeps the current error.

## Actual Behavior

```
Error: The source contains non-linear geometries (curved or surface WKB types), which DuckDB cannot parse and GeoParquet cannot represent. GeoParquet cannot represent curved geometries; linearize the source first, e.g. `ogr2ogr -nlt CONVERT_TO_LINEAR out.gpkg in.gpkg`. Original error: Invalid Input Error: Unsupported geometry type in WKB
```

## Environment

- gpio 1.5.0, `main` at 82146f0
- GDAL 3.8.4 for `ogr2ogr`
- Linux

## Would you like to submit a PR?

Yes. The fix mirrors the API fallback. When the write fails with `Unsupported geometry type in WKB`, `linearize_curves` is on, and the source is not Parquet, gpio registers the linearized view, builds the query again and writes again. The test that pins the failure changes to expect success, and a new test keeps the error for `--no-linearize-curves`.

cayetanobv:
## PR Title
fix(convert): linearize curved geometry under --skip-hilbert instead of failing

## Description
`gpio convert geoparquet --skip-hilbert` failed on a FileGDB, or any source without a pre-scan, that holds curved geometry. The default path converts the same file, because the bounds pass parses every geometry and linearizes on the spot. `--skip-hilbert` removes that pass, so nothing parsed the curves before the write and the conversion stopped with the `CONVERT_TO_LINEAR` message. The Python API already linearized the same file. `convert` now catches that error on the write, linearizes the source and writes again, and says so. `--no-linearize-curves` keeps the error.

## Technical Details
- `convert_to_geoparquet` wraps the conversion in one retry. On `Unsupported geometry type in WKB`, with `linearize_curves` on and a non-Parquet, non-CSV source, it warns and runs the conversion again with `force_linearize=True`. A second failure raises as before. Nothing at the output path is removed: the first attempt dies before its `COPY` starts and `COPY` overwrites the destination itself, so a file already there is either replaced by the second attempt or, if that fails too, left as it was. The predicate is `_is_linearizable_curve_error`, shared with the bounds pass and the Arrow read.
- `_convert_spatial_path` takes `force_linearize` and registers the linearized view without a pre-scan when it is set. Nothing else in the query path changes.
- Cost: one extra read of the source, only when it holds curves. A linear source pays nothing.
- `tests/test_linearize.py::TestCliCurveParityWithoutPrescan`: the test that pinned the failure now expects success and the note. New tests: a real FileGDB converts; a successful retry overwrites a pre-existing file at the output path; a failed retry (a PolyhedralSurface, built in-test by flipping the fixture's WKB type code) leaves pre-existing bytes there untouched; `--no-linearize-curves` keeps the actionable error and the same bytes.
- `tests/data/curved_geometry_test.gdb`: 124 KB, one CurvePolygon, made with `ogr2ogr -f OpenFileGDB` from `curved_geometry_test.gpkg`. The recipe is `generate_curved_gdb()` in `tests/data/generate_test_fixtures.py`; the optional `timestamps` file is left out because the end-of-file hook rewrites it.
- `docs/guide/convert.md`: the paragraph that described the limitation now describes the retry.

## Breaking Changes
**Does this PR introduce breaking changes?** No. A command that failed now succeeds. The opt-out flag keeps the old behaviour.

## Related Issue(s)
- Closes #985

## Verification
The CAL FIRE historic fire perimeters FileGDB, 23,334 MultiPolygons, 91 of them with arcs (https://data.ca.gov/dataset/california-fire-perimeters-all). On `main` this command fails with the `CONVERT_TO_LINEAR` message. On this branch:

```
$ gpio convert geoparquet California_Fire_Perimeters.gdb out.parquet --skip-hilbert
Converting California_Fire_Perimeters.gdb...
Geometry DuckDB cannot read directly; linearizing the source and converting again
Linearized 91 curved geometries (226 arcs stroked at <= 4.0 degrees per segment); GeoParquet cannot represent curves.
Repaired 408 invalid geometries
Done in 32.3s
Output: out.parquet (86.25 MB)
✓ Output passes GeoParquet validation (metadata checks)
rows 23334 row groups 1
```

The small fixture, on this branch, gives an area of 3.139 for the unit circle CurvePolygon.

Tests:

```
$ uv run pytest -q tests/test_linearize.py tests/test_curved_geometry.py
67 passed
$ uv run pytest -q tests/test_convert_layer.py tests/test_convert_validation_message.py
passed
```

`pre-commit run --files ...` passes on every changed file.

## Checklist
- [x] Integration tests for changes that cross module/format/service boundaries (CLI on a FileGDB)
- [x] All tests pass (`uv run pytest`) for the touched suites; the full suite runs in CI
- [x] At least one adversarial review (AI counts, rubber stamps don't) — @cholmes's review above; every point addressed in 0c2ecc5

## Summary by CodeRabbit

* **New Features**
  * The `convert` command now automatically linearizes supported curved geometries when needed and retries conversion.
  * Curved sources, including FileGDB and remote GeoPackage data, can be converted with `--skip-hilbert`.

* **Bug Fixes**
  * Conversion now preserves an existing output when a retry fails.
  * `--no-linearize-curves` continues to report an actionable error when curved geometries cannot be processed.


geoparquet_io/core/convert.py
```diff
@@ -2040,120 +2053,146 @@ def convert_to_geoparquet(
...
-                max_angle_deg=max_angle_deg,
-            )
+        # Curved geometry the pre-scan cannot see (FileGDB, a GeoPackage on S3)
+        # surfaces as a DuckDB error the first time something parses it. With
+        # Hilbert on, that is the bounds pass and it linearizes on the spot.
+        # With --skip-hilbert nothing parses before the write, so the write is
+        # where the curves show up: linearize the source and write again (#985).
+        # This costs one extra read of the source, and only when it holds curves.
+        force_linearize = False
+        while True:
```
cayetanobv: `convert_to_geoparquet` goes from radon **D (26)** on `main` to **E (33)** here. xenon's `--max-absolute=E` still passes, but this is the last grade it allows — the next branch anyone adds to this function fails the pre-push hook. And the diff is ~130 lines of the same code moved two levels right, which is what the reviewer has to read through to find the six lines that changed.

Suggest extracting the loop body into a helper and leaving only the retry here, e.g.

```python
force_linearize = False
while True:
    try:
        has_geometry = _convert_and_write(con, ..., force_linearize=force_linearize)
        break
    except Exception as e:
        if not _should_retry_linearized(e, force_linearize, is_parquet, is_csv, linearize_curves):
            raise
        warn(...)
        _discard_partial_output(output_file)
        force_linearize = True
```

The body already reads as one unit (read → no-geometry handling → repair → kv/covering → write), so the cut is clean, and the existing code drops back to its old indentation.

geoparquet_io/core/convert.py
```diff
@@ -2040,120 +2053,146 @@ def convert_to_geoparquet(
...
+                if (
+                    force_linearize
+                    or is_parquet
+                    or is_csv
+                    or not linearize_curves
+                    or "Unsupported geometry type in WKB" not in str(e)
+                ):
+                    raise
+                warn(
+                    "Curved geometries detected while writing; linearizing the "
```
cayetanobv: "while writing" is not where the first attempt dies in the default path. `repair_query_geometry` runs `_layered_invalid_count_sql` over the `ST_Read` query before the write, and that count is what parses the curves and raises. Your own verification log shows it: `Repaired 408 invalid geometries` prints once, *after* this note, so the first pass never reached the write. Something like `Curved geometries detected; linearizing the source and converting again` is true on both paths (repair on or off).

geoparquet_io/core/convert.py
```diff
@@ -2040,120 +2053,146 @@ def convert_to_geoparquet(
...
+                    memory_limit=memory_limit,
+                )
+                break
+            except Exception as e:
+                if (
+                    force_linearize
+                    or is_parquet
+                    or is_csv
+                    or not linearize_curves
+                    or "Unsupported geometry type in WKB" not in str(e)
```
cayetanobv: Fourth copy of this predicate: `_bounds_with_curve_fallback` (l.1190), `_read_spatial_to_arrow` (l.1665) and the outer handler (l.2221) all test `is_parquet or not linearize_curves or "Unsupported geometry type in WKB" not in str(e)`. A small `_is_linearizable_curve_error(e, *, is_parquet, linearize_curves)` would keep them in step if DuckDB ever changes the message. Optional, but it pairs naturally with the extraction above.

geoparquet_io/core/convert.py
```diff
@@ -1936,6 +1941,14 @@ def _report_conversion_results(output_file: str, start_time: float, is_geo: bool
         )
 
 
+def _discard_partial_output(output_file):
+    """Remove a local output a failed write may have left behind before a retry."""
+    from geoparquet_io.core.remote import is_remote_url
```
cayetanobv: `is_remote_url` is already imported at the top of this module (l.51, used at 1916, 1933, 2202). The local import can go.

tests/test_linearize.py
```diff
@@ -630,9 +631,17 @@ def test_cli_linearizes_source_the_prescan_cannot_see(self, tmp_path):
         area = con.execute(f"SELECT ST_Area(geom) FROM '{out}'").fetchone()[0]
         assert area == pytest.approx(math.pi, rel=0.005)
 
-    def test_cli_error_names_the_remedy_when_it_cannot_linearize(self, tmp_path):
-        """--skip-hilbert skips the pass that linearizes, so the conversion still
-        fails — but with the actionable message, not DuckDB's raw one."""
+    def _area(self, out):
```
cayetanobv: `test_cli_linearizes_source_the_prescan_cannot_see`, five lines above, still inlines the three lines this helper was added to replace. Use `self._area(out)` there too.

cayetanobv (commented):
**Recommendation: approve after pulling the retry body out into a helper**; the rest are nits. The behaviour change is right: the pre-scan cannot see a FileGDB, the bounds pass was the only other place curves got parsed, and `--skip-hilbert` removed it, so catching the write is the remaining hook. The predicate is conservative (parquet and CSV never retry, `--no-linearize-curves` keeps the error, a second failure raises), and the outer handler still turns that second failure into the `CONVERT_TO_LINEAR` message, which the new test pins.

What I checked: the three existing curve fallbacks (`_bounds_with_curve_fallback`, `_read_spatial_to_arrow`, and the outer `except`) to confirm this is the same predicate in a fourth place; `_register_linearized_view` (it `DROP TABLE IF EXISTS` first, so a retry cannot collide with anything the first attempt left in `con`); `is_remote_url` is already a module import; xenon on this branch passes (`--max-absolute=E`); and `tests/test_linearize.py` locally on this branch (all pass, including the new FileGDB test).

The fixture: 19 binary files for a 124 KB `.gdb` is fine. If dropping `timestamps` ever bites, the alternative is an `exclude: tests/data/.*\.gdb/` on the end-of-file-fixer rather than leaving the file out.

cholmes:
Adversarial review from a Claude Code agent, run at the maintainer's request. Everything below I reproduced myself before posting. **Not merging** — this is still a draft, and it's your branch.

The fix itself is right: I diffed the two function bodies with whitespace normalised and the only semantic additions are the `_convert_once(force_linearize)` wrapper, the `nonlocal`, the pass-through and the 15-line retry. Everything else is byte-identical, re-indented. The retry is properly bounded (the second call sits outside the `try`, so a second failure propagates — no recursion, no loop), it doesn't double-linearize under `--skip-hilbert=False`, the FileGDB fixture genuinely holds a `MULTISURFACE(CURVEPOLYGON(COMPOUNDCURVE(CIRCULARSTRING …)))`, and the tests are real regression tests that fail on `main`.

## One blocker: `_discard_partial_output` deletes a file the run never wrote

`core/convert.py:1944-1949`, called at `:2195`.

The docstring says it removes "a local output a failed write may have left behind". The first attempt doesn't fail during the write — it fails in the writer's **pre-write** geo-stats pass:

```
File ".../core/convert.py", line 2159, in _convert_once
    write_parquet_with_metadata(
File ".../core/write_strategies/duckdb_kv.py", line 549, in _compute_missing_metadata
    bbox, geometry_types = compute_geo_stats_via_sql(
_duckdb.InvalidInputException: Invalid Input Error: Unsupported geometry type in WKB
```

`--verbose` confirms the ordering: `Computing bbox/geometry types via SQL...` fires; `Writing via DuckDB COPY TO...` never does on attempt 1. So there is no partial output. The only thing at that path is whatever the user already had — and it gets removed.

**Reproduced** with a `POLYHEDRALSURFACE` GPKG (the surface family `docs/guide/convert.md` says will still error), with a file the user cares about at the output path:

```
$ printf 'PRECIOUS USER DATA' > keep.parquet      # 18 bytes
$ gpio convert geoparquet surf.gpkg keep.parquet --skip-hilbert
Converting surf.gpkg...
Curved geometries detected while writing; linearizing the source and writing again
Error: surf.gpkg contains non-linear geometries (POLYHEDRALSURFACE) ...
$ ls keep.parquet
ls: keep.parquet: No such file or directory
```

Same input on `origin/main`: the command fails with the same actionable error and **the file is untouched**. So it is a regression, not pre-existing.

Two things make it broader than it first looks:

- **It is not limited to `--skip-hilbert`.** The same command without the flag destroys the file too — the retry fires on any source holding geometry DuckDB cannot parse.
- **It fires on inputs with no curves at all.** A GPKG with a bogus WKB type code produces the same DuckDB string, so the retry runs, prints "Curved geometries detected", deletes the file, and fails.

## The line is unnecessary — I checked

Stubbed `_discard_partial_output` to a no-op and ran the happy path against the committed FileGDB fixture, with a pre-existing file at the output path:

```
Linearized 1 curved geometry (2 arcs stroked at <= 4.0 degrees per segment)
Done in 0.6s
  -> valid parquet, rows=1
```

DuckDB's `COPY TO` overwrites the destination itself. The retry succeeds and the output is correct with nothing deleted.

**Suggested fix: delete `_discard_partial_output` and its call.** The repo already documents the invariant, in `core/check_fixes.py:398-402`: *"The destination is never unlinked or truncated first, so a failure here leaves the original file intact and the rewrite still sitting in the temp file (#959)."* If a partial output ever does become possible, `_staging_path_beside()` + `_move_temp_output_into_place()` is the established pattern.

Worth adding alongside: a regression test that puts bytes at the output path, runs the surface case, and asserts non-zero exit **and** the bytes unchanged. `diff-cover` flags exactly one uncovered line in the diff — `convert.py:1949`, the `os.remove` — because no test puts a file there first. The `--no-linearize-curves` test's `assert not out.exists()` is vacuous for the same reason.

## Smaller things

- **S2** — if the output path is an existing directory, `os.remove` raises inside the `except`, and the outer `except OSError` swallows the actionable message: the user gets `File system error: [Errno 1] Operation not permitted` instead of the `ogr2ogr -nlt CONVERT_TO_LINEAR` remedy. Goes away with the fix above.
- **S2** — the "Curved geometries detected while writing" warning prints before anything confirms curves, so for surface and corrupt-WKB inputs gpio asserts curves and then contradicts itself in its own error. It also costs a second full read of the source before failing — on a multi-GB FileGDB that is a long wait to reach the error `main` gave immediately.
- **S3** — the fixture is reproducible (`ogr2ogr -f OpenFileGDB repro.gdb tests/data/curved_geometry_test.gpkg` gives the identical file list) but the recipe and the deliberate `timestamps` omission live only in the PR body, which a squash merge discards. Per #958, a committed binary fixture wants a committed recipe — four lines in `generate_test_fixtures.py` would do it. (The hook tension you mentioned isn't real: `pre-commit run --all-files` passed and modified nothing; the text hooks skip `.gdbtable` binaries.)
- **S4** — `nonlocal geoparquet_version, effective_crs` lets attempt 1 mutate state attempt 2 inherits. Not reachable to harm today since detection is deterministic, but it is state leaking across a retry.

## Gates

Fast suite **7292 passed**, 2 failures both on the known cold-run flake list (2 passed serially) · `diff-cover` 97%, the one miss being the `os.remove` · meta 205 · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit clean and modified nothing · docs lane clean after serial re-run.

Happy to push the one-line deletion and the regression test if you'd like — say the word and I will, or leave it to you.

cayetanobv:
@cholmes thanks for the review — every point is addressed in 0c2ecc5, and the PR description is updated to match.

- **Blocker, `_discard_partial_output`**: deleted with its call. Reproduced first: the same PolyhedralSurface case took a pre-existing file at the output path on the default path as well as under `--skip-hilbert`. New regression test seeds bytes there, runs a surface input (built in-test by flipping the curved fixture's WKB type code to 15, so no committed binary), asserts non-zero exit and the bytes unchanged. It fails on 4714a1e and passes now. A second test checks a successful retry overwrites a pre-existing file, and the `--no-linearize-curves` test now checks real bytes instead of a path nothing wrote to.
- **S2, directory at the output path**: gone with the removal.
- **S2, warning claims curves before anything confirms them**: reworded to "Geometry DuckDB cannot read directly; linearizing the source and converting again". The extra read before a surface input fails remains; avoiding it needs a surface detector for sources without a pre-scan, which I left out of this PR.
- **S3, fixture recipe**: `generate_curved_gdb()` in `tests/data/generate_test_fixtures.py`, with the `timestamps` omission explained. A rerun gives the same file set; two system tables differ in bytes because they record the creation time, and the docstring says so.
- **S4, `nonlocal`**: each attempt now decides its own output version and CRS from the outer values.

Also from the earlier review: one `_is_linearizable_curve_error` predicate shared by the bounds pass, the Arrow read and the retry, and `convert_to_geoparquet` is at radon C (16), down from E (33) at the first commit.

CI is green on the full matrix. Ready to merge from my side.

Issue #600, opened 2026-07-22
auto version mode: native-geo-only inputs downgrade to 1.1 in sort/extract/partition


cholmes:
PR #594 added `resolve_geoparquet_version_from_file` (core/common.py) so that auto version mode (no `--geoparquet-version`) preserves a parquet input's GeoParquet version and upgrades native-geo-only (parquet-geo-only) inputs to 2.0, matching the shared `--geoparquet-version` help text. It was wired into `convert` (core/convert.py), and reproject now uses it too (todo 040).

The remaining commands still fall back through `write_parquet_with_metadata` -> `extract_version_from_metadata(original_metadata)`, which returns `None` for native-geo-only inputs (no `geo` key), so the write defaults to 1.1.0 WKB — silently stripping native GEOMETRY/GEOGRAPHY logical types.

Affected call sites (all pass `geoparquet_version=None` straight through):

- `geoparquet_io/core/hilbert_order.py:480` and `:523` (`gpio sort hilbert`)
- `geoparquet_io/core/sort_by_column.py:203` (`gpio sort column`)
- `geoparquet_io/core/sort_quadkey.py:250` (`gpio sort quadkey`)
- `geoparquet_io/core/extract.py:1022` (`gpio extract geoparquet`)
- `geoparquet_io/core/check_fixes.py` (three `write_parquet_with_metadata` calls used by `gpio check --fix`)
- partition writes via `geoparquet_io/core/partition/staging.py:225`

Repro:

```bash
duckdb -c "COPY (SELECT 1 id, ST_Point(0,0) geometry) TO 'pgo.parquet' (FORMAT PARQUET, GEOPARQUET_VERSION 'NONE')"
gpio sort hilbert pgo.parquet out.parquet   # out.parquet is 1.1.0 WKB, native type stripped
```

Suggested fix: resolve auto mode via `resolve_geoparquet_version_from_file(input)` in each command (as convert/reproject do), or push the resolution down into `write_parquet_with_metadata` now that it receives `input_file`.

Refs: PR #594, todo 040.

cholmes:
[...]

PR #990, opened 2026-09-11
fix(write)!: one row-group default, one auto-version rule and one geo-key rule for every write path

### LINKED ISSUE #773: parquet-geo-only leaks the geo key on three write strategies when the geometry column is absent
#753 fixed `write_geoparquet_table` so that an explicit
`geoparquet_version="parquet-geo-only"` drops the input's carried `geo` key even
when the geometry column has been projected away (#701). It did **not** fix the
documented Python API path, which does not go through `write_geoparquet_table`.

`Table.write()` dispatches to `strategy.write_from_table()`, and three of the
four strategies still write the carried `geo` key in the same situation.

## Reproducer

```python
import json
import pyarrow as pa
import pyarrow.parquet as pq
from geoparquet_io.api.table import Table

geo = {
    "version": "1.1.0",
    "primary_column": "geometry",
    "columns": {"geometry": {"encoding": "WKB", "geometry_types": ["Point"]}},
}

def attributes_only_with_geo():
    # A table whose geometry column has been projected away, but which still
    # carries the input's `geo` key plus an unrelated sidecar payload.
    return pa.table({"id": pa.array([1, 2])}).replace_schema_metadata(
        {b"geo": json.dumps(geo).encode(), b"fiboa": b'{"schemas":["example"]}'}
    )

for strategy in ["in-memory", "streaming", "disk-rewrite", "duckdb-kv"]:
    out = f"/tmp/{strategy}.parquet"
    Table(attributes_only_with_geo()).write(
        out, geoparquet_version="parquet-geo-only", write_strategy=strategy
    )
    keys = sorted(
        k.decode() for k in (pq.ParquetFile(out).metadata.metadata or {})
        if k != b"ARROW:schema"
    )
    print(f"{strategy:14s} -> keys: {keys}")
```

## Observed

```
in-memory      -> keys: ['fiboa', 'geo']   <- leaked
streaming      -> keys: ['fiboa', 'geo']   <- leaked
disk-rewrite   -> keys: ['fiboa', 'geo']   <- leaked
duckdb-kv      -> keys: []
```

The leaked `geo` key declares `"primary_column": "geometry"` for a file whose
only column is `id`, and declares a GeoParquet version for output that was
explicitly asked to carry no GeoParquet metadata at all — the same defect #701
described, on a different code path.

`duckdb-kv` (the `Table.write()` default) is correct about `geo`, but note that
it also drops the unrelated `fiboa` sidecar key, which the other three preserve.
That divergence may be worth a separate look; the fix here should land on
`['fiboa']` for all four.

## Cause

Each strategy guards its metadata work on the geometry column being present and
returns early otherwise, writing the table's schema metadata through verbatim.
This is the same shape of guard #753 widened in `common.py`:

- `geoparquet_io/core/write_strategies/arrow_memory.py:168` — `has_geometry = geometry_column in table.column_names`, and `_apply_geoparquet_metadata` is only called when true.
- `geoparquet_io/core/write_strategies/arrow_streaming.py:685` — `if geometry_column is None:` delegates to `_write_plain_parquet_from_table` (`arrow_streaming.py:956`), which writes `table.schema` unchanged.
- `geoparquet_io/core/write_strategies/disk_rewrite.py:189` (`write_from_table`), guard at `disk_rewrite.py:211` — same delegation to `_write_plain_parquet_from_table` (`disk_rewrite.py:258`).

## Suggested fix

Route the geometry-absent + explicit-`parquet-geo-only` case through
`_strip_geo_metadata_key` (`geoparquet_io/core/common.py`), which #753 already
uses and which drops `geo`, `ARROW:schema` and `pandas` while preserving
unrelated sidecar keys. Restricting it to the *explicit* request keeps auto mode
(`geoparquet_version=None`) unchanged, as #753 did.

This was deliberately kept out of #753: the project's convention is that a
shared/cross-cutting bug gets its own PR rather than being bolted onto a feature
branch.

Related: #701, #753, #687.

### LINKED ISSUE #971: fix(api): Table.write(row_group_rows=50000) writes 51,200-row groups while gpio sort --row-group-size 50000 writes 49,152

### LINKED ISSUE #981: fix(convert): convert geoparquet's docstring advertises row groups it does not write
## What

`core/convert.py:1968` lists the best practices the command applies:

```
Applies best practices:
- ZSTD compression
- 100k row groups
- Bbox column with metadata
- Hilbert spatial ordering (unless --skip-hilbert)
```

This is wrong twice.

### 1. Nothing asks for 100,000

`row_group_rows=None` (`convert.py:1946`) is passed straight through (`:2145`), so the writer picks. Measured on a 300,000-row input:

```
$ gpio convert geoparquet in.parquet out.parquet
groups 3  sizes [122880, 122880, 54240]
```

That is DuckDB's COPY default, not 100k.

### 2. Even 100k would be the self-contradiction #972 is about

A 100,000-row request lands at 100,352 after the writer's 2,048-row vector rounding — inside `GENERAL_ROW_COUNT_RANGE` but outside `SPATIAL_ROW_COUNT_RANGE`, the band `check optimization` scores. The file the command actually writes fails it today:

```
$ gpio check optimization out.parquet
[fail] Row Group Size: Average 100,000 rows per group (optimal 10k-50k for spatial queries)
```

So a command whose docstring claims to apply "best practices" *including Hilbert spatial ordering* produces a file gpio's own spatial check marks `[fail]` and advises re-partitioning — the same shape as #972, one command over. Sorting exists to make spatial predicates prune row groups; groups too large to prune undo what the sort just bought.

## Decision needed

Either correct the docstring to say the writer picks, or make `convert geoparquet` write `DEFAULT_SORT_ROW_GROUP_ROWS` (49,152) the way `gpio sort` (#967) and `check --fix` (#974) now do. The second is a behaviour change to a very widely used command and would need its own `!` PR title; the first is a one-line docs fix.

Refs #972, #961, #775, PR #974.

cholmes:
## What changed

Row-group size, auto-mode GeoParquet version and the output's `geo` key were each
decided independently in every write path. The paths disagreed, and the
disagreements shipped one bug at a time:

| issue | site | wrote |
|---|---|---|
| #967 (merged) | `gpio sort` | 50,000 → **49,152** fixed |
| #974 (merged) | `check --fix` | 100,000 → **49,152** fixed |
| #961 (merged) | `pmtiles` | **49,152** fixed |
| **#971** | `Table.write(row_group_rows=50000)` | **51,200** — API and CLI disagreed |
| **#981** | `convert geoparquet` | **122,880** (DuckDB's default; the docstring claimed 100k, nothing set either) |

`core/parquet_writer.py` becomes **the write facade** and owns exactly three
decisions, because each is a fact about the *output file* that no individual
write path is entitled to answer for itself:

1. **`resolve_row_group_rows`** — one default and one alignment rule.
   `DEFAULT_SORT_ROW_GROUP_ROWS` is renamed `DEFAULT_ROW_GROUP_ROWS` (still
   derived, never typed: `align_to_writer_vector(SPATIAL_BAND_TOP_ROWS)`) because
   it is no longer sort-only. Applied at three funnels:
   `common.write_parquet_with_metadata` (21 call sites + partition),
   `common.write_geoparquet_table` (arcgis, carto, wfs, bigquery, aggregate,
   overview) and `api.table.Table._write_geoparquet`.
2. **`resolve_output_geoparquet_version`** — auto mode consults the input *file*
   the way `convert` and `reproject` always did, instead of the carried `geo`
   key, which a native-geo-only input does not have.
3. **`apply_output_kv_metadata`** — the version decides whether a `geo` key is
   written at all, applied before the non-geo early return that used to skip it.

### The line the facade does not cross

Compression validation (`validate_compression_settings`), the *contents* of the
`geo` block (`build_geo_metadata`), CRS resolution (`apply_output_crs`), bbox
coverings and geometry-type computation each already have one owner. Pulling
them in would turn a facade into a rewrite. `core/common.py` is **not** split —
that is Tier 3.4.

`WriteOptions` (#761) is deliberately **not** in this PR: it is a public-API
design change (a new exported class, its CLI construction, the
`--write-memory`-is-duckdb-kv-only constraint it has to express) orthogonal to
this defect class, and folding it in would double the diff and force a second,
unrelated `!` decision into the same review.

## The eight contracts

`tests/test_write_facade_contract.py` was written in #880 with 8
`xfail(strict=True)` markers, to be turned green by this work. All eight now
pass. **Every marker is deleted and every assertion is unchanged** — the file's
6 controls and current-behaviour pins are untouched and still pass.

| contract | how |
|---|---|
| `parquet_geo_only_drops_geo_and_keeps_sidecar[in-memory]` | `apply_output_kv_metadata` before the write |
| `…[streaming]` | same, above the `geometry_column is None` early return |
| `…[disk-rewrite]` | same |
| `auto_mode_preserves_native_geo[sort-hilbert]` | `resolve_output_geoparquet_version` in `write_parquet_with_metadata`; these paths already pass `input_file=` |
| `…[sort-column]` | same |
| `…[sort-quadkey]` | same |
| `…[extract-geoparquet]` | same |
| `partition_auto_mode_preserves_native_geo` | the partition **driver** resolves once from the real input. A partition write sees only the staging file, which is always written `GEOPARQUET_VERSION 'NONE'`, so resolving from what it reads would call every input native-geo-only. |

## Behaviour changes (why the `!`)

1. **`convert geoparquet` row groups: 122,880 → 49,152.** Widely used command.
   The file it wrote failed `gpio check optimization`'s row-group factor; it now
   passes (measured below).
2. **`Table.write(row_group_rows=50000)`: 51,200 → 49,152** — matches
   `gpio sort --row-group-size 50000` (#971).
3. **`Table.write()` with no row-group option: 122,880 → 49,152.**
4. **Every other write path's default → 49,152**: `extract`, `add *`,
   `partition`, `reproject`, `process`, and the Arrow-side converters
   (`arcgis`, `carto`, `wfs`, `bigquery`), which fell through to
   `ParquetWriteSettings`' own 100,000. `ParquetWriteSettings.DEFAULT_ROW_GROUP_ROWS`
   (100,000) is removed in favour of the module constant.
5. **An explicit `--row-group-size` above one writer vector is now snapped on
   every path**, not just `sort` (`convert --row-group-size 50000` → 49,152).
6. **Auto version mode on a native-geo-only input now writes native 2.0** for
   `sort hilbert`, `sort column`, `sort str`, `sort quadkey`, `extract
   geoparquet`, `partition string` and `partition admin`, instead of rewriting
   it as 1.1 WKB and stripping the Parquet GEOMETRY logical type (#600).

   It does **not** reach `partition quadkey`/`h3`/`s2`/`a5`/`kdtree`, and that
   is deliberate. All six partition drivers share `partition_by_column`, but the
   five index-adding ones hand it the scratch file their `add <index>` step
   wrote rather than the user's input, so the facade resolves auto mode from a
   file that is already a 1.1 WKB rewrite. Measured on a native-geo-only input,
   before vs after:

   | driver | before | after |
   |---|---|---|
   | `partition string` | v1.1.0, no logical type | **v2.0.0, native logical type** |
   | `partition admin` | v1.1.0, no logical type | **v2.0.0, native logical type** |
   | `partition quadkey` / `h3` / `s2` / `a5` / `kdtree` | v1.1.0 | **byte-identical — unchanged** |

   `partition admin` needed a second fix to get there: it built its
   `PartitionWriteOptions` with the raw, unresolved `geoparquet_version` while
   `partition/staging.py` — added in this same diff — documented the field as
   already resolved by the driver. It reads the user's file directly (the
   enriched table is `a.*` plus admin columns, with only the join predicate
   reprojected), so resolving there is safe, and measured against a local
   Overture cache it now writes v2.0.0 native and keeps EPSG:5070 through the
   join. An `ast`-based test in `test_write_facade_version_owner.py` now checks
   every `PartitionWriteOptions` construction site obeys the invariant, since
   an invariant with an exempt site is a comment.

   Pointing them at the user's file would fix the version and import a worse
   defect, because the *data* would still come from a scratch file that lost the
   CRS: `gpio sort quadkey` does exactly that today, and on
   `tests/data/fields_pgo_5070_snappy.parquet` it now writes a native 2.0 file
   whose inline PROJJSON says `OGC:CRS84` over EPSG:5070 data — a wrong CRS
   *asserted* where before it was merely absent, which `gpio check spec` then
   blesses with `✓ valid inline PROJJSON CRS`. (Net reader-visible CRS is the
   same wrong answer either way, since absent already means CRS84, so this is
   not new corruption — but it is not something to spread to five more
   commands.) The thing to fix is the scratch write. Both gaps are pinned
   `xfail(strict=True)` in `tests/test_write_facade_version_owner.py`, so
   whoever fixes it has to come and delete the markers.
7. **`parquet-geo-only` no longer leaks a carried `geo` key** through the
   in-memory, streaming and disk-rewrite strategies (#773).
8. **A request of one writer vector (2,048 rows) or less is passed through
   instead of snapped.** This one needed measuring rather than assuming, on a
   120,000-row input:

   | request | DuckDB `COPY` | `pq.write_table` |
   |---|---|---|
   | 1 / 10 / 100 / 1,000 / 2,047 | 2,048 | honoured exactly |
   | 3,000 | 4,096 | 3,000 |
   | 50,000 | 51,200 | 50,000 |

   DuckDB quantises a sub-vector request to 2,048 whatever gpio does, while
   pyarrow obeys it. Rounding it first would therefore be gpio changing the
   output on the only writer that could have obeyed the user — the inverse of
   what the alignment is for. `align_to_writer_vector` keeps its
   ceiling-everywhere rule and its tests untouched (it answers "what would the
   DuckDB writer do"); the *facade* decides when to ask.

   For `gpio sort str` this changes row **order**, not just layout: the STR tile
   size *is* the row-group request, so `--row-group-size 20` now packs 20-row
   tiles where the old rounding made them 2,048-row tiles.
9. **`Table.write(row_group_rows=0)` now raises `InvalidParameterError`** where
   it previously meant "let the writer default". An improvement — 0 was never a
   row-group size — but an API-breaking one, so it is named here rather than
   left to be discovered. The rejection names `row_group_rows`, not the CLI's
   `--row-group-size`, because a Python caller never typed that flag.
10. **`--row-group-size --help` now states the rule on every command.** It used
    to say "if neither is given the writer's own default applies (122,880 rows
    for DuckDB-backed writes)" on all 22 non-sort commands, which this PR makes
    false. `tests/data/cli_surface.json` structurally cannot catch that — it
    records Click's *declared* default, which stays `None` — so
    `tests/test_write_facade_row_groups.py` now checks the help string of every
    command that offers the option.
11. **`gpio sort hilbert` / `sort str` stop advising `--geoparquet-version 2.0`
    on native-geo-only inputs.** `hilbert_order._resolve_output_version` decides
    what those two commands *say* they are writing, and it kept its own copy of
    the pre-facade rule (`extract_version_from_metadata(metadata) or
    DEFAULT_GEOPARQUET_VERSION`), so on a native-geo-only input it said "1.1"
    and printed "Consider using --geoparquet-version 2.0" over a file it was
    writing as 2.0.0 — #738's bug arriving from the other side, on the exact
    input class this PR exists to fix. It now asks the facade the same question
    the write asks it, with the same two witnesses and the same
    `or DEFAULT_GEOPARQUET_VERSION` fallback. The failed-read guard stays local:
    the facade swallows an unreadable input and returns `None`, which the `or`
    would turn straight back into the guess #738 removed. Second consequence,
    also fixed: `hilbert_order`'s row-group guidance is gated on the effective
    version being 2.0/parquet-geo-only, so it could never fire for these inputs
    either.

## Measured, on a real 300,000-row file

| path | before | after |
|---|---|---|
| `gpio convert geoparquet` | `[122880, 122880, 54240]` | `[49152 × 6, 5088]` |
| `gpio sort hilbert` | `[49152 × 6, 5088]` | unchanged |
| `gpio extract geoparquet` | `[122880, 122880, 54240]` | `[49152 × 6, 5088]` |
| `Table.write(row_group_rows=50000)` | `[51200 × 5, 44000]` | `[49152 × 6, 5088]` |
| `Table.write()` | `[122880, 122880, 54240]` | `[49152 × 6, 5088]` |

```
$ gpio check optimization <convert output>
before: [fail] Row Group Size: Average 100,000 rows per group (optimal 10k-50k for spatial queries)
after:  [pass] Row Group Size: Average  42,857 rows per group (optimal 10k-50k for spatial queries)
```

### File size

More, smaller row groups means more per-group footer and column statistics, so
output files change size. The effect is strongly data-dependent and there is no
single percentage to quote. `gpio convert reproject`, 300,000 rows, `origin/main`
vs this branch:

| fixture | main | branch | Δ |
|---|---|---|---|
| random global points | 5,018,820 | 5,045,217 | **+0.5%** |
| spatially clustered points | 4,351,763 | 4,347,157 | **−0.1%** |
| highly compressible (3 distinct attribute values, 30 distinct points) | 6,836 | 12,133 | **+77.5%** |

Immaterial on realistic data — sometimes smaller, because tighter per-group
statistics compress better than they cost. Tens of percent on near-constant
data, where a few kilobytes of payload cannot absorb seven row-group footers
instead of three. That is the trade the 10,000–50,000 band already asks for, and
`gpio check optimization` was already failing the files gpio wrote without it.

## Verification

- `uv run pytest -n auto -m "not slow and not network and not meta"` — **7378
  passed, 76 skipped, 7 xfailed** (0 xpassed; 6 of the 7 xfails are the new
  strict pins in `test_write_facade_version_owner.py`)
- `uv run pytest -m meta` — **205 passed**
- `uv run pytest docs/guide -n 4 -m "not network"` — **236 passed, 224 skipped**
- `uv run pre-commit run --all-files` and `--hook-stage pre-push` — all green
  (ruff, codespell, duckdb-antipatterns, doc-sync, deptry, mypy, vulture, xenon,
  import-linter, menard-check)
- `diff-cover` vs `origin/main` — **100% on 61 changed lines**, 0 missing
- `tests/data/cli_surface.json` — **unchanged**, as it should be: no CLI surface
  moved, only help text

Four tests needed updating, each because it pinned a claim this PR changes on
purpose, not because an assertion was inconvenient:

- `test_str_order.py::…preserves_the_in_memory_ordering` asked
  `align_to_writer_vector(20)`; it now asks `resolve_row_group_rows(20, None)`,
  which is the function the file path actually uses.
- `test_check_parquet_structure.py::…label_each_band` pinned the string
  "gpio sort defaults to 49,152"; the guideline line now says "gpio writes
  49,152 by default", because naming one command would under-claim.
- `test_sort_row_group_default.py::_help_default_rows` compared the help's
  default against `str(rows)`; the help now groups thousands ("49,152"), so it
  compares the string it parsed instead of a reformatting of it.
- `test_cli_api_call_parity_scaffold.py` gained one `KNOWN_PARITY_GAPS` entry
  with its reason. It is **measurement depth, not behaviour** — the same gap the
  case's own notes already concede for `geoparquet_version`: the CLI's patch
  point (`convert_to_geoparquet`) runs *before* the facade call and the table's
  (`WriteStrategyFactory.get_strategy`) runs *after* it, so the harness sees
  `None` on one side and `49152` on the other. The new
  `tests/test_write_facade_row_groups.py` asserts the half the harness cannot
  see: both front ends write the same row-group layout, read back from the
  footer.

`.menard/reviewed.json` gains three entries. `menard-check` passes on a clean
`origin/main` worktree and fails here only because this PR *touches*
`cli/commands/sort.py` (a symbol rename, nothing else), which re-arms drift that
#948/#959/#973/#977 introduced. All three targets — `CLAUDE.md#CLI Command
Groups`, `docs/contributing.md`, `skills/geoparquet.md#Command Reference` — are
generated by `doc_sync.py`, which reports them up to date. The review record
expires the next time those files change.

Closes #971
Closes #981
Closes #773
Refs #600 — fixed for `sort`, `extract`, `partition string` and
`partition admin`; the five index-adding `partition` drivers still resolve from
their scratch file (behaviour 6 above), so the issue stays open until that
scratch write preserves the input's version and CRS. That remainder is **#993**,
which the six `xfail(strict=True)` markers in
`tests/test_write_facade_version_owner.py` point at — they fail loudly the
moment it lands.


cholmes:
## Multi-agent adversarial review — five S2s found and fixed in a642fb1

Taken last and slowly, as asked. Reviewed against a clean extraction of `origin/main`; every claim below is a measured A/B, and I re-verified the load-bearing ones myself.

### What held up

- **The xfail claim is mechanically true.** Extracting every `assert` / `def test` / `def _` line from both revisions of `tests/test_write_facade_contract.py` and diffing: **19 → 19, one change only** (a helper signature). No assertion text moved, no test renamed or dropped. The 6 controls are untouched.
- **No CRS is lost on the geo-key drop.** All four strategies on `fields_gpq2_5070_brotli.parquet` keep the EPSG:5070 PROJJSON in the GEOMETRY logical type; the only delta is `streaming` no longer leaking `kv=['geo']` — exactly #773.
- **The sub-vector measurement reproduces exactly.** DuckDB `COPY` quantises 1/10/100/1,000/2,047 → 2,048 and 3,000 → 4,096, while `pq.write_table` honours them. Rule 8 follows from the measurement.
- **Native 2.0 outputs are readable by DuckDB, pyarrow and geopandas**, keep the logical type and a real CRS. The one `check spec` complaint fires identically on the *source fixture itself*, so it is pre-existing.

`ParquetWriteSettings.DEFAULT_ROW_GROUP_ROWS`' removal is not a public-surface break — it is reachable only by deep import and fails loudly.

### S2 — `--help` stated the rule this PR deleted, contradicting the guide in the same diff

```
before:  "the Parquet writer rounds this up ... the writer's own default applies (122,880 rows)"
actual:   no option -> 49,152, and gpio snaps DOWN
after:   "snapped to a whole 2,048-row writer vector ... (default: 49,152 ...)"
```

False on ~20 commands, while `docs/guide/convert.md` in the same commit said 49,152. That is the #967 shape for the third time. Fixed by deleting the `default_rows` branch entirely and building one help constant from the facade's own constants, with a tree-walking test asserting all 26 commands offering `--row-group-size` name `49,152` and contain neither `122,880` nor "writer's own default".

Worth recording: **`cli_surface.json` could not have caught this.** It records Click's *declared* default (`None`), so an effective-default move is structurally invisible to it.

### S2 — behaviour 6 claimed "partition" but only `partition string` delivered it

| driver | before | after |
|---|---|---|
| `partition string` | 1.1.0, WKB | **2.0.0, native** |
| `partition admin` | 1.1.0, WKB | **2.0.0, native, EPSG:5070** (fixed here) |
| `quadkey` / `h3` / `s2` / `a5` / `kdtree` | 1.1.0 | unchanged — now pinned `xfail(strict=True)` |

The obvious fix was rejected **with evidence**, and I agree with the call: resolving from the user's file in the five index drivers would fix the version and *import `sort quadkey`'s wrong-CRS defect onto five more commands*, because the data still comes from the `add <index>` scratch file. Narrowing the claim and pinning the gap beats shipping a wider wrong answer.

`partition admin` also violated an invariant this PR itself added (`staging.py` documents that `geoparquet_version` arrives resolved, while `admin_hierarchical.py:625` passed it raw). Now resolved, with an `ast`-based test asserting *every* `PartitionWriteOptions` construction resolves — it went red naming exactly that line, then green.

### S2 — a stale second owner of the facade's own decision

`hilbert_order._resolve_output_version` still guessed independently, so on a native-geo-only input it returned `"1.1"` while the facade wrote `"2.0"`:

```
$ gpio sort hilbert fields_pgo_5070_snappy.parquet out.parquet
Hilbert sorting to GeoParquet v1.1 provides no spatial filter pushdown benefit.
Consider using --geoparquet-version 2.0 ...      # the file it just wrote is 2.0.0
```

It advised a flag that changes nothing, on this PR's own headline input class, and suppressed the row-group guidance gated on the 2.0 branch. #738 fixed the mirror image; this re-opened it from the other side. Now routed through the facade — fixing `sort str` too — with the failed-read probe kept local, because the facade returns `None` on an unreadable input and the `or` would restore the guess #738 removed.

### S2 — the contract test certified a gap as green

`sort quadkey` writes `logical crs=OGC:CRS84` over EPSG:5070 data — a wrong CRS *asserted* where it was previously merely absent, which `check spec` then blesses. The test could not see it because `parquet_geo_only_file` is DuckDB-written with `crs=<null>`. A new `tests/test_write_facade_version_owner.py` uses the repo's real `fields_pgo_5070_snappy.parquet`: four direct-read commands assert 2.0 + native + EPSG:5070 and pass; the gap is six `xfail(strict=True)` markers carrying the cause and the exit path. Verified genuinely failing, not decorative.

### S2 — file sizes change, and the body did not say so

The reviewer reported "+7% to +74%". **I could not reproduce that as a general claim and it is not in the body.** My measurements, `convert reproject`, 300,000 rows:

| fixture | main | branch | Δ |
|---|---|---|---|
| random global points | 4,993,067 | 4,988,108 | −0.1% |
| spatially clustered | 4,359,077 | 4,340,084 | −0.4% |
| highly compressible | 5,075 | 7,155 | **+41%** |

The effect is real but strongly data-dependent — more, smaller row groups means more per-group footer overhead, immaterial for typical data and large where it dominates a tiny payload. The body now carries two contrasting fixtures and the mechanism, with no blanket percentage.

### Also corrected

`docs/api/python-api.md` claimed `row_group_rows=10` writes 10-row groups; measured, the default `duckdb-kv` strategy writes **2,048** (10 only lands on the Arrow strategies). `docs/guide/check.md` offered pyarrow's `1,048,576` as an in-band example when `assess_row_count` scores it **`poor`**. Three undeclared behaviour changes are now named: `Table.write(row_group_rows=0)` raising where it meant "writer default", CLI flag names leaking into Python API messages (now parameterised), and `sort str --row-group-size ≤2048` changing row **order**.

### `Closes #600` → `Refs #600`, and why

Five of seven partition drivers still downgrade, so closing #600 would retire an issue with a live remainder. The shared root cause — which neither the review nor I had named — is that the **whole `add *` family** rewrites a native-geo-only input as 1.1 WKB, dropping the logical type and the CRS. That is what produces both this gap and the `sort quadkey` CRS defect, and fixing it changes output for eight more commands, so it is its own `!` change: **#993**. The six xfail markers point at it and fail loudly when it lands.

### Gates

Fast suite **7379 passed, 76 skipped, 7 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100% on 65 lines** · `pytest -m meta` 205 passed · docs lane 236 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push clean (xenon, menard-check included) · `cz check` accepts the title with `!` intact. Baseline was 7310 passed / 1 xfailed; delta is +69 passes and +6 strict xfails.

No S1 anywhere: no security issue, no new data corruption, no file made unreadable.

Issue #700, opened 2026-08-22
`gpio add bbox-metadata` crashes on any file with an `ARROW:schema` footer key (unquoted KV_METADATA keys)


cholmes:
Found as a drive-by while reviewing #681.

`core/add/bbox_metadata.py` builds DuckDB's `KV_METADATA` clause by interpolating existing footer keys **unquoted**:

```python
for key, value in existing_metadata.items():
    if key == "geo":
        continue
    escaped_value = value.replace("'", "''")
    kv_parts.append(f"{key}: '{escaped_value}'")
```

pyarrow writes an `ARROW:schema` footer key by default, so for any pyarrow-written GeoParquet this produces:

```sql
KV_METADATA {ARROW:schema: '...'}
```

which fails with `Parser Error: syntax error at or near ":"`. The command is unusable on a large class of real inputs.

Two fixes, both required:

1. **Quote the key** — `quote_identifier(key)` from `core/duckdb_utils.py`.
2. **Stop hand-rolling the value escape** — `value.replace("'", "''")` should be `_escape_sql_string(value)`, per the CLAUDE.md rule ("Never hand-roll SQL escaping"). Same for the `geo_json` escape a few lines below.

Worth grepping for the same `KV_METADATA {` construction elsewhere — the pattern looks copy-pasted.

Repro: write a GeoParquet with pyarrow (which emits `ARROW:schema`), then run `gpio add bbox-metadata` on it.

cholmes:
Retested on `main` @ `25a3e2d` (v1.4.0) — **no longer reproduces.** Fixed by #756 (`b1468cf`), whose third commit converged `add bbox-metadata` on the shared KV clause builder.

Repro from the issue, a pyarrow-written GeoParquet carrying `ARROW:schema` plus a bbox column:

```
$ gpio add bbox-metadata w700b.parquet
Added bbox covering metadata for column 'bbox'
```

Output verified: `keys: ['ARROW:schema', 'geo']` — the key survives and the covering is written, with no `Parser Error: syntax error at or near ":"`.

[...]

PR #995, opened 2026-09-11
fix(cli): answer a DuckDB-rejected file with an error line instead of a traceback

### LINKED ISSUE #983: fix(cli): four commands answer a DuckDB-rejected input with a traceback, not an error line

cholmes:
Fixes #983.

## What changed

A GeoParquet file whose `geo` block DuckDB's Parquet reader refuses now gets
`Error: <DuckDB's message>` and exit 1 from every command, instead of a raw
Python traceback from most of them.

The block on an input file is arbitrary JSON written by somebody else's tool, so
DuckDB refusing it is news about the *file*, not a gpio bug — and DuckDB's
message ("Geoparquet metadata does not have a columns object") already says
exactly what is wrong. Only the traceback around it was wrong.

## Why this shape

`gpio inspect head` and `gpio convert geoparquet` already answered such a file
cleanly, because `cli/commands/inspect.py` and `core/convert.py` each carry
their own `except` block. Eleven commands never reach one of those. Adding
eleven more would have left the twelfth command, written next month, uncovered.

The root group is already the single funnel every subcommand passes through —
`SpillAwareGroup` was installed there for exactly that reason when DuckDB's
out-of-spill-space message needed a `TMPDIR` line (#752). So the fix widens that
boundary rather than adding a new one:

- `core/duckdb_utils.py` gains `INPUT_FILE_DUCKDB_ERRORS` and
  `is_input_file_duckdb_error()`, beside `spill_space_hint`. That is the whole
  taxonomy — "is this DuckDB error news about the user's file, or a bug in
  something we generated?" — and it lives in `core/` so `api/` can reach the
  same verdict; `api/` may not import `cli/`.
- `cli/exception_handler.py` gains `cli_error_for()`, which decides nothing
  itself. It asks three questions that already have answers in `core/` — spill
  shortage, unpublished community extension, input-caused DuckDB failure — in
  that order, most specific first, then wraps. Everything else returns `None`.
- `SpillAwareGroup` becomes `ErrorBoundaryGroup`, since naming it after one of
  the failures it now handles would be misleading.
- `None` means "not mine": the exception is re-raised untouched, traceback and
  all. The traceback that *is* hidden is logged at DEBUG rather than dropped.

Exit code is unchanged at 1, which is what `inspect head` and
`convert geoparquet` already produced — confirmed by running them, not assumed.

## `--verbose` really does restore the hidden traceback now

It did not, on four of the eleven commands, and the comment in
`cli/decorators.py` promised that it did. Measured on the `{"version":"1.1.0"}`
file, `--verbose` **after** the subcommand, looking at stdout+stderr together:

| Command | before this commit | after |
|---|---|---|
| `check optimization` | traceback | traceback |
| `check all` | traceback | traceback |
| `check compression` | traceback | traceback |
| `check row-group` | traceback | traceback |
| `check spatial` | traceback | traceback |
| `add quadkey` | traceback | traceback |
| `sort str` | traceback | traceback |
| `add bbox` | **nothing** | traceback |
| `sort hilbert` | **nothing** | traceback |
| `extract geoparquet` | **nothing** | traceback |
| `add geometry-metrics` | **nothing** | traceback |

7 of 11, now 11 of 11. Two measurement details account for every reading anyone
got, including "0 of 6":

- **Position matters.** The root group has no `--verbose`; it is a per-command
  flag. `gpio --verbose check optimization file.parquet` is a usage error and
  exits 2, printing no traceback for any command. Pinned by a test, because it
  is a real trap.
- **Stream matters.** The log handler writes to stdout on a TTY and to stderr
  when stdout is piped, so `gpio ... --verbose | grep Traceback` finds nothing
  on any of the eleven.

**Root cause.** `configure_verbose()` in `core/logging_config.py` is
deliberately one-way, and is called from ~60 nested *core* functions. On a path
that fails before reaching one, the package logger is still at INFO when
`Group.invoke` emits its DEBUG record, so the record is dropped. Measured
directly: the logger's level at the boundary was DEBUG on exactly the seven
commands that printed the traceback and INFO on the four that did not.
`core/hilbert_order.py` never calls it at all.

**Fix.** `--verbose` now raises the level as *the flag is parsed*, via an eager
Click option callback (`decorators.enable_verbose_logging`), applied to
`verbose_option` and to `check.py`'s six inline declarations. An option callback
runs inside `make_context` for the subcommand — after the root group's
`setup_cli_logging(verbose=False)` and before the command body — which is the
only point early enough to be unconditional. 58 of 59 leaf commands carry the
flag (`gpio skills` is the exception, and touches no DuckDB).

Pinned by real `CliRunner().invoke(cli, [...], "--verbose")` runs on the four
commands that were broken plus one that was not, asserting
`"Traceback (most recent call last)"` in the output — not by a synthetic group
under `caplog.at_level(DEBUG)`, which passed while the real path dropped the
record.

This matters because gpio's own bugs surface as `InvalidInputException` too —
measured: `Could not create projection: EPSG:4326 -> EPSG:999999`,
`Expected geometry type at position '0'`, `Values were not provided for the
following prepared statement parameters`. With the traceback unrecoverable, a
gpio bug on those commands printed one line and nothing else.

## Which DuckDB failures, and which deliberately not

**This is narrower than `duckdb.Error`, on purpose.** gpio authors every SQL
string it runs, so a query that will not parse or bind is never news about the
user's file — it is a bug in something we generated, the defect class behind
#700, #718 and #944. Answering those with `Error: Parser Error: ...` would use
the line gpio reserves for "your input is bad" to report gpio's own bugs,
sending a user to inspect data that is fine.

DuckDB's hierarchy gives no base class that splits along this line. Measured
against duckdb 1.5.5, `duckdb.Error`'s only direct subclass is `DatabaseError`,
and `InvalidInputException` — the one #983 is about — sits under
`ProgrammingError` beside `ParserException`, `BinderException` and
`CatalogException`, its exact opposite. So the set is enumerated:

| Caught (input-caused) | Falls through, traceback intact (gpio-caused) |
|---|---|
| `InvalidInputException` — the file's bytes/metadata are unreadable | `ParserException` — SQL gpio generated will not parse |
| `IOException` (and `HTTPException` below it) — missing, unreachable, unreadable | `BinderException` — a column gpio referenced does not exist |
| | `CatalogException` — a function gpio called does not exist |

Measured, not assumed:

```
#983 geo block DuckDB refuses    -> InvalidInputException
gpio SQL bug: syntax             -> ParserException
gpio SQL bug: unknown column     -> BinderException
gpio SQL bug: unknown func       -> CatalogException
gpio SQL bug: unquoted ident     -> ParserException
missing file                     -> IOException
unreadable remote                -> IOException
```

All four of #983's reproductions raise `InvalidInputException`, so they stay
fixed — pinned by a test that reads the wrapped exception's cause rather than
assuming it. The negative direction is pinned too, by really executing malformed
SQL so the class is DuckDB's answer and not the test's guess.

The classes are checked on the outermost exception only, while `spill_space_hint`
and `is_unpublished_extension_error` walk the `__cause__` chain. That asymmetry
is deliberate and documented: those two match on *text*, which survives being
re-raised inside a wrapper, so they have to look down the chain to find it. This
matches on *class*, and a class that a gpio frame has already wrapped is no
longer unowned — the wrapper is gpio saying what it thinks happened, and its
class is the answer to use.

### `ConversionException` was dropped

It was in the tuple in the first version of this branch. It is not any more, and
that is a decision rather than an oversight:

- **No reproduction reaches an unowned boundary with one.** Every other member
  has a measured one; this had none.
- **The live conversion site is already owned, with a better message.**
  `core/convert.py` answers a CSV whose sniffed type breaks on a later row with
  `Error: Conversion failed: Conversion Error: CSV Error on Line: 30002 ...` —
  measured on a 30,001-row CSV — and never reaches the boundary.
- **The one refusal known to carry "Conversion Error" in its text is not one.**
  #988's curved WKB is raised as `InvalidInputException` (measured against
  `tests/data/curved_geometry_test.gdb`: the full message is
  `Invalid Input Error: Unsupported geometry type in WKB`), so it is covered by
  the first entry anyway.
- **Membership has to be decidable.** gpio writes every `CAST` it runs, so
  whether such a failure is the data's fault or gpio's choice of target type is
  not decidable from the exception. Adding it back needs a reproduction of one
  reaching the boundary unowned.

### An unpublished extension is not a bad input file

`INSTALL geography FROM community` failing with an HTTP 404 arrives as
`HTTPException`, which hangs below `IOException` — so without asking first, it
would read as "your input file is unreachable" and lose #778's guidance. Three
`INSTALL` sites are unwrapped (`core/duckdb_utils._install_and_load_extension`,
`core/extract_bigquery`, `core/partition/admin_hierarchical`), so their failures
arrive here raw. `core/exceptions.unpublished_extension_hint()` — same shape as
`spill_space_hint` — is asked in the same place, and names the registry rather
than the file.

It requires the message to name an extension, not just carry a 404: a remote
*input* that is simply not there 404s too, and still gets the plain error line.
Pinned both ways.

### The message is sanitized

`core/exceptions.sanitize_url_for_logging` is applied at ~25 sites because
presigned URLs carry credentials in the query string; the boundary applied none
of it, and `IOException`/`HTTPException` carry the full URL:
`HTTP GET error on 'https://bucket/x.parquet?X-Amz-Signature=...'`. Pre-existing
and byte-identical to `main` on the checked paths — not a regression — but the
boundary is the single funnel, which is this PR's own argument for where such
things live.

`core/exceptions.sanitize_error_message()` now runs on every message
`cli_error_for` builds. It redacts each URL in the text through the existing
helper, and strips ANSI escapes and control characters (tab and newline
excepted): DuckDB quotes offending values back verbatim, so a hostile file could
otherwise author a clean `Error:` line of its own with `\r` and `\x1b[2K`.

## The real count is eleven, not four

#983 found four commands while verifying something else. The sweep in
`tests/test_malformed_geo_metadata.py` (from #984) already runs every malformed
`geo` shape against 26 commands, so it gained a second assertion in the same
pass — no new harness, no extra runtime — plus the `{"version": "1.1.0"}` shape
from the issue, which is not *malformed* by gpio's standards and so was not in
the existing shape list.

On a clean `origin/main` it reports eleven commands leaking a `duckdb.Error`:

```
add bbox, add geometry-metrics, add quadkey, check all, check compression,
check optimization, check row-group, check spatial, extract geoparquet,
sort hilbert, sort str
```

across 13 of the 15 input shapes swept. The four in the issue are a subset.
Re-measured on a clean `origin/main` worktree after #988 landed: **identical**
— same eleven commands, same thirteen shapes. With the boundary in place the
count is **zero**, and the sweep's assertion is the wider `duckdb.Error`, so it
would also catch a malformed block driving gpio into generating bad SQL.

## Interaction with #988

#988 landed while this branch was open. It identifies DuckDB's curved-WKB
refusal by matching `"Unsupported geometry type in WKB"` on the exception,
linearizes the source and writes again. That refusal measures as an
`InvalidInputException` — exactly what this boundary claims — so if the boundary
ran anywhere but last, it would convert the error and the retry would never
fire.

It runs last, because it lives on `Group.invoke` and inner handlers are strictly
inside it. Two tests pin that on real code rather than on Python's exception
semantics: #988's real curved FileGDB still converts under `--skip-hilbert`
through the real `cli` and exits 0, and a refusal the handler *declines* still
becomes an error line. All of #988's own tests pass unchanged; `convert` never
appeared in the leak list at all (it carried its own handlers all along).

## Known, reasoned costs

- **gpio's own scratch files.** An `IOException` on a file gpio itself minted —
  partition staging, an `add <index>` intermediate — is now reported with the
  same line as one on the user's input, because the class is all the exception
  says. No live case could be constructed: apostrophe paths and
  glob-metacharacter paths behave identically with and without the boundary. The
  message still names the path that failed. Noted in `ErrorBoundaryGroup`'s
  docstring rather than guarded against.
- **`gpio process` is already wider than this policy.** `cli/commands/process.py`
  `_aggregate_error` catches `duckdb.Error` broadly, so Parser/Binder errors
  there already become error lines. Pre-existing, and arguably right for that
  one command: `--where` is *user-authored* SQL, so a binder error genuinely is
  news about the user's input (#612). Named here, not changed.

## Verification

The four reproductions from the issue, on a clean `origin/main` worktree with
the issue's one-row file:

```
$ gpio check optimization file.parquet
  File ".../core/duckdb_metadata.py", line 515, in get_file_metadata
    result = connection.execute(f"""
_duckdb.InvalidInputException: Invalid Input Error: Geoparquet metadata does not have a columns object

$ gpio add bbox file.parquet out.parquet
  File ".../core/common.py", line 3909, in add_computed_column
    total_count = con.execute(f"SELECT COUNT(*) FROM {sql_path(input_path)}").fetchone()[0]
_duckdb.InvalidInputException: Invalid Input Error: Geoparquet metadata does not have a columns object

$ gpio sort hilbert file.parquet out.parquet
  File ".../core/hilbert_order.py", line 541, in _hilbert_order_file_based
    empty_count_result = con.execute(f"""
_duckdb.InvalidInputException: Invalid Input Error: Geoparquet metadata does not have a columns object

$ gpio add quadkey file.parquet out.parquet
  File ".../core/write_strategies/duckdb_kv.py", line 737, in _write_plain_parquet_from_query
    con.execute(copy_query)
_duckdb.InvalidInputException: Invalid Input Error: Geoparquet metadata does not have a columns object
```

and on this branch, all four:

```
Error: Invalid Input Error: Geoparquet metadata does not have a columns object   (exit 1)
```

with no traceback in stderr — and the traceback back in full under `--verbose`.

Rebased on `origin/main` after #987 and #988 landed; clean rebase, no conflicts.

- `uv run pytest -n auto -m "not slow and not network and not meta"` —
  **7637 passed**, 76 skipped, 7 xfailed. One run of the same command failed
  `test_add_bbox_to_sort_hilbert`; it passes serially, and subprocess pipe tests
  are a known parallel-load flake on this tree.
- `uv run pytest -m meta` — **205 passed**.
- `uv run pre-commit run --all-files` and `--hook-stage pre-push` — all passed,
  `menard-check`, `vulture`, `xenon`, `import-linter` and `mypy` included.
- diff-cover against `origin/main`: **100%** (58 changed lines, 0 missing).

### On the `cli_surface.json` diff

One line, and it is only the root group's recorded class name
(`SpillAwareGroup` → `ErrorBoundaryGroup`). No command, group or option is
added, removed or changed — the `--verbose` callbacks are not part of the
recorded surface. Re-recorded with `GPIO_UPDATE_SNAPSHOT=1`.

*Authored by Claude (AI); every reproduction, exception class and `--verbose`
row above was measured by running it, not inferred.*

## Summary by CodeRabbit

- **Bug Fixes**
  - CLI commands now show clear error messages instead of raw Python tracebacks for invalid GeoParquet metadata and input-related database errors.
  - Error messages now provide more relevant guidance for insufficient temporary storage and unavailable extensions.
  - Sensitive URL signatures, terminal formatting codes, and unsafe control characters are removed from displayed errors.
  - Unrelated errors continue to surface for troubleshooting.

- **Improvements**
  - `--verbose` now enables detailed debug logging immediately, restoring traceback details when needed.

### DIFF
--- a/geoparquet_io/cli/commands/check.py
+++ b/geoparquet_io/cli/commands/check.py
@@ -16,6 +16,7 @@ from geoparquet_io.cli._shared import _activate_s3, create_default_group, init_g
 from geoparquet_io.cli.decorators import (
     GlobAwareCommand,
     check_partition_options,
+    enable_verbose_logging,
     handle_geoparquet_errors,
     overwrite_option,
     verbose_option,
@@ -177,7 +178,13 @@ class MultiFileCheckRunner:
 
 @check.command(name="all", cls=GlobAwareCommand)
 @click.argument("parquet_file")
-@click.option("--verbose", is_flag=True, help="Print detailed diagnostics")
+@click.option(
+    "--verbose",
+    is_flag=True,
+    help="Print detailed diagnostics",
+    callback=enable_verbose_logging,
+    is_eager=True,
+)
 @click.option("--fix", is_flag=True, help="Fix detected issues")
 @click.option(
     "--fix-output",
@@ -462,7 +469,13 @@ def check_all(
     show_default=True,
     help="Max rows for spatial order check",
 )
-@click.option("--verbose", is_flag=True, help="Print detailed diagnostics")
+@click.option(
+    "--verbose",
+    is_flag=True,
+    help="Print detailed diagnostics",
+    callback=enable_verbose_logging,
+    is_eager=True,
+)
 @click.option("--fix", is_flag=True, help="Fix with Hilbert ordering")
 @click.option(
     "--fix-output",
@@ -606,7 +619,13 @@ def check_spatial(
 
 @check.command(name="compression", cls=GlobAwareCommand)
 @click.argument("parquet_file")
-@click.option("--verbose", is_flag=True, help="Print detailed diagnostics")
+@click.option(
+    "--verbose",
+    is_flag=True,
+    help="Print detailed diagnostics",
+    callback=enable_verbose_logging,
+    is_eager=True,
+)
 @click.option("--fix", is_flag=True, help="Recompress geometry with ZSTD")
 @click.option(
     "--fix-output",
@@ -694,7 +713,13 @@ def check_compression_cmd(
 
 @check.command(name="bbox", cls=GlobAwareCommand)
 @click.argument("parquet_file")
-@click.option("--verbose", is_flag=True, help="Print detailed diagnostics")
+@click.option(
+    "--verbose",
+    is_flag=True,
+    help="Print detailed diagnostics",
+    callback=enable_verbose_logging,
+    is_eager=True,
+)
 @click.option("--fix", is_flag=True, help="Fix bbox (add for v1.x, remove for v2/parquet-geo)")
 @click.option(
     "--fix-output",
@@ -827,7 +852,13 @@ def check_bbox_cmd(
 
 @check.command(name="row-group", cls=GlobAwareCommand)
 @click.argument("parquet_file")
-@click.option("--verbose", is_flag=True, help="Print detailed diagnostics")
+@click.option(
+    "--verbose",
+    is_flag=True,
+    help="Print detailed diagnostics",
+    callback=enable_verbose_logging,
+    is_eager=True,
+)
 @click.option("--fix", is_flag=True, help="Optimize row group size")
 @click.option(
     "--fix-output",
@@ -1075,7 +1106,13 @@ def check_spec(
 
 @check.command(name="optimization", cls=GlobAwareCommand)
 @click.argument("parquet_file")
-@click.option("--verbose", is_flag=True, help="Print detailed diagnostics")
+@click.option(
+    "--verbose",
+    is_flag=True,
+    help="Print detailed diagnostics",
+    callback=enable_verbose_logging,
+    is_eager=True,
+)
 @check_partition_options
 @click.pass_context
 def check_optimization_cmd(
--- a/geoparquet_io/cli/decorators.py
+++ b/geoparquet_io/cli/decorators.py
@@ -34,7 +34,9 @@ class ErrorBoundaryGroup(click.Group):
       file, not a gpio bug -- and DuckDB's message already says what is wrong
       with it. Eleven commands answered such a file with a traceback because the
       handling lived in per-site ``except`` blocks that those paths never
-      reached (#983).
+      reached (#983). Only the DuckDB errors the *input* causes are translated;
+      the ones gpio's own generated SQL causes keep their traceback, which is
+      the line ``INPUT_FILE_DUCKDB_ERRORS`` draws.
 
     Both belong on the *root* group rather than on a decorator, because every
     command can hit them: ``handle_geoparquet_errors`` is applied to four
--- a/geoparquet_io/cli/decorators.py
+++ b/geoparquet_io/cli/decorators.py
@@ -20,29 +20,30 @@ from geoparquet_io.core.parquet_writer import (
 class ErrorBoundaryGroup(click.Group):
     """Root group that turns a DuckDB failure into gpio's error line.
 
-    Two failures reach this point with no gpio frame willing to own them, and
-    both used to leave a raw Python traceback on the terminal:
+    Failures reach this point with no gpio frame willing to own them, and used
+    to leave a raw Python traceback on the terminal -- an out-of-spill-space
+    error with no ``TMPDIR`` line (#752), and a ``geo`` block DuckDB's Parquet
+    reader refuses, on eleven commands whose paths never reached a per-site
+    ``except`` (#983). :func:`~geoparquet_io.cli.exception_handler.cli_error_for`
+    decides what is answered and what is not; the reasoning for where that line
+    falls lives once, on
+    :data:`~geoparquet_io.core.duckdb_utils.INPUT_FILE_DUCKDB_ERRORS`.
 
-    * DuckDB runs out of room to spill. It caps its spill at
-      ``max_temp_directory_size`` (90% of the volume by default), so a small or
-      RAM-backed temp volume fails cleanly rather than filling the disk -- but
-      it fails saying "Out of Memory Error", and its list of "possible
-      solutions" is entirely about memory. It never mentions the one knob that
-      fixes a disk shortage, so the boundary adds that line (#752).
-    * DuckDB's Parquet reader refuses the input's ``geo`` block. That block is
-      arbitrary JSON from somebody else's tool, so its refusal is news about the
-      file, not a gpio bug -- and DuckDB's message already says what is wrong
-      with it. Eleven commands answered such a file with a traceback because the
-      handling lived in per-site ``except`` blocks that those paths never
-      reached (#983). Only the DuckDB errors the *input* causes are translated;
-      the ones gpio's own generated SQL causes keep their traceback, which is
-      the line ``INPUT_FILE_DUCKDB_ERRORS`` draws.
+    This belongs on the *root* group rather than on a decorator, because every
+    command can hit those failures: ``handle_geoparquet_errors`` is applied to
+    four command modules, and ``gpio sort hilbert`` -- the worked example in the
+    docs -- is not one of them. ``Group.invoke`` is the single funnel every
+    subcommand passes through, so a command added tomorrow inherits this, and
+    inner handlers are strictly inside it and still run first (#988's
+    curved-geometry retry depends on that).
 
-    Both belong on the *root* group rather than on a decorator, because every
-    command can hit them: ``handle_geoparquet_errors`` is applied to four
-    command modules, and ``gpio sort hilbert`` -- the worked example in the docs
-    -- is not one of them. ``Group.invoke`` is the single funnel every
-    subcommand passes through, so a command added tomorrow inherits this.
+    A caveat worth naming: an ``IOException`` on a scratch file gpio itself
+    minted -- partition staging, an ``add <index>`` intermediate -- is reported
+    with the same line as one on the user's input, because the class is all the
+    exception says. No live case has been constructed (apostrophe and
+    glob-metacharacter paths behave identically with and without this boundary),
+    and the message still names the path that failed, so it is a known,
+    reasoned cost rather than an unnoticed one.
 
     Anything ``cli_error_for`` does not claim is re-raised untouched: a genuine
     gpio bug still gets its traceback.
@@ -58,8 +59,12 @@ class ErrorBoundaryGroup(click.Group):
             error = cli_error_for(exc)
             if error is None:
                 raise
-            # The traceback is hidden, not discarded: --verbose still prints it,
-            # which is what a gpio bug that surfaces as a DuckDB error needs.
+            # The traceback is hidden, not discarded: --verbose prints it, which
+            # is what a gpio bug that surfaces as a DuckDB error needs. That
+            # holds because `verbose_option` raises the package logger to DEBUG
+            # as the flag is parsed -- before the command body runs, so before
+            # anything can fail on the way to a core function's own
+            # `configure_verbose` call.
             get_logger(__name__).debug(
                 "Converted a low-level failure to an error line", exc_info=exc
             )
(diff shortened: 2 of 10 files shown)


cholmes:
Reviewed the boundary itself rather than the reproductions. The mechanism is right — one funnel, spill hint asked first, `None` meaning "not mine" so real bugs keep their traceback. One narrowing worth making before this lands.

## `duckdb.Error` is wider than the reasoning behind it

The docstring's justification is sound but specific:

> A `duckdb.Error` on a read is almost always a property of the input file.

"Almost always" is carrying the weight. **gpio authors every SQL string it runs**, so two `duckdb.Error` subclasses are never a property of the input file and always a gpio bug:

```
gpio-authored SQL bug (syntax)   -> ParserException    isinstance(duckdb.Error)=True
gpio-authored SQL bug (binder)   -> BinderException    isinstance(duckdb.Error)=True
```

(`duckdb.Error`'s only direct subclass is `DatabaseError`, and everything hangs below it, so `isinstance(exc, duckdb.Error)` is effectively "any DuckDB error at all".)

After this change, a quoting bug that produces malformed SQL — the class of defect #700, #718 and #944 were all about — stops printing a traceback and prints `Error: Parser Error: syntax error at or near ...` with exit 1, which is the same shape gpio uses to say "your file is bad". A user reports a broken file; the actual bug is ours, in a query we generated.

The DEBUG-level traceback mitigates this for whoever knows to pass `--verbose`, but the default answer now misattributes gpio's bugs to the user's data.

## Suggested narrowing

Catch the subclasses whose cause really is the input, and let the rest re-raise:

- `InvalidInputException` — the one #983 is about (`Geoparquet metadata does not have a columns object`)
- `IOException` — unreadable/missing/permission
- `ConversionException` — a value the file holds that will not cast

and leave `ParserException` / `BinderException` (and anything else gpio constructs) falling through to `None`, traceback intact.

That keeps every reproduction in the issue working — all four are `InvalidInputException` — while not converting our own SQL defects into input-file diagnoses. It also makes the docstring's claim literally true instead of "almost always".

Worth a test in `tests/test_cli_error_boundary.py` pinning the negative direction: a deliberately malformed gpio-authored query must still raise, not become a `ClickException`. The suite currently proves the positive direction only.

*Review by Claude (AI); the subclass behaviour above was measured against duckdb 1.5.5, not inferred.*

cholmes:
Narrowed as suggested, and I verified the subclass behaviour against duckdb 1.5.5 independently rather than taking the list on trust. Your three are right; one measurement worth adding is *why* they have to be enumerated.

`duckdb.Error`'s only direct subclass is `DatabaseError`, as you said — but the split you want does not exist anywhere further down either. `InvalidInputException` sits under `ProgrammingError`, the same parent as `ParserException`, `BinderException` and `CatalogException`:

```
Error -> DatabaseError -> ProgrammingError -> BinderException
                                              CatalogException
                                              InvalidInputException   <- #983
                                              ParserException
                                              SyntaxException
```

So any ancestor wide enough to admit the one we want admits the three we don't. The tuple is now `(InvalidInputException, IOException, ConversionException)`, enumerated, with that reason recorded next to it and a test asserting the collapse is impossible.

Measured triggers:

```
#983 geo block DuckDB refuses    -> InvalidInputException
gpio SQL bug: syntax             -> ParserException
gpio SQL bug: unknown column     -> BinderException
gpio SQL bug: unknown func       -> CatalogException     (a fourth for the excluded side)
gpio SQL bug: unquoted ident     -> ParserException      (#718's shape)
missing file                     -> IOException
unreadable remote                -> IOException
bad cast of a file's value       -> ConversionException
```

`CatalogException` is the one addition to your list, on the excluded side — it is already excluded by the tuple, but it is now named in the guard so a future widening has to confront it.

Confirmed rather than assumed: all four of #983's reproductions are `InvalidInputException`, checked through the wrapped exception's `__cause__` in a test, so a future narrowing that dropped one would fail loudly instead of quietly restoring a traceback.

Negative direction added, raised by really executing malformed SQL so the class is DuckDB's answer and not the test's guess — plus an end-to-end one through the real boundary, and a guard that the tuple is not `duckdb.Error`/`DatabaseError` in disguise.

Sweep re-run: still zero leaks, and its assertion is still the wider `duckdb.Error`, so it would also catch a malformed block driving gpio into generating bad SQL.

One note on `HTTPException`: it hangs below `IOException`, so `isinstance` picks it up without a separate entry. A remote URL that will not fetch is a property of the path the user named, so I kept it in scope and said so in a test.

*Reply by Claude (AI); every class above was measured on duckdb 1.5.5, not inferred.*

Issue #953, opened 2026-09-09
fix(metadata): the secondary geometry column's covering is attached to the primary


cholmes:
## What

Under `--write-strategy in-memory` and `--write-strategy streaming`, the **primary** geometry column comes out of a multi-geometry file claiming a `covering` that belongs to the **secondary**:

```
columns.geometry.covering.bbox → ["boundary_bbox", …]
```

The primary here is a Point column; `boundary_bbox` is the secondary Polygon column's bbox struct. The metadata is simply wrong — a reader using the covering to prune row groups for the primary column would be consulting the wrong column's extents.

## Scope

**Pre-existing.** Reproduced byte-identically on `origin/main` (21b14fe); it is not a regression from #949, which is why it is filed separately rather than fixed there.

Found by an adversarial review of #949.

PR #997, opened 2026-09-11
fix(add)!: keep native geo and the CRS when a gpio add subcommand rewrites a file

### LINKED ISSUE #993: fix(add)!: the add * scratch write drops native geo and the CRS, so five partition drivers and sort quadkey still downgrade
## What

`gpio add bbox/quadkey/h3/s2/a5/kdtree/geometry-metrics` rewrite a native-geo-only input as **1.1 WKB**, losing the Parquet GEOMETRY logical type *and* the CRS with it. None of them pass `input_file=` to `write_parquet_with_metadata`, so the write facade's auto-version rule (#990) never sees the real input.

That single cause produces two visible defects downstream, both pinned as `xfail(strict=True)` in `tests/test_write_facade_version_owner.py`:

### 1. Five partition drivers do not deliver #600

`partition quadkey/h3/s2/a5/kdtree` hand `partition_by_column` the scratch file their `add <index>` step wrote, not the user's file, so auto mode resolves from something that is already a 1.1 rewrite. Measured on `tests/data/fields_pgo_5070_snappy.parquet`:

| driver | before #990 | after #990 |
|---|---|---|
| `partition string` | 1.1.0, WKB | **2.0.0, native** ✅ |
| `partition admin` | 1.1.0, WKB | **2.0.0, native, EPSG:5070** ✅ |
| `quadkey` / `h3` / `s2` / `a5` / `kdtree` | 1.1.0, WKB | 1.1.0, WKB — unchanged |

### 2. `sort quadkey` asserts the wrong CRS

It resolves the *version* from the user's file but reads *data* from the scratch file, so the two disagree. The 2.0 native write then has to materialise some CRS and materialises `OGC:CRS84` over EPSG:5070 data:

```
sort hilbert / sort column / extract geoparquet  ->  2.0.0, native, EPSG:5070   ✅
sort quadkey                                     ->  2.0.0, native, OGC:CRS84   ❌
```

Net reader-visible CRS is the same wrong answer as before (an absent `crs` already means CRS84), so this is not new corruption — but the file now *affirmatively asserts* the wrong CRS, and `gpio check spec` prints `✓ column "geometry" has valid inline PROJJSON CRS` over it.

## Why it was not fixed in #990

The obvious fix — resolve from the user's file in the five index drivers — would have fixed the version and **imported `sort quadkey`'s wrong-CRS defect onto five more commands**, since the data still comes from the scratch file. #990 therefore narrowed its behaviour-6 claim to `sort hilbert`/`sort column`/`sort str`/`extract geoparquet`/`partition string`/`partition admin` and pinned the rest, rather than shipping a wider wrong answer.

Fixing the scratch write is the right cut, and it is a separate breaking change: it alters the output bytes of eight more commands, and each needs its own CRS measurement.

## Scope

- Make the `add *` family preserve the input's version and CRS on its scratch write (or teach it to pass `input_file=` through the facade).
- Delete the six `xfail(strict=True)` markers in `tests/test_write_facade_version_owner.py` — they will fail loudly once this lands, which is the point.
- This closes the remainder of **#600**, which #990 downgraded to `Refs` for exactly this reason.

Needs a `!` title: it changes output for every `gpio add` subcommand.

Found by an adversarial review of #990.

cholmes:
## What changed

Every `gpio add` subcommand rewrites its whole input to append one column, and
none of them passed `input_file=` to `write_parquet_with_metadata`. The write
facade (#990) therefore had no witness to answer either of the two questions a
rewrite of a **native-geo-only** file has to answer:

- **Which version?** Auto mode read the carried `geo` key, and a native-geo-only
  input has none. It landed on the 1.1 default, DuckDB re-encoded the geometry
  column as plain WKB, and the Parquet `GEOMETRY` logical type went with it.
- **Which CRS?** That logical type is the *only* place a native-geo-only file
  stores its CRS, so `input_crs` was `None` and nothing described the output's
  CRS at all. `EPSG:5070` in, nothing declared out — which every reader resolves
  as `OGC:CRS84`.

Two commands read that downgraded scratch file rather than the user's, so the
loss reached them too, and both were pinned `xfail(strict=True)` in
`tests/test_write_facade_version_owner.py`: `gpio sort quadkey` and
`gpio partition quadkey/h3/s2/a5/kdtree`.

### The two answers have to come from the same witness

The first draft of this PR passed `input_file=` at the seven `add`-family sites
and stopped there. That fixes the version and leaves the CRS unanswered — which
is **worse than the bug it replaces**, because the output is then native 2.0
with EPSG:5070 in the Parquet logical type and no `crs` key in the `geo` block.
One file, two CRSs, and gpio's own checker saying so:

```
✗ CRS in geo metadata must match CRS in Parquet schema
    Metadata: OGC:CRS84 (default), Schema: NAD83 / Conus Albers (EPSG:5070)
```

So the facade gains a third resolver, `parquet_writer.resolve_input_crs`,
reading the CRS from the same `input_file` witness the version already came
from, applied next to it in `write_parquet_with_metadata`. Not at seven call
sites: splitting one question between many answerers is the defect class #990
exists to end.

It also repairs a quieter loss on the same inputs. `--geoparquet-version 1.1` on
a native-geo-only file has nowhere but the `geo` block to record a CRS, and
before this it recorded none — a silent downgrade to CRS84 that no version
assertion could have caught.

The facade still does **not** own what `crs_utils.apply_output_crs` owns: the
null-vs-default rule, stripping a stale value, omitting the default. This only
*finds* the input's CRS. An explicit `input_crs` always wins — `gpio convert
reproject` passes the CRS it transforms *to*, and `gpio convert` the one a
non-Parquet source's geometry is in (`--crs`, for CSV/TSV input) — so those
paths are untouched (measured below).

"Only finds" is not "only informs one place", though, and the facade docstring
now says so: a non-`None` answer where the write previously had `None` moves
three things at once — `apply_output_crs` writes an explicit `crs` into the
`geo` block, `_wrap_query_with_crs` wraps the query in `ST_SetCRS` so DuckDB
stamps the CRS into the Parquet logical type, and `geoarrow_encoding` attaches
it to the Arrow extension type. The `ST_SetCRS` wrap is the load-bearing one:
without it DuckDB restates whatever the column it read carried, which is how
`sort quadkey` came to stamp `OGC:CRS84` on EPSG:5070 data.

### An output with no geometry column declares no CRS

Review found the one case where the facade's answer reaches an output that has
nowhere to put it. A column projection can drop the geometry column entirely,
and the resolver then answered a question with no subject:

```
$ gpio extract geoparquet tests/data/fields_pgo_5070_snappy.parquet out.parquet --exclude-cols geometry
ValueError: geometry_column is required when input_crs is specified — cannot apply CRS without a geometry column
   rc=1
```

`_wrap_query_with_crs` got a CRS and no column to `ST_SetCRS`. A traceback, not
even an error line, out of a command that worked before this branch. It needed
all three of: an output resolving to 2.0/native (so the plain-`COPY` fast path
is taken), an input with a non-default CRS, and a projection that drops the
geometry column — which is why a 1.1 input, taking the rewrite-strategy path,
never showed it.

The gate is in `resolve_input_crs`, before any witness is consulted and whatever
the caller named: the one place that owns the decision is the one place that
knows the answer can be `None`. Both of review's triggers now exit 0 —
`--exclude-cols geometry` on the native-geo-only fixture, and
`--exclude-cols geometry,bbox` on a 2.0 input with a `geo` block — each writing
the seven attribute columns and 100 rows, with no `geo` key.

### The Vecorel schema fix, measured rather than assumed

`constants._fix_vecorel_schema`, which `add geometry-metrics` runs last, is a
pyarrow read/write of the whole file. Plain pyarrow reads a Parquet `GEOMETRY`
column as `binary` and writes it back as `binary`, so with the fix above in
place it would turn a native 2.0 file into a 2.0 file with no native geo type.

This turns out to depend on nothing in the file and everything in the process:
`geoarrow.pyarrow`'s import registers the extension types that make pyarrow
materialise those columns as `geoarrow.wkb`, and that registration is global. The
same code, same input:

| | `check spec` on `add geometry-metrics` output |
|---|---|
| under pytest (fixtures import `geoarrow.pyarrow`) | 34 passed, **0 failed** |
| from a fresh CLI process | 26 passed, **4 failed** |

The fix is therefore one line — `_fix_vecorel_schema` imports `geoarrow.pyarrow`
itself — and `tests/test_add_preserves_native_geo.py` runs that one command in a
**subprocess**, because no in-process test can see this class of defect.
Re-verified as load-bearing by deleting that one line: exactly the subprocess
test fails (`KeyError: 'geometry'` — the native type gone), while its
in-process sibling over the same command stays green.

An earlier draft carried ~100 lines for this instead: a `native_geo_types_in_file`
helper that parsed each column's logical type out of JSON, a `_geography_edge_type`
mapper, and a `_conform_to_schema` that duplicated `disk_rewrite._conform_row_group`.
Review found four defects in it, and the measurement dissolved all four along with
the code they were in — pyarrow already round-trips the logical type, the CRS *and*
a GEOGRAPHY's edge type once its extension types are registered. What review found,
and where it went:

| finding | severity | disposition |
|---|---|---|
| `json.loads` on a CRS that is `EPSG:32633` or `srid:...` rather than PROJJSON would raise | **latent, not CLI-reachable** — no gpio write path emits a non-PROJJSON CRS into a logical type, so no command could reach it | code deleted |
| `_geography_edge_type("planar")` returned `PLANAR`, against its own docstring's guarantee that every GEOGRAPHY algorithm is non-planar | would have silently replanned a GEOGRAPHY's edges | code deleted; pyarrow preserves the edge type, verified on a spherical fixture |
| `_conform_to_schema` duplicated `disk_rewrite._conform_row_group` | duplication | code deleted |
| the helper lived in a module named `constants` while importing from `write_strategies` | layering | code deleted |

No separate PR was opened for it: there is nothing left to review that this
does not contain.

## Measured, per command

`gpio check spec --json` on the output of every command, plus the `geo` block's
`crs` and the Parquet logical type's `crs` read **separately** — because
`crs_utils.source_crs_string` reads the first, falls back to the second, and
returns whichever answered. It reports *an* answer for a file that holds two,
which is exactly how the first draft of this PR passed its own tests.

Input: 200 polygons spread across the lower 48, EPSG:5070 in the Parquet
`GEOMETRY` logical type, **no** `geo` key, `6 passed / 0 failed`. Built inside
5070's real area of use on purpose: `tests/data/fields_pgo_5070_snappy.parquet`
is labelled 5070 but its coordinates are over Europe, so it always scores one
`✗ coordinates outside valid range for CRS` and "clean" could never mean zero.

`before` is `origin/main` in a clean worktree; `after` is this branch rebased
onto it.

| command | `geo` block | Parquet logical type | `check spec` |
|---|---|---|---|
| **before** | | | |
| `add bbox` | 1.1.0, no `crs` key | WKB, no native type | 27p / **1F** |
| `add quadkey` / `h3` / `s2` / `a5` / `kdtree` / `geometry-metrics` | 1.1.0, no `crs` key | WKB, no native type | 23p / **1F** each |
| `add bbox-metadata` (on `add bbox`'s output) | 1.1.0, no `crs` key | WKB, no native type | 27p / **1F** |
| `partition quadkey` / `h3` / `s2` / `a5` / `kdtree` | 1.1.0, no `crs` key | WKB, no native type | 23p / **1F** (every part file) |
| `sort quadkey` | 2.0.0, no `crs` key | `Geometry`, **`OGC:CRS84`** | 33p / **1F** |
| **after** | | | |
| `add bbox` | 2.0.0, **EPSG:5070** | `Geometry`, **EPSG:5070** | 38p / **0F** |
| `add quadkey` / `h3` / `s2` / `a5` / `kdtree` / `geometry-metrics` | 2.0.0, **EPSG:5070** | `Geometry`, **EPSG:5070** | 34p / **0F** each |
| `add bbox-metadata` (on `add bbox`'s output) | 2.0.0, **EPSG:5070** | `Geometry`, **EPSG:5070** | 38p / **0F** |
| `partition quadkey` / `h3` / `s2` / `a5` / `kdtree` | 2.0.0, **EPSG:5070** | `Geometry`, **EPSG:5070** | 34p / **0F** (all 13 part files) |
| `sort quadkey` | 2.0.0, **EPSG:5070** | `Geometry`, **EPSG:5070** | 34p / **0F** |

The single `1F` on every `before` row is `coordinates outside valid range for
CRS`: the downgrade left EPSG:5070 metres in a file claiming CRS84. It is the
defect, not fixture noise — the input scores zero failures, and so does every
`after` row.

`sort quadkey`'s `before` row is the one worth reading twice. It is already
native 2.0, and it **asserts** `OGC:CRS84` inside the logical type over
EPSG:5070 data — a wrong CRS stated where the others merely omit one, which
`gpio check spec` then blesses with `✓ column "geometry" has valid inline
PROJJSON CRS`.

### What else the user sees change

**Files get bigger by a fixed amount.** The full EPSG:5070 PROJJSON (2,477
bytes) is now written into the `geo` block *and* the Parquet logical type, and
for `geometry-metrics` into `ARROW:schema` as well. On the same 200-row,
18,437-byte input:

| command | before | after | delta |
|---|---|---|---|
| `add bbox` | 18,989 | 23,460 | +4,471 |
| `sort quadkey` | 14,300 | 20,318 | +6,018 |
| `add geometry-metrics` | 14,491 | 25,814 | +11,323 |

It is a **constant, not a ratio** — metadata, not data. On a 200-row fixture
that reads as +24% to +78%; on a real file it is noise. Row groups are
unchanged: this does not touch `resolve_row_group_rows`, and the #990 row-group
concern does not apply.

**A `crs: null` input now prints a warning it did not print before.** The
resolver reaches `extract_crs_from_parquet`, which calls `warn_null_crs_once`,
so `gpio add *` on a file with an explicit `"crs": null` now says:

```
Input has an explicit null CRS (unknown). An explicit null CRS means the CRS is
*unknown* (not the OGC:CRS84 default). If the coordinates are really lon/lat
WGS84, run `gpio convert reproject <input> <output> --assume-crs84` to set the
default.
```

`main` was silent. This is an improvement — the information was always true and
never surfaced — but it is new output on a breaking-change PR, so it is declared
rather than discovered. The written `crs: null` is unchanged either way.

### Controls, unchanged before and after

`partition string`, `sort hilbert`, `sort column`, `sort str`,
`extract geoparquet`, `convert geoparquet` — all `2.0.0, EPSG:5070` in both
places, `34 passed / 0 failed`, on both sides. The paths that already read the
user's file directly do not move, and neither does the one that passes its own
`input_crs`.

`add admin-divisions` is not in the table: it needs a boundary dataset to join
against and belongs to the network lane. Its two write sites, and the
`add/country_codes.py` join it is built on (reached by `admin-divisions` and by
`gpio benchmark`), take the same `input_file=` change as the rest.

## Two defects this does not fix, pinned rather than left unmeasured

Both are now exercised end-to-end through the real command, with
`xfail(strict=True)` — so whoever fixes them has to come back and delete a
marker.

**#999 — a native `GEOGRAPHY` column comes back planar.** `add bbox` on a
spherical input writes a planar `Geometry` logical type while the `geo` block it
builds declares `"edges": "spherical"`, and `check spec` warns. Still better
than `main`, which wrote 1.1 WKB and dropped the `edges` declaration **outright**
— silently replanning the edges with nothing in the file saying so. The witness
here carries the CRS, not the edge type: DuckDB reads a GEOGRAPHY column as a
planar GEOMETRY, so the write has nothing left to restate it from.

**#1000 — a secondary geometry column is left undescribed.** Two native columns
(`geometry` EPSG:5070, `centroid` EPSG:3857) through `add bbox`: `centroid` keeps
its logical type, loses its own CRS, and never appears in `geo.columns`, which
2.0 requires. `main` scores this input **2 failures** (it wrote a 1.1 file that
still carried a native `centroid` type); this narrows it to 1. Its sibling test
`test_add_attaches_the_witness_crs_to_the_primary_column_only` **passes** and
pins the half that is right — a per-file witness must not become a per-file CRS
stamped onto every column. Sits next to #952 and #953.

## Also found, out of scope by this repo's separate-PRs rule

`core/check_fixes.py:115`, `:258` and `:437` call `write_parquet_with_metadata`
without `input_file=`, so `gpio check --fix` has the identical defect on the
identical inputs. Not bolted onto this branch: it is a different command with its
own before/after measurement to take. Review found a second `check --fix` defect
next to it — it drops a bbox covering on a 2.0 input, which this PR newly routes
users into by making more outputs 2.0 — filed with it.

## Docs

`docs/guide/add.md` gains the note that `add` preserves the input's version and
CRS, that `sort quadkey` and the five index `partition` drivers inherit it, and
that the two shapes above are not covered yet. It also states the reader
contract this change implies: a native 2.0 output needs a reader that
understands the Parquet `GEOMETRY` logical type — older pyarrow, geopandas and
DuckDB builds may not open one — and `--geoparquet-version 1.1` is the escape
hatch, writing plain WKB with the CRS in the `geo` block.

## Tests

`tests/test_add_preserves_native_geo.py` is rewritten around the oracle the first
draft lacked. Every end-to-end assertion reads the `geo` block and the Parquet
logical type separately and then asks `gpio check spec` whether they agree —
one test, four asserts, per subcommand, rather than three pipeline runs each. The
readers and fixture builders live in `tests/native_geo_probes.py` so
`test_write_facade_version_owner.py` measures its side of the scratch write
against the same input.

The six `xfail(strict=True)` markers in `test_write_facade_version_owner.py` are
deleted **with their assertions unchanged**, which is what the markers were for.
Those assertions go through `source_crs_string`, so they are not the evidence:
two new tests beside them read both sources separately and require `check spec`
to be clean, for `sort quadkey` and for every part file of all five index
partition drivers, on a projected fixture (the file's own `native_geo_only`
fixture is lon/lat, where an absent `crs` already means the right thing and a
CRS assertion cannot fail).

Three more from review: the dropped-geometry case above, through the CLI on both
input shapes and directly against `resolve_input_crs`. `spec_report` now tries
each `{` in turn rather than slicing from the first, so a `check spec` that
crashes before emitting JSON says so instead of raising `ValueError: substring
not found`. The subprocess probe gets a wedge cap of 300 s — deliberately not
the 30 s the repo's other subprocess tests use, because this one pays a cold
interpreter's full CLI import: 0.8 s idle, and over 30 s under `-n auto` with
coverage on 12 workers, measured after a 30 s cap turned that into a failure.

## Verification

- `uv run pytest -q -n auto -m "not slow and not network and not meta" --cov=geoparquet_io --cov-report=xml --cov-fail-under=0` — **7700 passed, 76 skipped, 3 xfailed** in 199s
- `uv run diff-cover coverage.xml --compare-branch=origin/main --fail-under=90` — **100%**, 12 changed lines, 0 missing
- `uv run pytest -q -m meta` — **205 passed**
- `uv run pre-commit run --all-files` and `--hook-stage pre-push` — all green (ruff, codespell, duckdb-antipatterns, mypy, vulture, xenon, import-linter, deptry, menard-check, doc-sync)
- `uv run lint-imports` — 4 contracts kept, 0 broken
- `uv run pytest docs/guide -n 4 -m "not network"` — **236 passed, 224 skipped**
- `tests/data/cli_surface.json` unchanged

Refs #600. Closes #993.

Deferred from review, each with its own issue rather than riding along:

- **#1003** — `add bbox` → `check --fix` drops the covering on a native 2.0 file. Pre-existing in `check_fixes` (the three `write_parquet_with_metadata` calls there pass no `input_file=`); this PR newly routes users into it by making `add bbox` produce 2.0.
- **#1004** — when the `geo` block and the Parquet logical type disagree on the CRS, the resolver picks the block and writes it into both, silently. The input is already invalid, but a loud inconsistency becomes a clean-looking assertion; it should warn.
- **#1005** — the partition path is the one exemption from "version and CRS from one witness, together": `staging.py` passes neither, and the AST invariant asserts only that the version is pre-resolved, not that the CRS came from the same file.
- **#1006** — two streaming sites (`cli/commands/convert.py`, `core/streaming.py`) have the same latent `geoarrow.pyarrow` registration dependency fixed here for `_fix_vecorel_schema`; lossless end to end today, same shape. Plus a regression test for the GEOGRAPHY edge-type round-trip this PR relies on but does not pin.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rewriting GeoParquet files now preserves native geometry encoding, CRS, and format version across column additions, sorting, and spatial partitioning.
  * Inputs using native Parquet `GEOMETRY` types retain their metadata and remain valid after processing.
  * CRS information is preserved for projected datasets, including EPSG:5070.
  * Added support for explicitly writing GeoParquet 1.1 when compatibility with older readers is required.

* **Documentation**
  * Documented file-rewrite behavior, metadata preservation, supported format versions, and compatibility options.

### DIFF
--- a/docs/guide/add.md
+++ b/docs/guide/add.md
@@ -15,6 +15,11 @@ The `add` commands enhance GeoParquet files with spatial indices, geometry metri
     the CRS is preserved either way, including when you ask for 1.1, which has
     nowhere but the `geo` block to record it.
 
+    A native 2.0 output needs a reader that understands the Parquet `GEOMETRY`
+    logical type — older pyarrow, geopandas and DuckDB builds may not open one.
+    `--geoparquet-version 1.1` is the escape hatch: it writes plain WKB, and
+    keeps the CRS in the `geo` block.
+
     It is the same rewrite behind `gpio sort quadkey` and
     `gpio partition quadkey/h3/s2/a5/kdtree`, which add their index column to a
     scratch file first, so those commands preserve the version and the CRS now
--- a/docs/guide/add.md
+++ b/docs/guide/add.md
@@ -2,6 +2,32 @@
 
 The `add` commands enhance GeoParquet files with spatial indices, geometry metrics, and administrative metadata.
 
+!!! note "The output keeps the input's GeoParquet version and CRS"
+    Every `add` command rewrites the whole file to append its column, and with
+    no `--geoparquet-version` that rewrite preserves what the input was: a 1.1
+    file stays 1.1, a 2.0 file stays 2.0, and a file that carries a native
+    Parquet `GEOMETRY` type with no `geo` key is written as native 2.0. Until
+    [#993](https://github.com/geoparquet/geoparquet-io/issues/993) that last
+    case was re-encoded as 1.1 WKB, which stripped the logical type **and the
+    CRS stored inside it** — an `EPSG:5070` file came back out declaring
+    nothing, which every reader resolves as `OGC:CRS84`. Pass
+    `--geoparquet-version` to ask for something other than the input's version;
+    the CRS is preserved either way, including when you ask for 1.1, which has
+    nowhere but the `geo` block to record it.
+
+    It is the same rewrite behind `gpio sort quadkey` and
+    `gpio partition quadkey/h3/s2/a5/kdtree`, which add their index column to a
+    scratch file first, so those commands preserve the version and the CRS now
+    too.
+
+    Two shapes are not fully covered yet, both pinned by failing-on-purpose
+    tests: a native `GEOGRAPHY` column comes back as a planar `GEOMETRY` while
+    the metadata still declares spherical edges
+    ([#999](https://github.com/geoparquet/geoparquet-io/issues/999)), and a
+    *secondary* geometry column keeps its own Parquet type but is left out of
+    `geo.columns` and loses its CRS
+    ([#1000](https://github.com/geoparquet/geoparquet-io/issues/1000)).
+
 !!! note "Coordinate Reference Systems"
     The grid-index commands (`h3`, `s2`, `a5`, `quadkey`) and `admin-divisions` are
     **CRS-aware**. If your input declares a projected CRS (e.g. EPSG:5070 or
--- a/geoparquet_io/core/add/admin_divisions.py
+++ b/geoparquet_io/core/add/admin_divisions.py
@@ -776,6 +776,11 @@ def _execute_per_level_joins(
                 profile=profile,
                 geoparquet_version=geoparquet_version,
                 extra_kv_metadata=extra_kv,
+                # Every level is `SELECT a.*` over the one before it, so the
+                # geometry column the last write emits is the input's own --
+                # the witness auto mode needs to keep a native-geo-only input
+                # native rather than rewriting it as 1.1 WKB (#993).
+                input_file=input_source,
                 memory_limit=memory_limit,
             )
         else:
@@ -986,6 +991,9 @@ def add_admin_divisions_multi(
                     profile=profile,
                     geoparquet_version=geoparquet_version,
                     extra_kv_metadata=extra_kv,
+                    # The join is `SELECT a.*` plus admin columns, so the
+                    # geometry reaching the output is the input's own (#993).
+                    input_file=input_path,
                     memory_limit=memory_limit,
                 )
 
--- a/geoparquet_io/core/add/country_codes.py
+++ b/geoparquet_io/core/add/country_codes.py
@@ -747,6 +747,9 @@ def add_country_codes(
             row_group_size_mb=row_group_size_mb,
             row_group_rows=row_group_rows,
             verbose=verbose,
+            # The join keeps the input's own geometry column, so the input is
+            # the witness auto mode resolves the output version from (#993).
+            input_file=input_path,
         )
 
         _print_results_summary(con, output_parquet)
--- a/geoparquet_io/core/add/geometry_metrics.py
+++ b/geoparquet_io/core/add/geometry_metrics.py
@@ -298,6 +298,11 @@ def _add_vecorel_metadata_to_file(
                 original_metadata=metadata,
                 extra_kv_metadata=extra_kv,
                 verbose=verbose,
+                # A whole-file rewrite to attach one sidecar key must not also
+                # change the geometry encoding. Without this witness auto mode
+                # read the carried `geo` key, which does not say whether the
+                # column is native, and the rewrite re-encoded it as WKB (#993).
+                input_file=parquet_file,
                 memory_limit=memory_limit,
             )
         finally:
--- a/geoparquet_io/core/add/kdtree.py
+++ b/geoparquet_io/core/add/kdtree.py
@@ -659,6 +659,10 @@ def add_kdtree_column(
         verbose=verbose,
         profile=profile,
         geoparquet_version=geoparquet_version,
+        # The rows this write reads are the input's, so auto mode resolves the
+        # output version from the input file rather than from a `geo` key a
+        # native-geo-only input does not have (#993).
+        input_file=input_parquet,
         memory_limit=memory_limit,
     )
 
--- a/geoparquet_io/core/add/quadkey.py
+++ b/geoparquet_io/core/add/quadkey.py
@@ -703,11 +703,8 @@ def _add_quadkey_file_based(
             profile=profile,
             geoparquet_version=geoparquet_version,
             custom_metadata=quadkey_metadata,
-            # The rows this write reads are the input's, so the input is the
-            # witness auto mode resolves the output version from. Without it a
-            # native-geo-only input was re-encoded as 1.1 WKB and lost the CRS
-            # its GEOMETRY logical type carried -- which `gpio sort quadkey`
-            # then read back as its data (#993).
+            # The precondition: this query is `SELECT *` plus the quadkey
+            # column, so the rows it writes are `input_parquet`'s own.
             input_file=input_parquet,
             memory_limit=memory_limit,
         )
--- a/geoparquet_io/core/add/quadkey.py
+++ b/geoparquet_io/core/add/quadkey.py
@@ -703,6 +703,12 @@ def _add_quadkey_file_based(
             profile=profile,
             geoparquet_version=geoparquet_version,
             custom_metadata=quadkey_metadata,
+            # The rows this write reads are the input's, so the input is the
+            # witness auto mode resolves the output version from. Without it a
+            # native-geo-only input was re-encoded as 1.1 WKB and lost the CRS
+            # its GEOMETRY logical type carried -- which `gpio sort quadkey`
+            # then read back as its data (#993).
+            input_file=input_parquet,
             memory_limit=memory_limit,
         )
 
--- a/geoparquet_io/core/common.py
+++ b/geoparquet_io/core/common.py
@@ -75,6 +75,7 @@ from geoparquet_io.core.logging_config import (
 from geoparquet_io.core.parquet_writer import (
     ParquetWriteSettings,
     note_duckdb_copy_rounding,
+    resolve_input_crs,
     resolve_output_geoparquet_version,
     resolve_row_group_rows,
 )
@@ -2871,7 +2872,14 @@ def write_parquet_with_metadata(
         show_sql: Whether to print SQL statements before execution
         profile: AWS profile name (S3 only, optional)
         geoparquet_version: GeoParquet version to write (1.0, 1.1, 2.0, parquet-geo-only)
-        input_crs: PROJJSON dict with CRS from input file
+        input_crs: PROJJSON dict naming the CRS the output declares. Set by a
+            write that changes the coordinates (reproject, convert --to-crs),
+            where the output's CRS is not a fact any reading of the input could
+            supply. Left None by every rewrite that keeps its input's
+            coordinates: the facade then resolves it from ``input_file``, so a
+            native-geo-only input's CRS — which lives only in the Parquet
+            GEOMETRY logical type — reaches the output's ``geo`` block too
+            (#993).
         write_strategy: Write strategy to use. Options:
             - "duckdb-kv" (default): Use DuckDB COPY TO with KV_METADATA
             - "in-memory": Load entire dataset into memory
@@ -2886,9 +2894,15 @@ def write_parquet_with_metadata(
         extra_kv_metadata: Additional Parquet file-level KV metadata as {key: json_string}.
             Written alongside the 'geo' key (e.g., for Vecorel collection metadata).
         input_file: Path to the input parquet file, when the write rewrites an
-            existing file. Enables full-fidelity non-planar edges preservation
-            (native GEOGRAPHY logical types are only visible in the file's
-            schema); without it, edges still fall back to original_metadata.
+            existing file. The witness the facade resolves the output's version
+            and CRS from, and what enables full-fidelity non-planar edges
+            preservation (native GEOMETRY/GEOGRAPHY logical types, and the CRS
+            stored inside them, are only visible in the file's schema); without
+            it, both fall back to original_metadata's ``geo`` key, which a
+            native-geo-only input does not have. Must be the file whose *rows*
+            this write reads, or one lossless with respect to it: naming the
+            user's file while reading a scratch rewrite that dropped the native
+            type is how a wrong CRS comes to be asserted rather than omitted.
         invalidate_derived_stats: When True, strip the carried per-column
             ``bbox`` and ``geometry_types`` from ``original_metadata`` before
             building output geo metadata. Set by callers that transform geometry
@@ -2970,12 +2984,19 @@ def write_parquet_with_metadata(
         con, query, original_metadata, geometry_column, verbose
     )
 
-    # The facade owns both of these: how many rows a row group gets, and which
-    # version auto mode resolves to. Every caller of this function -- 21 of them
-    # -- used to inherit whatever its writer defaulted to and whatever its own
-    # `geo` key happened to say, which is how `convert` came to write 122,880-row
-    # groups (#981) and how `sort`/`extract`/`partition` came to rewrite a
+    # The facade owns all three of these: how many rows a row group gets, which
+    # version auto mode resolves to, and which CRS the output describes its
+    # geometry with. Every caller of this function -- 21 of them -- used to
+    # inherit whatever its writer defaulted to and whatever its own `geo` key
+    # happened to say, which is how `convert` came to write 122,880-row groups
+    # (#981) and how `sort`/`extract`/`partition` came to rewrite a
     # native-geo-only input as 1.1 WKB while `convert` kept it native (#600).
+    #
+    # The version and the CRS are resolved from the same `input_file` witness,
+    # together, because they are the same question asked of the same file.
+    # Answering only the version made the output native 2.0 while leaving its
+    # `geo` block with no `crs` key -- EPSG:5070 in the Parquet GEOMETRY logical
+    # type, OGC:CRS84 in the geo block, one file disagreeing with itself (#993).
     row_group_rows = resolve_row_group_rows(row_group_rows, row_group_size_mb)
     geoparquet_version = resolve_output_geoparquet_version(
         geoparquet_version,
@@ -2983,6 +3004,7 @@ def write_parquet_with_metadata(
         original_metadata=original_metadata,
         verbose=verbose,
     )
+    input_crs = resolve_input_crs(input_crs, input_file=input_file, verbose=verbose)
 
     effective_version = geoparquet_version or "1.1"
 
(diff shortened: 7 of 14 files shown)


cholmes:
## Multi-agent adversarial review — one S1 found and fixed in e4798da / 8533160

Two independent reviews (adversarial + simplicity), rebased onto current `main` first. Every finding acted on was reproduced by me before acting; every claim below is a measured A/B against a clean `origin/main` worktree.

### What held up — and it is most of the PR

**The six strict xfails are the acceptance test, and they pass for the right reason.** Markers gone, assertions byte-identical (only reason text changed), nothing deleted or loosened. The whole new test surface ported onto `origin/main` gives **28 failed, 15 passed** — every de-xfailed test fails there; the controls pass on both.

**I could not build the next "one file, two CRSs".** The body admits a first draft shipped exactly that (EPSG:5070 in the logical type, nothing in the `geo` block), so I built four inputs where the two sources *disagree on purpose* and read both from the output separately — because `source_crs_string` returns whichever answered first, which is how the first draft fooled its own tests:

| input (geo block / logical type) | `main` | this PR |
|---|---|---|
| EPSG:4326 / EPSG:5070 | no key / 5070 — **2 failures** | 5070 / 5070 — clean |
| `crs: null` / EPSG:5070 | no key / 5070 — **2 failures** | 5070 / 5070 — clean |
| EPSG:3857 / EPSG:5070 | 3857 / 5070 — **1 failure** | 3857 / 3857 — clean |
| EPSG:5070 / none | 5070 / none — **2 failures** | 5070 / 5070 — clean |

Every PR output is self-consistent; main leaves three of four disagreeing. The two-column fixture (`geometry` 5070, `centroid` 3857) confirms the witness lands on the primary column only, not a per-file stamp.

**The subprocess test is genuinely load-bearing — sabotage-proven.** Deleting the one-line `import geoarrow.pyarrow` from `_fix_vecorel_schema` fails exactly `test_geometry_metrics_keeps_the_native_type_from_a_fresh_process` (`KeyError: 'geometry'`) while the in-process sibling over the *same command* stays green. The in-process oracle is structurally blind to a process-global registration; the subprocess is the only thing between this defect and the user. The body's central claim is verified, not asserted. A sibling sweep of every plain-pyarrow read/write in the tree found the only other whole-file round-trip (`common._rewrite_file_with_geo_metadata`) already carries the import.

**Explicit `input_crs` wins in both directions** (`reproject -d EPSG:3857` → 3857; `-d EPSG:4326` → spec default), and every `input_crs=` in the tree is a transform target — no stale-forward path. **The "quieter loss" is repaired at every version**: `--geoparquet-version 1.0`/`1.1`/`2.0`/`parquet-geo-only` all declare EPSG:5070 where the format allows. **Row groups, `bbox`, `geometry_types`, `covering` and the column set are byte-identical** to main on every command — the #990 file-size fear does not apply.

The `projected_conus` fixture is built, deterministic, and scores zero `check spec` failures — which is what makes "the output has no failures" an assertion rather than a fixture statement. The ~100 deleted lines are genuinely gone, and the GEOGRAPHY edge-type round-trip they hand-rolled is delivered by pyarrow once extension types register (`('Geography', None)` before and after — verified directly).

### S1 — fixed: a write with a CRS to declare and no geometry to declare it for

```
$ gpio extract geoparquet tests/data/fields_pgo_5070_snappy.parquet out.parquet --exclude-cols geometry
before:  ValueError: geometry_column is required when input_crs is specified   rc=1
main:    Extracted 100 rows (out of 100 total)                                 rc=0
after:   Extracted 100 rows, 7 attribute columns, no geo key                   rc=0
```

`resolve_input_crs` answered "which CRS does the output declare" for an output with no geometry to declare it for, and `_wrap_query_with_crs` raised. Trigger is all three of: 2.0/native output (so `_plain_copy_to` is taken), non-default input CRS, and a projection that drops geometry — which is why the suite was green. This is the #945 shape exactly: remove a downgrade, introduce a crash on the adjacent input. The gate now lives in the resolver, so the one place that owns the decision is the one place that knows the answer can be absent. Three tests, written first, all failing before the fix.

### Two corrections the fixer made to *my* instructions, both right

- I asked for `timeout=30` on the subprocess. It failed that test twice under `-n auto` plus coverage — a cold interpreter paying the full CLI import measures 0.8 s idle and >30 s under 12-way load. Raised to 300 s with the measurement in a comment.
- I suggested `rindex("{")` for the JSON parse. The last `{` in a pretty-printed report is a nested object. It walks each `{` with `raw_decode` instead.

It also found the PR had introduced the nonexistent `--to-crs` flag into `common.py`, not just the body (the flag is `--crs`, CSV/TSV only). Fixed in both.

### Body corrections, reproduced not transcribed

**File size grows by a fixed 4–11 KB** on every affected command: `projjson(5070)` = 2,477 bytes, now written into the `geo` block, the logical type, and (for `geometry-metrics`) `ARROW:schema` too. It is a constant, not a ratio — `add bbox` 18,989 → 23,460 on a 200-row fixture reads as +23%, and is nothing on a real file. **A `crs: null` input now emits the null-CRS warning** on `gpio add *` where main was silent. "Nine call sites" is seven. All three now in the body.

### Simplicity: keep the production side

One new function, six executable lines, one keyword repeated seven times, one import. #990's facade boundary holds — `resolve_input_crs` implements no CRS semantics; what it newly owns is a three-line *precedence* rule that had no owner before. The facade docstring now names all three things a non-`None` CRS moves, including the load-bearing `ST_SetCRS` wrap. ~25 lines of test cleanup taken (dead `extra_args` parametrisation, a redundant `spec_failures` alias, two import aliases, two over-long call-site comments).

### Deferred, each with its own issue

**#1003** `add bbox` → `check --fix` drops the covering on a 2.0 file (pre-existing in `check_fixes`; this PR newly routes users into it) · **#1004** a `geo`-block/logical-type disagreement is resolved silently into both — should warn · **#1005** the partition path is exempt from the one-witness rule and the AST guard checks half of it · **#1006** two stream sites with the same latent registration dependency, plus the untested GEOGRAPHY round-trip.

### Gates

Fast suite **7700 passed, 76 skipped, 3 xfailed, 0 failed** · `diff-cover --fail-under=90` → **100% on 12 lines** · `pytest -m meta` 205 passed · docs lane 236 passed · `lint-imports` 4 kept / 0 broken · `ruff` clean · pre-commit and pre-push clean · `cz check` accepts the title with `!` intact.

PR #1002, opened 2026-09-11
chore(core): remove unused calculate_row_group_size

oakhill87:
## My comments
Came across this piece of unused code while investigating a fever dream of auto-computing optimal row group sizes per file, and my mother said that cleaning up was often quite as productive as adding something new.

## Summary
- Remove `calculate_row_group_size` from `core/common.py`. No
  production callers, only its own unit test.
- Remove the matching test class from `tests/test_common_utils.py`.

## Why
gpio converts a `--row-group-size-mb` target to a row count in one
place: `write_strategies/row_group_sizing.py` calls
`common._estimate_row_size`, and both DuckDB write strategies and the
plain Arrow write path use it. `calculate_row_group_size` did the same
conversion differently - bytes-per-row from the existing file's
on-disk compressed size rather than the in-memory table - and nothing
called it. Its output does not match what any write path produces.

No behavior change: no write path called this function.

## Test plan
- `uv run pytest tests/test_common_utils.py` - 135 passed
- `uv run pytest -k "row_group or common_utils or disk_rewrite"` - 382
  passed, 1 skipped
- Full fast suite (`-m "not slow and not network and not meta"`) -
  7661 passed, 76 skipped, 7 xfailed
- `grep -rn calculate_row_group_size .` - no references
- `vulture` - nothing flagged

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed the row-group size calculation utility and its related behavior.

### DIFF
--- a/geoparquet_io/core/common.py
+++ b/geoparquet_io/core/common.py
@@ -379,43 +379,6 @@ def parse_size_string(size_str):
     return int(value * multiplier)
 
 
-def calculate_row_group_size(
-    total_rows, file_size_bytes, target_row_group_size_mb=None, target_row_group_rows=None
-):
-    """
-    Calculate optimal row group size for parquet file.
-
-    Args:
-        total_rows: Total number of rows in the file
-        file_size_bytes: Current file size in bytes
-        target_row_group_size_mb: Target size per row group in MB
-        target_row_group_rows: Exact number of rows per row group
-
-    Returns:
-        int: Number of rows per row group
-    """
-    if target_row_group_rows:
-        # Use exact row count if specified
-        return min(target_row_group_rows, total_rows)
-
-    if not target_row_group_size_mb:
-        target_row_group_size_mb = 130  # Default 130MB
-
-    # Convert target size to bytes
-    target_bytes = target_row_group_size_mb * 1024 * 1024
-
-    # Calculate average bytes per row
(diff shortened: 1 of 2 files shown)


cholmes:
Reviewed by a Claude Code agent at the maintainer's request, and verified rather than taken on trust — thank you for the cleanup, and welcome.

**The dead-code claim is exactly right.** On `main`, `calculate_row_group_size` appears in precisely two places: its own definition (`core/common.py:383`) and the test class that exercised it. Not in `geoparquet_io/__init__.py`, not in `api/`, not in `docs/` or any notebook, so nothing public loses a name. On this branch there are zero survivors.

**The live path the body names is still live.** `_estimate_row_size` is called from `core/common.py:1885` and `write_strategies/row_group_sizing.py:168` — so the one conversion gpio actually uses is untouched, and the body's account of *why* this one was redundant (bytes-per-row from on-disk compressed size vs. the in-memory table; nothing called it; its answer matched no write path) is accurate.

Gates on the branch: fast suite **7661 passed, 0 failed** (the 39-test gap to `main` is your 4 deleted tests plus the branch predating #997), `lint-imports` 4 kept / 0 broken, `vulture` and `deptry` both pass, `ruff` clean, all 17 CI checks green.

Merging this ahead of a larger tooling change that is about to land so you don't have to rebase over it. Your mother was right.

PR #1029, opened 2026-09-12
fix(check): stage an in-place row-group or bbox-removal fix beside the file

cholmes:
## What changed

`gpio check row-group --fix` and `gpio check bbox --fix` (the bbox-*removal* branch, for 2.0 / parquet-geo-only files), run **in place**, asked DuckDB to write over the file it was reading. `fix_row_groups` and `fix_bbox_removal` passed `output_file` — the input path — straight to `write_parquet_with_metadata`, so DuckDB's `COPY` found the destination already there, wrote `tmp_<basename>` beside it, and renamed that over the original while a reader was still open on it. POSIX permits that; Windows does not:

```
✓ Created backup: C:\...\pgo_tiny_groups.parquet.bak
Error: IO Error: Could not move file: Access is denied.
```

Both now do what `fix_spatial_ordering` already did (#941, #959): stage a **local** in-place rewrite beside the destination (`_staging_path_beside`, same filesystem so the move stays a rename), **close the connection**, then `os.replace` it into place. On any failure — of the write or of the move — the staging file is removed and the original, never touched, is the good copy. A **remote** in-place fix (`check row-group s3://… --fix --overwrite`) is not staged: the write facade already stages a remote output locally and uploads it, and there is no destination on this machine to rename over.

**Review addendum.** The adversarial review of the first version found: (S1) the remote in-place case was routed through `_move_temp_output_into_place`, whose remote branch calls `remote_write_context(..., profile=)` — a signature it does not have — so `check row-group s3://… --fix --overwrite` went from working on `main` to a raw `TypeError` with nothing uploaded; the remote case is now not staged at all. (S2) only the move sat inside the `try`, so a writer that died after creating the staging file left a dot-prefixed orphan `ls` cannot see; the write and the move now share one `try`/`finally`, matching `fix_spatial_ordering`, and a test kills the writer mid-way and asserts the directory is byte-for-byte what it was. (S3) the first version's comment said a failed move "keeps the rewrite for the user to recover" — it never did, and should not; corrected. (S3) the `fix_bbox_removal` test case used a fixture with no `bbox` column, which only the recorder let pass; it now uses `fields_pgo_crs84_bbox_snappy.parquet`, and an end-to-end test runs both real fixers in place. (S4) DuckDB's temp is `tmp_<basename>`, not `<file>.tmp`. The reviewer also measured, with `lsof` at the moment of the move, **0 open handles** on the destination for both fixers — the window is closed, not narrowed.

Not this PR's, ticketed separately: `fix_compression` decides "in place" by raw string equality (`parquet_file == output_file`), stages in `$TMPDIR`, and unlinks-then-moves inside the `try` before `con.close()` — the exact shape #959 removed elsewhere; `check compression in.parquet --fix --fix-output ./in.parquet --no-backup` deletes the user's file outright.

## Why now

This surfaced after #1009 landed. Its tests (`test_check_fix_preserves_native_geo.py::test_check_row_group_fix_keeps_the_native_type_and_the_crs`, and `test_check_cli_guards.py::test_geo_native_file_loses_its_bbox_column`) failed on `main`'s Windows 3.13 and 3.12 legs respectively at `fc15c04`, and again on #1010's Windows 3.11 leg — while passing on #1009's own run. Intermittent, because whether the rename is refused depends on which reader the collector has released by the time DuckDB moves the file. #1009 put three more reads of the input (`get_parquet_metadata`, the version witness, the CRS witness) immediately in front of the rewrite, which is what turned a latent shape into a two-legs-in-three failure. The bug is the routing, not the reads: an in-place rewrite should never target the path it reads from, whatever is open.

## Tests

`tests/test_check_fix_in_place_staging.py` asserts the invariant, not the platform error, so it fails on every OS: with `write_parquet_with_metadata` replaced by a recorder that writes a real file where it is told, both fixers must (a) be asked to write a path other than the input, (b) beside it, (c) leave no staging file behind, (d) leave the directory with exactly the files it had, and (e) a non-in-place call still writes straight to the output path. Both in-place cases fail on `main`.

## Verification

- `uv run pytest -q -n 4 -m "not slow and not network and not meta"` — see the gate line in the last commit's CI
- `diff-cover --compare-branch=origin/main --fail-under=90`, `pytest -m meta`, `lint-imports`, `pre-commit run --all-files` and `--hook-stage pre-push` — all green locally
- Reproduced the failure path from the two `main` Windows logs (run 34684922042, jobs 103530139112 / 103530139338) and #1010's (run 34687791917, job 103537728532); it cannot be reproduced on POSIX, which is why the test pins the invariant.

Refs #1009, #941, #959.

### DIFF
--- a/geoparquet_io/core/check_fixes.py
+++ b/geoparquet_io/core/check_fixes.py
@@ -278,13 +278,14 @@ def fix_bbox_removal(parquet_file, output_file, bbox_column_name, verbose=False,
     else:
         gp_version = "1.1"  # Fallback, shouldn't happen for removal
 
-    # In place: stage beside the destination and move once the connection is
-    # closed, or DuckDB renames its `.tmp` over a file it is still reading --
-    # `Could not move file: Access is denied` on Windows. See fix_row_groups.
+    # Local in-place: stage beside the destination and move once the connection
+    # is closed, or DuckDB renames its `tmp_<basename>` over a file it is still
+    # reading -- `Could not move file: Access is denied` on Windows. See
+    # fix_row_groups for the remote case.
     actual_output = output_file
     temp_output_file = None
     moved_into_place = False
-    if is_same_file_path(parquet_file, output_file):
+    if is_same_file_path(parquet_file, output_file) and not is_remote_url(output_file):
         temp_output_file = _staging_path_beside(output_file)
         actual_output = temp_output_file
 
@@ -315,20 +316,17 @@ def fix_bbox_removal(parquet_file, output_file, bbox_column_name, verbose=False,
             # restates whatever DuckDB read and declares nothing (#1001).
             input_file=parquet_file,
         )
-    except duckdb.IOException as e:
         con.close()
+        if temp_output_file:
+            _move_temp_output_into_place(temp_output_file, output_file, profile)
+            moved_into_place = True
+    except duckdb.IOException as e:
         if is_remote_url(parquet_file):
             hints = get_remote_error_hint(str(e), parquet_file)
             raise RemoteAccessError(parquet_file, f"{hints}\n\nOriginal error: {str(e)}") from e
         raise
     finally:
         con.close()
-
-    try:
-        if temp_output_file:
-            _move_temp_output_into_place(temp_output_file, output_file, profile)
-            moved_into_place = True
-    finally:
         if temp_output_file and not moved_into_place and os.path.exists(temp_output_file):
             os.remove(temp_output_file)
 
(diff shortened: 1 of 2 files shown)


cholmes:
Superseded by #1033, which covers the same in-place `check --fix` rename bug across all three fixers (`fix_compression` included) with a shared staging helper. Two findings from #1029's adversarial review to hold #1033 to: a remote in-place fix (`check row-group s3://… --fix --overwrite`) must not be staged locally, and a staging file must be removed when the *write* fails, not only when the move does.

PR #613, opened 2026-07-28
feat(process): add --metric-nodata sentinel handling to process aggregate (#566)

cholmes:
## Summary

Implements #566 (variant A from the issue): `--metric-nodata` on all three `gpio process aggregate` subcommands maps NoData sentinel values (e.g. `-999` in GlobalBuildingAtlas heights) to `NULL` for metric computation only. `sum`/`avg`/`min`/`max` then ignore the sentinels, while `count` still counts every feature — no pre-cleaning rewrite of the input needed.

```bash
gpio process aggregate a5 buildings.parquet cells.parquet --auto \
    --metric "avg:height,max:height" --metric-nodata "-999"
```

## Implementation

- `parse_metric_nodata()` in `core/process/aggregate/common.py` validates comma-separated numeric sentinels (tokens preserved verbatim in SQL, so integer sentinels stay integer literals)
- `build_metric_select()` wraps each metric column: `NULLIF("col", -999)` for a single sentinel, `CASE WHEN "col" IN (...) THEN NULL ELSE "col" END` for multiple — the admin path shares this builder, so all three subcommands behave identically
- Errors clearly if `--metric-nodata` is given without `--metric`
- Threaded through core, CLI (`grid_aggregate_options` + admin command), and Python API (`ops.aggregate_*`, `Table.aggregate_*` as `metric_nodata=`)

Per the issue, per-column sentinels (variant B) and `--metric-where` are left for later if demand appears.

## Tests

15 new tests in `tests/test_process_aggregate_metric_nodata.py`: parsing (valid/invalid/whitespace), exact SQL-string assertions matching the existing `test_process_aggregate_common.py` style, a live DuckDB verification of the generated SQL, requires-metric validation (core + CLI), CLI help presence on all three subcommands, and slow/network e2e (core, CLI, Table API). All pass locally; fast suite green (3526 passed).

## Docs

- `docs/guide/process-aggregate.md`: NoData note in "Which metric function to use" (as requested in the issue) + Common Options
- `docs/api/python-api.md` signatures + parameter row
- CHANGELOG entry

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added `--metric-nodata` support to A5, H3, and administrative aggregation commands, including the Python aggregation APIs.
[...]

PR #1104, opened 2026-09-16
fix(aggregate): apply --metric-nodata when the metric names a column in another case

cholmes:
Split out of #1101, where it surfaced. Pre-existing on `main`; independent of
that feature.

## The bug

`validate_agg_columns` deliberately accepts a metric that names a column in a
different case — its docstring says so: *"Names are matched the way DuckDB
matches them, so `sum:HEIGHT` is a valid request against a `Height` column."*

But `DESCRIBE` reports a column's **physical** name:

```
DESCRIBE SELECT "HEIGHT" FROM t   ->  [('height', 'FLOAT', ...)]
```

So `resolve_metric_column_types` keys its result by the physical name while the
sentinel lookup used the spelling the user asked for. It found nothing, and the
two things that lookup guards were both silently skipped:

- the REAL cast (#613) — so the sentinel literal stays DOUBLE
- the up-front non-numeric rejection — so a VARCHAR metric fails mid-query
  instead of with a clear message

## Why it matters

Skipping the cast does not merely lose precision — it leaves the sentinel
**aggregated as data**, because a bare DOUBLE literal never equals the widened
float32 value. That is the exact failure #613 fixed, still reachable through a
spelling the tool tells you is supported:

```
--metric "sum:HEIGHT" --metric-nodata "-3.4028235e+38"
```

The new end-to-end test fails on `main` with:

```
assert -3.4028234663852886e+38 == 6.0
```

A -3.4e+38 where a real total belongs, with no error and no warning.

## The fix

`_resolved_column_type` looks the type up the same way `_has_column` matches
the name: exact hit first, then the existing ASCII `_fold` comparison. One call
site changes.

## Tests

Three, written first and watched fail on `main`:

- `test_build_metric_select_casts_a_real_column_named_in_another_case` — the cast is emitted
- `test_build_metric_select_rejects_a_non_numeric_column_named_in_another_case` — the rejection still fires
- `test_real_column_sentinel_matches_when_the_metric_differs_in_case` — end-to-end through `build_grid_query`, the sentinel is dropped rather than summed

Full aggregate suite (250) and all pre-commit hooks green.

## Review pass (six-agent adversarial review, 2026-09-16)

Reports: adversarial-reproduction, kieran-python-reviewer, security-sentinel, performance-oracle, architecture-strategist, code-simplicity-reviewer. The bug and the fix were reproduced independently on `main` (`sum_HEIGHT = -3.40e+38`) and on the branch (`6.0`); a 1M-row CLI run confirmed `main` summed the sentinel into a `-3.40e+43` total.

**Fixed (one thing, four reviewers):** the fix was at the wrong layer. `resolve_metric_column_types` already DESCRIBEs the metrics' own spellings, one row per column in SELECT order, then threw that correspondence away by keying on DESCRIBE's physical name; the PR then re-derived the match with a second ASCII fold scan (`_resolved_column_type`). The resolver now keys by the requested spelling (`zip(columns, rows, strict=True)`), `build_metric_select` keeps its plain `.get()`, and the helper is gone. Same behaviour, no second model of DuckDB's binding rule in gpio, and no fold scan per metric per SELECT (the reviewer measured the PR's version at 22 ms on a pathological dict). The three new tests fold into the existing ones: the keying is pinned on the resolver, and the end-to-end sentinel test is parametrized over both spellings.

**Checked and clean:** the physical name never reaches SQL (only the user's spelling does, through `quote_identifier`); case-twin columns are renamed by DuckDB before this code runs; non-ASCII case (`HÖHE`/`Höhe`) is rejected by both `_fold` and DuckDB alike; the emitted REAL cast does not change the plan (on `main` DuckDB had constant-folded the DOUBLE comparison to `false`, which *was* the bug).

**Noted, not changed here:** a REAL sentinel outside float32 range (`1e39`) now correctly reaches the cast and fails mid-query as a ConversionException rather than up front; `--bucket-point`, `--bbox-column` and the overview `--cell-column` match column names exactly while the metric rule folds case; both are pre-existing and belong with a single column-name resolver (#1109).

**Verification:** fast suite 8378 passed (one xdist flake on the first run, clean on the second); diff-cover 100%; meta 223 passed; import-linter 5 kept; pre-commit and pre-push clean.

## Summary by CodeRabbit

- **Bug Fixes**
  - Metric column names are now matched without regard to letter case.
  - REAL no-data sentinel values are correctly handled when column casing differs.
  - Non-numeric metric columns are consistently rejected, regardless of casing.
  - Rows containing REAL sentinel values are correctly excluded from grid results.

- **Tests**
  - Added coverage for case-insensitive metric matching, validation, and no-data filtering.


cholmes:
#1101 is stacked on this branch and will retarget to `main` once this merges.

Issue #1103, opened 2026-09-16
pmtiles pyramid: let the level-to-zoom band plan be specified explicitly


cholmes:
### What

Let `gpio pmtiles pyramid` accept an explicit level→zoom band plan, rather than
choosing bands from its own tile-size budget.

Something like:

```
--bands "8:11,7:10,6:9,5:7,4:5,2:0"
```

reading as "level 8 serves z11, level 7 serves z10, …", with the existing
budget-driven selection remaining the default when `--bands` is absent.

### Why

Aggregating change detections into H3 cells for a map whose cell size should
track zoom the way the consuming app already does — z9→res 6, z10→res 7,
z11→res 8, roughly one resolution step per zoom.

Passing `--levels 6,4,2` produced:

```
aggregate  z0-9   (res 4)
aggregate  z10    (res 6)
aggregate  z11    (res 8)
features   z12+
```

One cell size across ten zoom levels. At z9 a res-4 cell is ~1,770 km², so a
1,391 km² area of interest is four hexagons — unreadable where the app would
draw ~250.

Offering a finer ladder does not help. Re-running with `--levels 7,6,5,4,3,2`
produced **the same three bands**: levels whose coarser sibling already fits
the budget are discarded, so the extra candidates never appear.

That is the right default — it minimises bytes — but it optimises for tile size
when the consumer cares about cell size. There is no way to express the latter.

### Workaround, and why it is unsatisfying

Tile each level to its own archive with explicit `-Z`/`-z`, then `tile-join`
them. It works, but it reimplements what `pyramid` already does well
(overview rollup, feature band, metadata), and the band plan then lives in a
shell script rather than in the archive's `gpio:pyramid` metadata, where a
consumer can read it back.

### Note on verification

The band plan a run produces is only discoverable after the fact, by reading
`gpio:pyramid` from the built archive. A build whose plan silently drifts looks
identical to one that did not — same exit code, same file, cells quietly the
wrong size at a given zoom. `--bands` would also make that assertable up front.

Happy to send a PR if the flag shape looks right.

cholmes:
Already implemented — `--bands` landed in #1096 (`c66ee90`, 2026-09-14). The
flag shape is the one sketched here, with the pairs in ascending zoom order:

```bash
gpio pmtiles pyramid cells.parquet out.pmtiles --bands "2:0,4:5,5:7,6:9,7:10,8:11"
```

[...]

PR #1106, opened 2026-09-16
feat(overview): let gpio process overview take the same --bands plan as pyramid

cholmes:
Follow-up from #1103. Stacked on #1105.

`gpio pmtiles pyramid` has taken an explicit `level:minzoom` band plan since
#1096. `gpio process overview` — which shares `levels.py` with it — took only
`--levels`, so driving both from one ladder meant hand-translating the plan:
drop the zooms, drop the base level, hope you got it right. That is the step
that fails quietly, because a missing overview level surfaces only as cells of
the wrong size at some zoom in the finished archive.

The same string now works on both:

```bash
gpio process overview cells.parquet --bands 5:0,8:6,10:9
gpio pmtiles pyramid  cells.parquet out.pmtiles --bands 5:0,8:6,10:9
```

Verified end-to-end on a res-8 H3 aggregate with the ladder from #1103,
`2:0,4:5,5:7,6:9,7:10,8:11`:

```
Wrote level 2 overview (1 rows)   -> cells_r2.parquet
Wrote level 4 overview (5 rows)   -> cells_r4.parquet
Wrote level 5 overview (14 rows)  -> cells_r5.parquet
Wrote level 6 overview (58 rows)  -> cells_r6.parquet
Wrote level 7 overview (326 rows) -> cells_r7.parquet
```

— exactly the five the pyramid builds for that plan, base 8 excluded.

## Semantics

An overview file has no zoom, so the plan is read for the **levels it names**.
The base level drops out: it is the input, not something to roll up. The
remainder goes through `parse_levels`, so a plan naming a level that is not
coarser than the base is rejected exactly as an explicit `--levels` would be,
and an admin plan resolves to `country`.

`levels_from_bands` is a pure function over `(spec, AggregateInfo)`, so the
translation is unit-testable without a database — which is where most of the
tests sit.

## Guards, checked against the real CLI

| invocation | result |
|---|---|
| `--bands 8:0` (base only) | `Band plan names no level coarser than the base (8); no overview levels to build` |
| `--bands … --levels 4` | `--bands already names every level to build; drop --levels or drop --bands.` |
| `--bands … --bytes-per-cell 50` | `Ignoring --bytes-per-cell: --bands skips the tile-size probe it feeds.` |
| `--bands 8:6,4:0` (reversed) | the #1105 hint, naming `4:0,8:6` back |

## Scope

`levels_from_bands` + the `create_overviews` wiring, the CLI option,
`ops.create_overviews`, the overview guide (the "Stating the bands instead"
section said *pyramid only*, which is no longer true) and the Python API
reference. Regenerated `tests/data/cli_surface.json`.

Nine tests, written first and watched fail. Full fast suite, docs lane and all
pre-commit/pre-push hooks green.


cholmes:
**CI note:** this PR's own check list shows only CodeRabbit and pr-title, because `Tests` is `pull_request: branches: [main]` and this PR is stacked on `fix/bands-reversed-hint`. The matrix does not run on a PR based off a non-main branch.

Ran it out-of-band via `workflow_dispatch` on this branch instead — [run 35086031529](https://github.com/geoparquet/geoparquet-io/actions/runs/35086031529), **18/18 green**, and because dispatch is not gated the way a PR is, it covered *more* than a normal PR run:

- the full 10-leg `test` matrix (ubuntu/macos/windows x 3.10–3.13)
- `slow-tests` on all three platforms, and `network-tests` — both of which a PR run skips
- `lint`, `security`, `notebooks`, `docs-examples`

Retargeting this to `main` once #1105 merges will run the normal PR matrix too.

Issue #1110, opened 2026-09-16
feat(aggregate): a gpio:aggregate manifest so overviews stop parsing column names (unlocks avg breakdown metrics)


cholmes:
From the #1101 review (architecture pass); the design is recorded in ADR-0009.

`overview/detect.py` recognises roll-up behaviour by column-name prefix and drops what it does not recognise. #1101 named its pivots `<func>_<col>_<value>` to fit that, which works for `sum`/`min`/`max` and is not invertible (underscores in both `<col>` and `<value>`), so two things cannot be expressed:

- `--breakdown-metric avg:<col>`: a per-category mean rolls up only weighted by the *category's* count, and a metric pivot replaces the `count_<value>` columns. The pairing `avg_<col>_<value>` ↔ `count_<value>` cannot be derived from names, so `avg` is refused today.
- a second breakdown column: two `sum_peak_other` candidates, disambiguated to `_2` arbitrarily.

**Proposal:** `gpio process aggregate` writes a `gpio:aggregate` KV key: `{scheme, cell_column, base_level, columns: {name: {func, source, breakdown, value, weight}}}`. Non-`geo` KV metadata already survives every gpio rewrite (`write_funnels.extract_preserved_kv_metadata`). `detect_aggregate_info` reads it when present and every named column exists, else falls back to the name grammar (a manifest goes stale after `gpio extract` drops a pivot, and files from older gpio or other tools have none). `rollup.py` weights `avg` columns by `weight` (default `count`; the change is `RollupColumn.weight` and `SUM(q * w) / NULLIF(SUM(w) FILTER (WHERE q IS NOT NULL), 0)`). Then `--breakdown-metric avg:<col>` emits `avg_<col>_<value>` beside `count_<value>` and rolls up exactly.

Also from that review, same area: move the recogniser next to the namer (`rollup_func_for_column(name)` in `aggregate/common.py`, called from `detect.py`) so the two cannot drift, and parametrize the round-trip test over sum/min/max × `<value>`/`_other`/`_2`.

Issue #1053, opened 2026-09-13
extract arcgis: validate maxRecordCount and count before deriving page sizes


cholmes:
## What

`core/arcgis.py` trusts three numbers the server sends without checking their type or sign:

- `maxRecordCount` — `get_layer_info` reads `data.get("maxRecordCount", 1000)` (`arcgis.py:413`) and `fetch_all_features` derives the page size from it (`arcgis.py:598`). A layer that advertises `0`, `-1`, `null` (present but null, so the default does not apply) or a string produces either a `resultRecordCount=0` request loop, a `TypeError` deep in the paging arithmetic, or a division by zero in the worker-count math, none of which names the field.
- `count` from `returnCountOnly=true` — `get_feature_count` returns it as-is; a negative or non-integer count clamps `max_features` wrongly and can make the parallel path compute zero windows.
- `resultRecordCount`/offset arithmetic in the batch ladder (`_get_reduced_batch_size`) assumes an `int`.

## What to do

Validate at the boundary in `get_layer_info`/`get_feature_count`: coerce with `int()`, reject `<= 0` for `maxRecordCount` (fall back to `DEFAULT_PAGE_SIZE` with a `warn`) and `< 0` for `count` (raise `RemoteAccessError` naming the field and the value). Add cases to `tests/test_arcgis_transport.py` using `_layer_json(max_record_count=...)` — the offline harness makes these a one-line reply each.

## Refs

Found in the six-lens review of #1042 (security and adversarial lenses); not a regression, present since the extractor was written.

Issue #1035, opened 2026-09-12
fix(check): --fix turns a valid 1.0 file into one that fails gpio check spec


cholmes:
Found by WP-1 of #1018 (every `--fix` output judged by `gpio check spec`).

## What happens

A GeoParquet **1.0** file with a bbox column whose struct fields are in Overture's
`xmin, xmax, ymin, ymax` order is **valid**: 1.0 has no `covering` key, so nothing
points at the column and `gpio check spec` passes it.

Run any of the three rewrite fixes over it — `check compression --fix`,
`check row-group --fix`, `check spatial --fix` — and the output **fails gpio's own
`check spec`**:

```
covering_bbox_structure_geometry: bbox column fields must be
['xmin', 'ymin', 'xmax', 'ymax'] or ['xmin', 'ymin', 'zmin', 'xmax', 'ymax', 'zmax'],
in that order (found: ['xmin', 'xmax', 'ymin', 'ymax'])
```

Measured on `origin/main` @ 42530b4, all three fixes, same failure:

```
  compression: exit=0 ran=True in_failed=[]
     out gv=1.1.0 failed=["covering_bbox_structure_geometry: ..."]
  row-group:   exit=0 ran=True in_failed=[]
     out gv=1.1.0 failed=["covering_bbox_structure_geometry: ..."]
  spatial:     exit=0 ran=True in_failed=[]
     out gv=1.1.0 failed=["covering_bbox_structure_geometry: ..."]
```

`tests/data/unsorted.parquet` reproduces it as shipped — it is 1.0.0 with
`struct<xmin: float, xmax: float, ymin: float, ymax: float>` — and so do the four
`tests/data/country_partition/*.parquet` fixtures (those already fail at 1.1, so
they only *keep* the failure).

## Why

Two steps compound.

1. **The rewrite fixes silently upgrade 1.0 → 1.1.** They call
   `write_parquet_with_metadata(..., geoparquet_version=None)`, so the facade picks
   its own default rather than the input's version. `check all --fix` does not do
   this: `get_geoparquet_version_from_check_results` names `"1.0"` explicitly, and
   that output stays 1.0 and stays clean. So the same file repaired by two spellings
   of the same tool comes out at two different versions.
2. **Having reached 1.1, the write declares a `covering` over the bbox column
   without checking whether that column can legally be one.** `covering` is a 1.1+
   key, so the very act of upgrading turns a column nobody was pointing at into a
[...]

PR #1112, opened 2026-09-16
chore(docs): drop the three ADRs this session added; the decisions live with the code

cholmes:
Chris asked for the ADRs I added in review passes to go. ADR-0006 (covering gated on struct shape, #1070), 0007 (console script leaves via `os._exit`, #1073) and 0009 (aggregate output names are the rollup contract, #1101) are deleted; ADR-0008 from the other session (coverage floor) is renumbered 0006 so `tests/test_adr.py`'s sequence check holds, with CLAUDE.md and the `pyproject.toml` comment following.

Nothing recorded there is lost. Each decision now sits with the code that owns it:

- `geo_metadata.bbox_column_to_declare` docstring: the three covering rules, and the two rejected repairs from #1035 (keep the input's version; reorder the struct) with the measured reason.
- `cli.main.main` docstring: the alternatives ruled out for #1053 (close connections, import order, skip atexit too), the first version's embedder bug, and the retirement condition.
- `aggregate.common.breakdown_prefix` docstring: the naming grammar, its non-invertibility, collision-by-suffix, why `avg` is refused, and the manifest follow-up #1110; `overview/detect.py` points at it.

Two dangling `ADR-0006`/`ADR-0007` references in `parquet_writer.py` and `test_cli_process_exit.py` now point at the functions instead.

Verification: pre-commit and pre-push clean, meta 223 passed (ADR sequence and index tests included), the three affected suites 125 passed.

## Summary by CodeRabbit

- **Documentation**
  - Updated coverage guidance to reference the correct policy record.
  - Expanded explanations of command-line exit behavior, bounding-box covering rules, aggregate column naming, and overview rollups.
  - Corrected and consolidated the architecture decision record index.
  - Removed outdated records covering bounding-box gating, console exit behavior, and aggregate output naming.
  - Clarified that these updates document existing behavior without changing runtime functionality.


cholmes:
I'm removing these ones that my claude started putting in. We might just remove the ADR's entirely, it doesn't seem like they're valuable, or that we're using them at all here.

PR #1128, opened 2026-09-19
feat(convert): --encoding for sources GDAL cannot decode and --force-2d to drop Z/M

cayetanobv:
## Why

Two national open datasets hit `gpio convert geoparquet` on the same day (2026-09-19, mirroring for Portolan catalogs) and neither had a knob:

- **Encoding.** The Estonian Land Board's EHAK shapefiles carry Windows-ANSI DBFs (LDID `0x03`) and no `.cpg`. The GDAL behind DuckDB's spatial extension hands the bytes through unchanged, DuckDB rejects `"Lääne maakond"` as invalid UTF-8, and the conversion fails with no way to say what the file actually is. Verified fix: `ST_Read(..., open_options := ['ENCODING=ISO-8859-1'])` decodes it (`CP1252` does not, that build lacks the codec).
- **Z/M.** Every layer of the Estonian Topographic Database is 3D (`POINT Z`, `LINESTRING Z`, `POLYGON Z`). Conversion and tiling succeed, but downstream rendering of the tiles produced empty images; there was no way to ask for 2D.
- **Compound CRS** (found once `--force-2d` existed). The same shapefiles declare a compound CRS, "EST97 + EVRF2007 height" (EPSG:3301 + EPSG:5621). A compound CRS has no authority id of its own, so `crs_string_for_transform` returned `None` for it and `gpio pmtiles create` handed metre coordinates to Tippecanoe as lon/lat: every tile collapsed at the pole, with no error anywhere. Dropping Z while keeping a "horizontal + height" CRS is also simply wrong metadata.

## What

- `--encoding TEXT` on `gpio convert geoparquet` (and `encoding=` on `convert_to_geoparquet` / `read_spatial_to_arrow`): passed as GDAL open option `ENCODING` into every `ST_Read` the conversion issues, including geometry-column detection and the `keep_wkb` linearize path. Refused for Parquet input. Open options are validated as `KEY=VALUE` before touching SQL.
- `--force-2d` (and `force_2d=`): wraps the read expression in `ST_Force2D`, so bounds, bbox, Hilbert ordering and the write all see the same 2D geometry. On Parquet input it flattens WKB blobs (1.x) and native `GEOMETRY` columns (2.0) alike. With `force_2d`, a PROJJSON `CompoundCRS` with exactly one horizontal component is reduced to that component (`crs_utils.horizontal_crs`), on both the Arrow and the streaming path.
- `_extract_crs_identifier` falls back to the horizontal component of a `CompoundCRS`, so `crs_string_for_transform` (reproject, pmtiles, admin joins) identifies compound-CRS data by its XY part, with or without `--force-2d`.
- Streaming CLI path carries both options. CLI surface snapshot regenerated.

## Tests

`tests/test_convert_encoding_force2d.py` (25 tests): a hand-built Windows-ANSI DBF fixture over `buildings_test.shp`, a 3D GeoJSON fixture, a compound-CRS 3D GeoPackage written by DuckDB (`SRS 'EPSG:3301+5621'`), API and CLI paths, Parquet 1.1 and 2.0 inputs, option validation, the CRS helper and the identifier fallback. `tests/test_crs_utils.py`, `test_convert.py`, `test_pmtiles.py` green (192 passed); ruff clean.

## Open questions for the team

- Should `--encoding` also apply to `gpio convert` sibling commands (geopackage, flatgeobuf) that read through `ST_Read`?
- Should the DBF language-driver byte be honoured automatically (LDID → encoding) instead of, or in addition to, the explicit option? The explicit option is build-independent, which is why it comes first.
- Is `--force-2d` the right name, or should it mirror GDAL's `-dim XY`?
- Should `pmtiles create` (and `reproject`) refuse, rather than silently pass through, a CRS they cannot identify? The pole-collapse failure mode above was silent end to end.

Draft for discussion next week; opened from the Portolan mirror work where all three fixes are needed now.


cayetanobv:
Split into one PR per issue as requested: https://github.com/geoparquet/geoparquet-io/pull/1129 (--encoding), https://github.com/geoparquet/geoparquet-io/pull/1130 (--force-2d, stacked), https://github.com/geoparquet/geoparquet-io/pull/1131 (compound CRS, stacked). Closing this one.

PR #1129, opened 2026-09-19
feat(convert): --encoding for sources whose text encoding GDAL cannot detect

cayetanobv:
## Why

The Estonian Land Board's EHAK shapefiles (2026-09-19, mirrored for a Portolan catalog) carry Windows-ANSI DBFs (LDID `0x03`) and no `.cpg`. The GDAL behind DuckDB's spatial extension hands the bytes through unchanged, DuckDB rejects `"Lääne maakond"` as invalid UTF-8, and `gpio convert geoparquet` fails with no way to say what the file actually is. Verified fix: `ST_Read(..., open_options := ['ENCODING=ISO-8859-1'])` decodes it (`CP1252` does not, that build lacks the codec).

## What

- `--encoding TEXT` on `gpio convert geoparquet` (and `encoding=` on `convert_to_geoparquet` / `read_spatial_to_arrow`): passed as GDAL open option `ENCODING` into every `ST_Read` the conversion issues, including geometry-column detection and the `keep_wkb` linearize path. Refused for Parquet input. Open options are validated as `KEY=VALUE` before touching SQL.
- Streaming CLI path carries it. CLI surface snapshot regenerated.

## Tests

`tests/test_convert_encoding.py` (8 tests): a hand-built Windows-ANSI DBF fixture over `buildings_test.shp`, API and CLI paths, option validation, Parquet refusal.

## Open questions

- Should `--encoding` also apply to `gpio convert` sibling commands (geopackage, flatgeobuf) that read through `ST_Read`?
- Should the DBF language-driver byte be honoured automatically (LDID → encoding) instead of, or in addition to, the explicit option? The explicit option is build-independent, which is why it comes first.

One of three PRs split from #1128 (one issue each): this one, `--force-2d` (stacked on it), and compound-CRS handling (stacked on that).


geoparquet_io/core/convert.py
```diff
@@ -1296,8 +1345,10 @@ def _convert_spatial_path(
             if verbose:
                 debug("Curved geometries detected; linearizing via keep_wkb read")
             table_expr = _register_linearized_view(
-                con, input_file, layer, geom_column, max_angle_deg
+                con, input_file, layer, geom_column, max_angle_deg, open_options=open_options
             )
+        elif open_options:
```
cayetanobv, line 1462: **Medium — setting `table_expr` here disables the curved-geometry bounds fallback.**

`_bounds_with_curve_fallback` infers "already linearized" from the value: `already_linearized = table_expr is not None` (line 1261). Before this PR `table_expr` was non-`None` only when the linearized view had been registered; now any `--encoding` run sets it to a plain `ST_Read`, so the inference is wrong.

Scenario: `gpio convert geoparquet curved.gdb out.parquet --encoding ISO-8859-1` (or a GeoPackage on S3 — the pre-scan at line 1337 only covers *local* `.gpkg`). The bounds pass raises `Unsupported geometry type in WKB`, line 1262 re-raises instead of linearizing, and the error escapes `_convert_spatial_path`. It is only caught by the whole-conversion `force_linearize` retry (#985) — a second full read of the source plus a `Geometry DuckDB cannot read directly; linearizing the source and converting again` warning that the same file would not have produced without `--encoding`.

Related: if that fallback *does* fire, line 1268 calls `_register_linearized_view(con, input_file, layer, geom_column, max_angle_deg)` with no `open_options`, so the re-read silently loses the ENCODING.

Fix: thread `open_options` into `_bounds_with_curve_fallback` and pass an explicit `already_linearized` flag rather than inferring it from `table_expr`.

geoparquet_io/core/convert.py
```diff
@@ -1675,8 +1736,15 @@ def _read_spatial_to_arrow(
...
-        table_expr = _build_st_read_expr(input_file, layer)
+            return _read_spatial_linearized(
+                con,
+                input_file,
+                layer,
+                geom_column,
+                max_angle_deg,
+                open_options=open_options,
+            )
+        table_expr = _build_st_read_expr(input_file, layer, open_options=open_options)
```
cayetanobv, line 1879: **Medium — the error-triggered linearize fallback in this same function drops `open_options`.**

Line 1771 still reads `return _read_spatial_linearized(con, input_file, layer, geom_column, max_angle_deg)` — the two sibling call sites (1740 and this line) were updated, this one was not.

Scenario: a source that needs `ENCODING` *and* whose curves the local-GPKG pre-scan cannot see (FileGDB, a remote GeoPackage, DXF). The query above fails with the curve error, `_is_linearizable_curve_error` sends it down the `keep_wkb` path, and that read is issued **without** `ENCODING=...` — so it fails on exactly the invalid-UTF-8 this option exists to fix, or silently mojibakes the attributes.

```suggestion
        table_expr = _build_st_read_expr(input_file, layer, open_options=open_options)
```
(the real edit is at line 1771: add `open_options=open_options` to that `_read_spatial_linearized` call.)

cayetanobv (commented):
**Recommendation: changes requested (posted as a comment — own PR).** The feature works end-to-end (the 8 new tests pass here on DuckDB 1.5.5, and `--encoding` on a GeoJSON/GPKG that doesn't support the option is harmlessly ignored by GDAL), but the option leaks out of three paths and is silently a no-op for CSV. Five findings inline: two where `open_options` is dropped on a fallback read, one CSV no-op (verified), plus the validation layering and the missing `gpio.convert()` kwarg/docs.

geoparquet_io/cli/commands/convert.py
```diff
@@ -172,6 +172,12 @@ def convert(ctx):
     is_flag=True,
     help="Allow conversion to plain Parquet when no geometry column is detected (default: error)",
 )
+@click.option(
+    "--encoding",
```
cayetanobv, line 176: **Medium — `--encoding` is refused for Parquet but silently a no-op for CSV/TSV.**

`_convert_csv_path` / `_read_csv_to_arrow` never receive it, and nothing warns. Verified on this branch:

```
csv.write_bytes(b"name,wkt\n" + "L\u00e4\u00e4ne maakond,POINT(1 2)\n".encode("latin-1"))
convert_to_geoparquet(csv, out, encoding="ISO-8859-1")
# GeoParquetError: Conversion failed: Invalid Input Error: CSV Error on Line: 2
```

A user whose Latin-1 *CSV* fails on an accented value will reach for exactly this flag, per its own help text ("Source text encoding for drivers that cannot tell"), and get nothing. DuckDB's `read_csv` has its own `encoding` parameter, so either wire it through `_build_csv_read_expr` or refuse it for CSV the way Parquet is refused — silently accepting it is the worst of the three options.

geoparquet_io/core/convert.py
```diff
@@ -1261,14 +1306,18 @@ def _convert_spatial_path(
                and their metadata. Returns (None, None) if no geometry found.
     """
 
+    open_options = source_open_options(encoding)
+    if open_options and is_parquet:
+        raise InvalidParameterError("encoding", "only applies to sources GDAL reads, not Parquet")
```
cayetanobv: **Low — this validation is too deep to survive as a typed error.** `convert_to_geoparquet`'s trailing `except Exception` (line ~2346) re-wraps it, so the caller gets `GeoParquetError("Conversion failed: Invalid parameter 'encoding': ...")`, not the `InvalidParameterError` raised here — `except InvalidParameterError:` will not catch it (the new test only matches the substring `Parquet`, so it passes either way). `read_spatial_to_arrow` does the same via its `Reading failed:` wrapper.

It also fires only after the output path is validated, the DuckDB connection is opened and `Converting ...` is printed. Validating `encoding` against `is_parquet`/`is_csv` once at the top of `convert_to_geoparquet` and `read_spatial_to_arrow` keeps the type and fails before any work.

geoparquet_io/cli/commands/convert.py
```diff
@@ -172,6 +172,12 @@ def convert(ctx):
...
+    default=None,
+    help="Source text encoding for drivers that cannot tell, e.g. a shapefile DBF without "
+    ".cpg (ISO-8859-1, UTF-8, ...). Passed to GDAL as open option ENCODING. Not for Parquet.",
+)
```
cayetanobv, line 181: **Low — no Python API and no docs for the new option.** `gpio.convert()` (`geoparquet_io/api/table.py:369`) forwards a fixed kwarg list to `read_spatial_to_arrow` and has no `encoding` parameter, so the Estonian EHAK shapefile that motivated this PR still cannot be converted through the documented Python API — only through the CLI or the private core function. CLAUDE.md's checklist asks for `api/table.py` (+`api/ops.py`) and a `docs/guide/` update with CLI + Python examples; the diff touches no docs.

geoparquet_io/core/convert.py
```diff
@@ -87,8 +89,94 @@ def _validate_layer_name(layer: str) -> str:
...
+
+
+#: DuckDB's CSV reader spells its encodings its own way; keyed by the codec
+#: registry's canonical name so ``ISO-8859-1``, ``latin1`` and ``latin-1`` all
+#: reach it as the one it knows.
+_DUCKDB_CSV_ENCODINGS = {
+    "iso8859-1": "latin-1",
+    "utf-8": "utf-8",
+    "utf-16": "utf-16",
+    "cp1252": "cp1252",
```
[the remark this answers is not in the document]
cayetanobv: Confirmed offline: with autoload disabled, `encoding='cp1252'` fails with "The CSV Reader does not support the encoding". Fixed in 2d7f770: `_prepare_csv_encoding` now loads the `encodings` extension explicitly (via the shared `_install_and_load_extension`) whenever the requested encoding is not one of the reader's built-ins (UTF-8, UTF-16, Latin-1), and a failed load is reported in terms of `--encoding`. Tests cover the built-in path (no load), the CP1252 path (load called) and the failure message; docs updated.

PR #1130, opened 2026-09-19
feat(convert): --force-2d drops Z and M so 3D sources become 2D GeoParquet

cayetanobv:
## Why

Every layer of the Estonian Topographic Database (ETAK, 2026-09-19) is 3D (`POINT Z`, `LINESTRING Z`, `POLYGON Z`). Conversion and tiling succeeded, but downstream rendering of the tiles produced empty images, and there was no way to ask `gpio convert geoparquet` for 2D output.

## What

- `--force-2d` (and `force_2d=` on `convert_to_geoparquet` / `read_spatial_to_arrow`): wraps the read expression in `ST_Force2D`, so bounds, bbox, Hilbert ordering and the write all see the same 2D geometry. On Parquet input it flattens WKB blobs (GeoParquet 1.x) and native `GEOMETRY` columns (2.0) alike; the `keep_wkb` linearize path applies it after stroking.
- Streaming CLI path carries it. CLI surface snapshot regenerated.

## Tests

`tests/test_convert_force2d.py` (7 tests): a 3D GeoJSON fixture, API and CLI paths, Parquet 1.1 and 2.0 inputs.

## Open question

Is `--force-2d` the right name, or should it mirror GDAL's `-dim XY`?

Stacked on the `--encoding` PR (shared plumbing in `_read_spatial_to_arrow`); the diff shown is the force-2d part only. Split from #1128.


geoparquet_io/core/convert.py
```diff
@@ -1743,13 +1786,21 @@ def _read_spatial_to_arrow(
...
         table_expr = _build_st_read_expr(input_file, layer, open_options=open_options)
 
     # Convert geometry to WKB for geoarrow compatibility
+    geometry_expr = quoted_geom
+    if force_2d:
+        geometry_expr = (
+            f"ST_Force2D(ST_GeomFromWKB({quoted_geom}))"
+            if is_parquet
+            else f"ST_Force2D({quoted_geom})"
+        )
```
cayetanobv: **Bug (high): `force_2d` on a Parquet input always fails here.**

`ST_GeomFromWKB` only binds against `BLOB`. DuckDB 1.5 surfaces a GeoParquet geometry column as native `GEOMETRY` (it reads the Parquet GEOMETRY logical type), so this branch raises for *both* 1.1 and 2.0 files that `gpio` itself just wrote:

```
>>> read_spatial_to_arrow("z1.1.parquet", force_2d=True)
GeometryError: Invalid geometry data: Binder Error: No function matches the given name
and argument types 'st_geomfromwkb(GEOMETRY('EPSG:4979'))'
```

`_convert_spatial_path` gets this right (it `DESCRIBE`s the column first, lines 1376-1382) — this path never got the same branch, and the new tests only exercise `read_spatial_to_arrow` on GeoJSON, so nothing catches it.

```suggestion
    geometry_expr = quoted_geom
    if force_2d:
        if is_parquet:
            (column_type,) = con.execute(
                f"SELECT column_type FROM (DESCRIBE SELECT {quoted_geom} FROM {table_expr})"
            ).fetchone()
            geometry_expr = (
                f"ST_Force2D({quoted_geom})"
                if column_type.upper().startswith("GEOMETRY")
                else f"ST_Force2D(ST_GeomFromWKB({quoted_geom}))"
            )
        else:
            geometry_expr = f"ST_Force2D({quoted_geom})"
```

(verified: with this change both the 1.1 and the 2.0 file read back fine.)

geoparquet_io/core/convert.py
```diff
@@ -1743,13 +1786,21 @@ def _read_spatial_to_arrow(
...
     query = f"""
         SELECT * EXCLUDE ({quoted_geom}),
-               ST_AsWKB({quoted_geom}) AS geometry
+               ST_AsWKB({geometry_expr}) AS geometry
         FROM {table_expr}
```
cayetanobv, line 2023: **Bug (medium): the curve fallback at the bottom of this function silently drops `force_2d`.**

Line 1822:

```python
return _read_spatial_linearized(con, input_file, layer, geom_column, max_angle_deg)
```

The `strategy == "linearized"` call above was updated to forward `force_2d=force_2d`, but this one — the runtime fallback for sources with no cheap pre-scan (FileGDB, a GeoPackage on S3, per the comment right above it) — was not. A curved 3D FileGDB read with `force_2d=True` therefore comes back with Z still attached, with no error and no warning: exactly the silent-3D failure this PR exists to prevent. (It also still drops `open_options`, which is pre-existing.)

geoparquet_io/cli/commands/convert.py
```diff
@@ -178,6 +178,11 @@ def convert(ctx):
     help="Source text encoding for drivers that cannot tell, e.g. a shapefile DBF without "
     ".cpg (ISO-8859-1, UTF-8, ...). Passed to GDAL as open option ENCODING. Not for Parquet.",
 )
+@click.option(
+    "--force-2d",
+    is_flag=True,
+    help="Drop Z and M coordinates (ST_Force2D) so 3D sources become 2D GeoParquet",
+)
```
cayetanobv, line 186: **Bug (medium): `--force-2d` is a silent no-op for CSV/TSV input.**

`convert_to_geoparquet` routes CSV through `_convert_csv_path`, which takes no `force_2d` argument, so the flag is accepted and quietly ignored. Reproduced on this branch:

```
id,geom
1,"POINT Z (1 2 3)"
```

`gpio convert geoparquet p.csv out.parquet --force-2d` exits 0 and writes `"geometry_types": ["Point Z"]`.

Either thread `force_2d` into `_convert_csv_path` (wrapping the `ST_GeomFromText` / `ST_Point` expression in `ST_Force2D`), or reject the combination up front so the user is not told the Z was dropped when it was not.

geoparquet_io/core/convert.py
```diff
@@ -1350,6 +1362,31 @@ def _convert_spatial_path(
...
+            # A GeoParquet 2.0 file arrives as a native GEOMETRY column, a 1.x
+            # file as WKB blobs; keep whichever shape the column had.
+            (column_type,) = con.execute(
+                f"SELECT column_type FROM (DESCRIBE SELECT {quoted} FROM {source})"
+            ).fetchone()
+            if column_type.upper().startswith("GEOMETRY"):
+                flattened = f"ST_Force2D({quoted})"
+            else:
+                flattened = f"ST_AsWKB(ST_Force2D(ST_GeomFromWKB({quoted})))"
+            table_expr = f"(SELECT * REPLACE ({flattened} AS {quoted}) FROM {source})"
```
cayetanobv: **Bug (medium): only the *primary* geometry column is flattened; secondary geometry columns keep their Z.**

`* REPLACE (... AS {quoted})` touches `geom_info["primary"]` only, and `geom_info["secondary"]` (which this function goes on to preserve, and which the writer emits into `geo.columns`) passes through untouched. Verified on a two-geometry-column GeoParquet input:

```
input : geometry ['Point Z'], centroid ['Point Z']
--force-2d output: geometry ['Point'], centroid ['Point Z']   # centroid still ST_HasZ
```

For a source like ETAK, where *every* layer is 3D, a file with a secondary geometry column still ships 3D data after `--force-2d`. Worth applying the replacement to each of `[geom_column, *secondary_columns]` (the non-Parquet `force_2d_expr` branch below has the same single-column limitation).

geoparquet_io/core/convert.py
```diff
@@ -1350,6 +1362,31 @@ def _convert_spatial_path(
...
+        else:
+            table_expr = force_2d_expr(
+                table_expr or _build_st_read_expr(input_file, layer, open_options=open_options),
+                geom_column,
+            )
```
cayetanobv, line 1581: **Issue (low): setting `table_expr` unconditionally disables the in-place curve fallback during the bounds pass.**

`_bounds_with_curve_fallback` decides whether it may linearize with `already_linearized = table_expr is not None` (line 1271). Before this PR a plain non-Parquet source reached it with `table_expr is None`, so a curved FileGDB/remote-GPKG was linearized on the spot. With `--force-2d` the expression is always non-`None`, so the bounds error is re-raised and recovery falls to the outer `_convert_once(force_linearize=True)` retry — correct output, but the entire source is read one extra time for every curved input converted with `--force-2d`.

Passing a flag that says "this expression is *not* a linearized view" (or checking `force_linearize`/strategy instead of `table_expr is not None`) would keep the cheap path.

geoparquet_io/core/convert.py
```diff
@@ -1350,6 +1362,31 @@ def _convert_spatial_path(
         elif open_options:
             table_expr = _build_st_read_expr(input_file, layer, open_options=open_options)
 
+    if force_2d:
```
cayetanobv, line 1568: **Issue (low): the output keeps a 3-dimensional CRS after the geometry is flattened.**

`ST_Force2D` drops the coordinates but nothing revises the CRS that gets written. Parquet→Parquet with `--force-2d` on this branch:

```
geometry_types: ['Point']
crs axes:       ['Geodetic latitude', 'Geodetic longitude', 'Ellipsoidal height']   # EPSG:4979
```

So the file advertises a 3D CRS over 2D geometry. Given that the motivation for the flag is downstream renderers mis-handling 3D, it would be worth demoting the CRS to its 2D equivalent (or at least warning) when `force_2d` fires.

geoparquet_io/core/convert.py
```diff
@@ -1483,6 +1521,8 @@ def read_spatial_to_arrow(
         encoding: Source text encoding for drivers that cannot tell, e.g. a
             shapefile DBF without ``.cpg``; passed to GDAL as open option
             ``ENCODING`` (``ISO-8859-1``, ``UTF-8``, ...). Not for Parquet.
+        force_2d: Drop Z and M coordinates (``ST_Force2D``) so 3D sources
+            become 2D geometry (default: False).
```
cayetanobv, line 1722: **Nit (low): `force_2d` never reaches the documented Python API.**

`geoparquet_io.api.convert()` (`api/table.py:371`) is the public wrapper over `read_spatial_to_arrow`, and it has no `force_2d` parameter, so the new keyword is unreachable from `gpio.convert(...)`. CLAUDE.md's "every CLI command needs a Python API" applies to new options too — `api/table.py` and `api/ops.py` both need the passthrough. (The stacked `--encoding` PR has the same gap, so this may be worth fixing once for both.)

cayetanobv (commented):
Reviewed the `--force-2d` diff. The feature works for the cases the new tests cover (GeoJSON in, Parquet 1.1/2.0 in, CLI, streaming), but there are a few paths where it either errors outright or silently does nothing.

Highest-value one is the `_read_spatial_to_arrow` Parquet branch, which cannot bind at all against a DuckDB `GEOMETRY` column; after that, the three places where `--force-2d` is accepted and quietly ignored (CSV input, secondary geometry columns, the runtime curve fallback).

On the open question: `--force-2d` reads well and matches `ST_Force2D`; GDAL's `-dim XY` is the odd one out among the tools here. I'd keep the current name.

PR #1131, opened 2026-09-19
fix(convert): reduce a compound CRS to its horizontal component when --force-2d drops Z

cayetanobv:
## Why

The ETAK shapefiles declare the compound CRS "EST97 + EVRF2007 height" (EPSG:3301 + EPSG:5621). After `--force-2d` the GeoParquet still carried the `CompoundCRS`. A compound CRS has no authority id of its own, so `crs_string_for_transform` returned `None` and `gpio pmtiles create` handed metre coordinates to Tippecanoe as lon/lat: every tile collapsed at the pole (bbox `-180, 85.05, 180, 85.05`), with no error anywhere. Keeping a "horizontal + height" CRS on 2D geometry is also wrong metadata.

## What

- `crs_utils.horizontal_crs()` / `horizontal_component()`: reduce a PROJJSON `CompoundCRS` with exactly one horizontal component to that component (keeping `$schema`); anything else passes through unchanged.
- `convert` applies it on both the Arrow and the streaming path when `force_2d` is set.
- `_extract_crs_identifier` falls back to the horizontal component of a `CompoundCRS`, so `crs_string_for_transform` (reproject, pmtiles, admin joins) identifies compound-CRS data by its XY part, with or without `--force-2d`.

## Tests

`tests/test_convert_compound_crs.py` (10 tests): a compound-CRS 3D GeoPackage written by DuckDB (`SRS 'EPSG:3301+5621'`), the helper, the identifier fallback, API and CLI paths. `test_crs_utils.py` and `test_pmtiles.py` still green.

## Open question

Should `pmtiles create` (and `reproject`) refuse, rather than silently pass through, a CRS they cannot identify? The pole-collapse failure mode above was silent end to end.

Stacked on the `--force-2d` PR. Split from #1128.


geoparquet_io/core/crs_utils.py
```diff
@@ -187,6 +187,11 @@ def _extract_crs_identifier(crs_info):
                         return (authority, int(code))
                     except (ValueError, TypeError):
                         return (authority, str(code).upper())
+        # A CompoundCRS carries no id of its own; the XY part of any transform
+        # is fully described by its horizontal component, so identify by that.
+        component = horizontal_component(crs_info)
+        if component is not None:
+            return _extract_crs_identifier(component)
```
cayetanobv: **Compound CRS over EPSG:4326 now reads as the default and its `crs` is dropped on write.**

`is_default_crs()` (line 216) is built straight on `_extract_crs_identifier`, so this fallback makes `is_default_crs(CompoundCRS(4326 + 3855))` return `True`. `apply_output_crs()` then takes the second branch and `pop`s the key, and `note_default_crs_normalized()` logs "Normalized an explicit default CRS" — for a CRS that was never the default.

Reproduced on this branch vs `main`, **no `--force-2d`, Z retained in the output**:

```python
# 3D GPKG written with SRS 'EPSG:4326+3855'  (WGS 84 + EGM2008 height)
convert_to_geoparquet(gpkg, out)          # no force_2d
# pr1131 : geo['columns'][g] has no 'crs' key at all  -> file claims plain OGC:CRS84
# main   : geo['columns'][g]['crs']['type'] == 'CompoundCRS'
```

The geometry still carries Z in EGM2008 heights, but the vertical datum is gone from the metadata with no warning — the same class of silently-wrong CRS metadata this PR is fixing, reintroduced on the path that did *not* ask for it. `inspect_utils._detect_metadata_mismatches` won't flag it either, because its `is_default_crs(parquet_crs)` guard now answers `True` as well.

Note also that the two "is this the default?" predicates now disagree: `_is_crs84_equivalent()` reads `crs['id']` directly and falls back to pyproj, so it still answers `False` for the same value — exactly the drift the docstring at line 136 warns about.

Suggest moving the fallback out of `_extract_crs_identifier` and into the transform choke points that actually need it (`crs_string_for_transform`, `resolve_crs_to_string`), or, if it stays here, making `is_default_crs()` refuse to call a `CompoundCRS` the default.

geoparquet_io/core/crs_utils.py
```diff
@@ -187,6 +187,11 @@ def _extract_crs_identifier(crs_info):
                         return (authority, int(code))
                     except (ValueError, TypeError):
                         return (authority, str(code).upper())
+        # A CompoundCRS carries no id of its own; the XY part of any transform
+        # is fully described by its horizontal component, so identify by that.
+        component = horizontal_component(crs_info)
```
cayetanobv: **Downstream of this hunk — `format_writers._get_srs_parameter` (line 139) now drops the vertical CRS on GDAL export.**

With a compound-CRS GeoParquet as input:

```
pr1131 : _get_srs_parameter(p) -> 'EPSG:3301'
main   : _get_srs_parameter(p) -> '{"type": "CompoundCRS", "name": "EST97 + EVRF2007 height", ...}'
```

So `gpio convert geopackage 3d_estonia.parquet out.gpkg` (and flatgeobuf) hands GDAL only the horizontal CRS, and the EVRF2007 height declaration is lost from a file whose geometries still have Z — a GeoPackage can hold the compound CRS in `gpkg_spatial_ref_sys`, so this is a real fidelity loss on a round trip. Same for the 4326 case, which additionally short-circuits through `is_default_crs` at line 135 and yields a flat `"EPSG:4326"`.

geoparquet_io/core/crs_utils.py
```diff
@@ -187,6 +187,11 @@ def _extract_crs_identifier(crs_info):
                         return (authority, int(code))
                     except (ValueError, TypeError):
                         return (authority, str(code).upper())
+        # A CompoundCRS carries no id of its own; the XY part of any transform
```
cayetanobv: **Downstream of this hunk — `inspect` now mislabels a compound CRS.**

`inspect_utils._format_crs_for_display` (line 126) and `_crs_are_equivalent` (line 178) both go through `_extract_crs_identifier`:

```
_format_crs_for_display(CompoundCRS('EST97 + EVRF2007 height')) -> 'EPSG:3301'
_crs_are_equivalent(CompoundCRS(3301+5621), 'EPSG:3301')        -> True
```

So `gpio inspect meta` reports a 3D compound-CRS file as plain `EPSG:3301` with nothing hinting at the vertical component, and `_detect_metadata_mismatches` stops reporting a genuine difference between a 2.0 file's native geo type and its `geo` block. Scoping the fallback to the transform helpers (see the comment above) keeps the display and comparison paths honest.

cayetanobv (commented):
Recommendation: do not land as-is — the `_extract_crs_identifier` fallback is placed in the shared identifier helper, and `is_default_crs()` sits directly on top of it, so **a compound CRS whose horizontal part is EPSG:4326 (e.g. "WGS 84 + EGM2008 height", EPSG:4326+3855) now reads as the GeoParquet default and `apply_output_crs()` deletes the `crs` key entirely** — on the plain path, with no `--force-2d` and with Z still in the output. Verified end to end against this branch vs `main`.

The `--force-2d` reduction itself (`horizontal_crs`, the two `convert` call sites) looks right, and the 2.0 fast path was checked too — it also writes the reduced ProjectedCRS. Three findings below, all rooted in the same placement choice.

PR #1135, opened 2026-09-20
fix: resolve #1134 - fix(extract arcgis): JSON Error 500 page responses bypass th

### LINKED ISSUE #1134: fix(extract arcgis): JSON Error 500 page responses bypass the adaptive batch-size fallback
## Summary

`gpio extract arcgis` advertises that `--batch-size` "Auto-reduces on server errors", and the core extractor has a fallback ladder (`1000, 500, 100, 50, 10, 1`). However, that ladder is reached only when a paged request returns a non-JSON response and `http_retry.py` raises `BatchTooLargeError`.

Many ArcGIS services return HTTP 200 with a valid JSON error body when a page is too large or geometrically expensive:

```json
{
  "error": {
    "code": 500,
    "message": "Error performing query operation.",
    "details": []
  }
}
```

`fetch_features_page()` sends this through `_handle_arcgis_response()`, which raises `GeoParquetError`. `fetch_all_features()` catches only `BatchTooLargeError`, so extraction stops instead of retrying the same offset with a smaller page.

## Evidence from current code

- `geoparquet_io/core/arcgis.py:543-548`: the page request passes `batch_size=limit`, but a parsed JSON `error` is handled by `_handle_arcgis_response()`.
- `geoparquet_io/core/arcgis.py:184-196`: `_handle_arcgis_response()` raises `GeoParquetError` for every non-auth ArcGIS error.
- `geoparquet_io/core/arcgis.py:626-657` and `727-773`: sequential and parallel fallback logic responds only to `BatchTooLargeError`.
- `geoparquet_io/core/http_retry.py:149-174`: `BatchTooLargeError` is produced only when JSON parsing fails.
- `tests/test_arcgis_transport.py:651-752`: fallback tests cover HTML/non-JSON rejection, not ArcGIS's normal JSON error envelope.

## Real failures

This has blocked catalog ingestion against multiple public ArcGIS services.

### Wallonia

The server advertises `maxRecordCount=2000`, but heavy polygon layers require smaller pages. Direct probes on 2026-09-20 showed:

| Layer | 2000 | 1000 | 500 | 250 | 100 |
|---|---:|---:|---:|---:|---:|
| `EAU/RES_HYDRO_WAL/2` | JSON error 500 | works | | | |
| `EAU/PROTECT_CAPT/4` | JSON error 500 | 500 | 500 | 500 | works |
| `FAUNE_FLORE/NATURA2000/7` | JSON error 500 | 500 | 500 | works | works |
| `FAUNE_FLORE/LIMITES_DNF/0` | JSON error 500 | 500 | works | works | works |

### Norway / NVE

In `CartoDB/portolan-pipeline#61`, all four build shards failed and only 33/39 Collections completed. One exact example from run `35527650921`:

```text
skred/fjellskred-tsunami-faresoner
Fetching features 1-55 of 55...
Failed to fetch features at offset 0:
Feature query: Error 500 - Error performing query operation.
```

The layer contained only 55 features, showing that complexity rather than row count can make even a server-advertised page fail.

## Expected behavior

For a paged feature query, a retryable ArcGIS JSON error that plausibly indicates page/serialization pressure should enter the same adaptive ladder as a non-JSON oversized-page response:

1. retry the same offset at the next smaller batch size;
2. keep that reduced size for subsequent pages;
3. support both sequential and parallel paths;
4. fail normally if `batch_size=1` still fails.

Auth errors 498/499 and clearly non-size-related client/semantic errors must remain fatal. It may be safest to classify only known retryable page-query codes/messages rather than treating every ArcGIS error as batch pressure.

## Suggested regression tests

Add transport tests beside the existing HTML fallback cases:

- HTTP 200 + JSON `error.code=500` at size 100, success at 50: assert requests are `offset=0,size=100`, then `0,50`, then `50,50`.
- The equivalent parallel-window case.
- JSON error at `batch_size=1`: assert the useful terminal error.
- Auth 498/499 and a non-size-related 400 remain fatal and do not walk the ladder.

## Related API/docs mismatch

The CLI already exposes `--batch-size`, so users can manually work around a known threshold. The defect is specifically that the advertised automatic reduction does not recognize ArcGIS's standard JSON error response.

fathirramadhan-web:
Closes #1134

### Overview
Addresses the problem reported in #1134 by resolving the underlying issue and ensuring test coverage passes.

### Verification
- Tested locally
- Linters/tests pass

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved ArcGIS data retrieval when servers return JSON errors for oversized requests. The system now automatically reduces batch sizes and retries affected pages across sequential and parallel processing.
  - Preserved clear error reporting for other server errors, including HTTP 400 responses.
  - Improved handling of feature-count retrieval failures during integration workflows.


yharby (changes requested):
Thanks for picking this up. The current implementation is not safe to merge yet.

### Blocking: every ArcGIS 5xx is classified as page-size pressure

```python
if code >= 500 and limit is not None:
    raise BatchTooLargeError(...)
```

This makes unrelated service failures (backend outage, database failure, unsupported operation, timeout surfaced in the JSON envelope, etc.) walk the entire `1000 -> 500 -> 100 -> 50 -> 10 -> 1` ladder and end with the misleading message that the layer geometry is too complex. Issue #1134 explicitly asks to classify only known retryable page/serialization-pressure responses and preserve fatal handling for other errors.

Please introduce a narrow classifier based on the ArcGIS code/message/details observed for this case, and add a negative regression test for a non-size-related 5xx JSON error. Also preserve immediate auth handling for 498/499.

### Out of scope

Please remove:

- the generated `.gitignore` additions;
- the Mozambique integration fixture change. Catching every `GeoParquetError` and skipping can hide authentication, malformed-response, query, and product regressions. It is unrelated to #1134.

### Tests

The sequential/parallel fallback tests are useful, but should be accompanied by:

- a non-size-related JSON 5xx that remains fatal without walking the ladder;
- 498/499 remaining fatal;
- the existing size-pressure JSON 500 case entering the ladder.

CI is also currently blocked by the lint job.

fathirramadhan-web:
Updated as requested. Thanks for the feedback @codecov-commenter!

fathirramadhan-web:
Updated as requested. Thanks for the feedback @yharby!

yharby:
Solved #1140

Issue #382, opened 2026-04-13
extract arcgis: JSONDecodeError when server returns non-JSON for large batch — no batch-size fallback or --batch-size flag


maxmalynowsky:
## Description

`gpio extract arcgis` fails with a `JSONDecodeError` on some ArcGIS FeatureServer layers. The server returns a non-JSON response (likely an HTML error page) when asked to serve all features in a single batch, even when the batch size is within the advertised `maxRecordCount`. The same layer loads successfully in QGIS, suggesting the issue is the page size used per request.

The error is not caught — there is no adaptive batch-size retry, and there is no `--batch-size` flag (unlike `extract wfs`, which has `--page-size`).

## Reproduction

**Works** (Admin 2):
```
gpio extract arcgis \
  https://www.mozgis.gov.mz/server/rest/services/Data/UnidadesAdministrativasCenso2017/FeatureServer/2 \
  moz_admin2.parquet
```

**Fails** (Admin 3, 458 features, `maxRecordCount: 2000`):
```
gpio extract arcgis \
  https://www.mozgis.gov.mz/server/rest/services/Data/UnidadesAdministrativasCenso2017/FeatureServer/3 \
  moz_admin3.parquet
```

## Error output

```
Streaming features to temp file...
Fetching features 1-458 of 458...
Traceback (most recent call last):
  ...
  File ".../geoparquet_io/core/arcgis.py", line 170, in _make_request
    return response.json()
  File ".../httpx/_models.py", line 832, in json
    return jsonlib.loads(self.content, **kwargs)
  ...
json.decoder.JSONDecodeError: Expecting value: line 3 column 1 (char 4)
```

## Analysis

The layer has only 458 features with `maxRecordCount: 2000`, so `gpio` attempts to fetch all 458 in one request. The server responds with a non-JSON body (likely an HTML error page), indicating the payload size of the complex Admin 3 polygons overloads the server even though the record count is within its advertised limit.

Two issues in `arcgis.py`:
1. `_make_request` does not catch `JSONDecodeError`, so it propagates as an unhandled crash rather than a retriable error.
2. There is no mechanism to reduce batch size on failure — either automatically or via a `--batch-size` CLI flag.

## Suggested fix

1. **Catch `JSONDecodeError` in `_make_request`** and treat it as a retryable error (server-side failure), similar to how `httpx.RemoteProtocolError` is already handled.

[...]

PR #1136, opened 2026-09-21
feat(api): expose batch_size on extract_arcgis and from_arcgis

cayetanobv:
## The defect

`geoparquet_io/core/arcgis.py` supports a caller-chosen page size end to end:
`arcgis_to_table` (`core/arcgis.py:1379`) and `fetch_all_features`
(`core/arcgis.py:551`) both take `batch_size: int | None = None`, and
`fetch_all_features` sanitizes it against `DEFAULT_PAGE_SIZE` and caps it at the
layer's `maxRecordCount`. The CLI exposes it too: `--batch-size`, with
`IntRange(1, 5000)`, lives at `geoparquet_io/cli/commands/extract.py:332` and is
forwarded at line 484.

The **Python API did not re-declare it**. Neither `extract_arcgis`
(`geoparquet_io/api/table.py:441`) nor `ops.from_arcgis`
(`geoparquet_io/api/ops.py:2259`) had a `batch_size` parameter, so the core
default silently applied and every API caller was pinned to whatever the server
advertised as `maxRecordCount`. There was no way to page smaller from Python.

That gap is invisible to the repo's own drift guards: `extract arcgis` is in the
`_NETWORK` skip set of `tests/test_cli_api_call_parity_scaffold.py:1483`, the one
suite designed to catch "a parameter the API omits entirely so the core default
silently applies".

## Evidence: this blocked a real publication

In `CartoDB/portolan-pipeline#61`, mirroring Norwegian national geodata, five
ArcGIS layers failed at the server's advertised page size with `JSON Error 500`
or 60-second timeouts. A collaborator proved by direct extraction that these
settings complete:

| Layer | Proven setting | Result |
|---|---|---|
| naturvernomrader | batch-size 100 | 3,512 features |
| bestander-laks | batch-size 50 | 479 features |
| bestander-sjoorret | batch-size 50, timeout 180 | 455 features |
| fjellskred-tsunami-faresoner | batch-size 10 | 55 features |
| vannkraft-magasin | batch-size 100 | 1,542 features |

All five were withdrawn from the first publication wave purely because the batch
size could not be passed through the API path the pipeline uses.

## What changed

- `geoparquet_io/api/table.py` — `extract_arcgis`: added keyword-only
  `batch_size: int | None = None` beside `limit`, mirroring the core parameter
  order; documented in the Args section; forwarded to `arcgis_to_table`. Its
  signature is already fully keyword-only, so the position is free.
- `geoparquet_io/api/ops.py` — `from_arcgis`: added `batch_size: int | None =
  None` **last**, not beside `limit`. Its parameters are positional-or-keyword,
  so inserting mid-signature would silently re-map existing positional calls.
  This also matches the repo's own precedent: d9c9282 appended
  `max_allowable_offset` at the end rather than grouping it.
- `docs/api/python-api.md` — one row in the `extract_arcgis` parameter table.
- `tests/test_arcgis.py` — new `TestApiBatchSize`, following the existing
  `TestApiOutputCrs` pattern (patch `geoparquet_io.core.arcgis.arcgis_to_table`,
  assert on `call_args.kwargs`): four tests covering `batch_size=50` reaching the
  core extractor and an omitted `batch_size` still forwarding `None`, for both
  front doors.

**The default is unchanged.** Omitting `batch_size` forwards `None`, exactly the
value the core received before this PR, so the emitted requests are identical.

## What I deliberately did **not** change

- **No CLI change.** `--batch-size` already exists (since #382) and is already
  wired. The premise I was handed said the CLI lacked it; that is not what the
  code shows.
- **No core change.** `arcgis_to_table`, `fetch_all_features` and
  `_get_reduced_batch_size` are untouched, as is the default.
- **No parity-allowlist edit.** `extract arcgis` is skipped as a network command
  in the call-parity scaffold and had no `KNOWN_DIVERGENCES` /
  `KNOWN_PARITY_GAPS` entry for `batch_size`, so there was nothing to delete.

## Why the reduction ladder did not rescue these five layers

`_get_reduced_batch_size` (`core/arcgis.py:117`) steps down
`BATCH_SIZE_FALLBACKS = [1000, 500, 100, 50, 10, 1]`, which covers every proven
setting in the table above. The ladder did not fire because of its **trigger**,
not its rungs:

1. Both call sites — sequential (`core/arcgis.py:640`) and parallel
   (`core/arcgis.py:731`) — enter the ladder only from `except
   BatchTooLargeError`. The parallel path's other handler is a bare `except
   Exception` that re-raises as `RemoteAccessError`, fail-fast.
2. `BatchTooLargeError` has exactly **one** raise site in the whole package:
   `core/http_retry.py:170`, inside the `json.JSONDecodeError` branch. It is
   raised only when the body is not JSON at all (an HTML error or WAF block
   page) *and* a `batch_size` was supplied.
3. An ArcGIS server rejecting an oversized or geometrically expensive page
   normally answers **HTTP 200 with a valid JSON error envelope**
   (`{"error": {"code": 500, "message": "Error performing query operation."}}`).
   That parses fine, so `fetch_features_page` routes it to
   `_handle_arcgis_response` (`core/arcgis.py:184`), which raises
   `GeoParquetError` for every non-auth code. Fatal, no retry, no reduction.
4. The timeout failures never reach the ladder either. `httpx.TimeoutException`
   is caught in `core/http_retry.py:186` and retried at the **same** batch size
   until `max_retries` is exhausted.

`fjellskred-tsunami-faresoner` is the clean proof: 55 features total, so a single
page at the advertised size, which returned JSON error 500 and killed the
extraction — while batch-size 10 completes. Complexity, not row count, made the
advertised page fail.

So the `--batch-size` help text's promise that it "Auto-reduces on server errors"
holds only for non-JSON responses. This is a genuine gap, and it is **already
tracked as #1134** (verified: "fix(extract arcgis): JSON Error 500 page responses
bypass the adaptive batch-size fallback", open, with the same Norway evidence
plus a Wallonia probe matrix). I am deliberately **not** widening the trigger
here — classifying which ArcGIS error codes indicate page pressure versus a
genuine semantic error is a separate, riskier change. This PR only makes the
manual escape hatch reachable from Python, which is what #1134's affected layers
need today.

## Quality gates actually run

`pixi` is not installed on this machine, but this repo documents **uv**
(`CLAUDE.md`: "uv only"), so the real documented gates ran, not a fallback.

| Gate | Command | Result |
|---|---|---|
| Install | `uv sync --all-extras` | pass |
| pre-commit (commit stage) — ruff, ruff-format, codespell, actionlint, zizmor, import-linter, no-click-echo, duckdb-antipatterns, check-api-for-cli, and 11 more | `uv run pre-commit run --all-files` | pass, 20/20 hooks |
| pre-commit (pre-push stage) | `uv run pre-commit run --all-files --hook-stage pre-push` | pass |
| Docs drift | `uv run python scripts/doc_sync.py --check` | pass, all up to date |
| ArcGIS suite | `uv run pytest tests/test_arcgis.py -m "not network"` | 127 passed, 9 deselected |
| CLI/API parity guards + API suite | `uv run pytest tests/test_cli_api_default_parity.py tests/test_cli_api_call_parity_scaffold.py tests/test_api.py` | 334 passed, 1 skipped, 1 xfailed |
| Full non-network suite | `uv run pytest -m "not network and not slow and not meta" -n 4` | 8529 passed, 76 skipped, 6 xfailed (20m14s) |

Not run: the `network`-marked and `slow`-marked tests (both need live services),
the coverage/diff-cover leg, and the notebook and docs-example legs. CI covers
those on this PR.

## Reviewer notes

- `ops.from_arcgis` takes `batch_size` last, unlike core and unlike
  `extract_arcgis`. That asymmetry is intentional (positional-call safety) and is
  called out in its docstring. Say the word if you would rather take the
  positional break and group it with `limit`.

Closes nothing on its own; unblocks the five NVE layers in
`CartoDB/portolan-pipeline#61`. Related: #1134.


geoparquet_io/api/table.py
```diff
@@ -451,6 +451,7 @@ def extract_arcgis(
     include_cols: str | None = None,
     exclude_cols: str | None = None,
     limit: int | None = None,
+    batch_size: int | None = None,
```
cayetanobv, line 460: **Invalid `batch_size` is swallowed instead of rejected — the API path silently does the opposite of what the caller asked.**

The CLI guards this with `IntRange(min=1, max=5000)` before any network call. Neither API front door validates it, and core does not either: `fetch_all_features` (core/arcgis.py:599) has `sanitized_batch = batch_size if batch_size and batch_size > 0 else DEFAULT_PAGE_SIZE`, so:

- `extract_arcgis(url, batch_size=0)` or `batch_size=-50` → silently pages at `min(2000, maxRecordCount)`, i.e. the full server page the caller was trying to escape (the #1134 failure mode), with no error.
- `batch_size="50"` (a value read from a pipeline config) → `TypeError: '>' not supported between 'str' and 'int'` raised deep in the paginator, *after* the layer-info and count-probe requests have already gone out.

`max_workers` right above it gets `raise ValueError("max_workers must be at least 1")`. Suggest the same guard next to it in `fetch_all_features` (one place covers the CLI and both API doors), or `if batch_size is not None and batch_size < 1: raise ValueError(...)` here and in `ops.from_arcgis`.

tests/test_arcgis.py
```diff
@@ -1311,6 +1311,52 @@ def test_extract_arcgis_forwards_max_allowable_offset(self, mock_to_table):
         assert mock_to_table.call_args.kwargs["max_allowable_offset"] == 0.005
 
 
+class TestApiBatchSize:
```
cayetanobv, line 1356: **Altitude: this patches one missing parameter but leaves the guard that let it go missing.**

The PR body is right that the drift suites cannot see this class of bug, but the reasons are both fixable in this PR:

1. `tests/test_cli_api_default_parity.py:222` — `if pname not in api_defaults: continue`. A CLI option the API omits entirely is skipped silently, with no allowlist entry required. That is why `batch_size` never showed up, and why `ops.from_arcgis` can still omit `token_file`, `username`, `password`, `portal_url` (all `gpio extract arcgis` options) with nothing watching.
2. `tests/test_cli_api_call_parity_scaffold.py:1483` skips `extract arcgis` as `_NETWORK` ("not reachable offline"), but that is stale since #1042: `stub_arcgis_service(FakeTransport.install(monkeypatch))` in this very file "routes a whole FeatureServer layer ... so a CLI or API call runs end to end with no network and the request it issued stays assertable".

Either turning the `continue` into a recorded divergence (with a `KNOWN_*` allowlist for the intentional ones) or writing the `extract arcgis` call-parity case over the fake transport would have caught `batch_size`, and will catch the next one. Right now the fix is a special case on top of a guard that still has the hole.

tests/test_arcgis.py
```diff
@@ -1311,6 +1311,52 @@ def test_extract_arcgis_forwards_max_allowable_offset(self, mock_to_table):
...
+    """The page size must be reachable from the Python API, not just the CLI.
+
+    ``gpio extract arcgis --batch-size`` has existed since #382, but the API
+    front doors omitted the parameter, so every API caller was pinned to the
+    server's advertised ``maxRecordCount`` -- fatal for layers the server
+    cannot serialize a full page of (see #1134).
+    """
+
+    @patch("geoparquet_io.core.arcgis.arcgis_to_table")
```
cayetanobv: **These four tests prove forwarding to a mock, not that the page size changes.**

All four patch `arcgis_to_table` and assert on `call_args.kwargs`. That passes even if core later clamps or ignores the value (e.g. a change to `sanitized_batch` / `max_batch` in `fetch_all_features`) — nothing here observes `resultRecordCount` on the wire. The offline harness in this file already makes that assertion cheap:

```python
http = stub_arcgis_service(FakeTransport.install(monkeypatch), count=120)
extract_arcgis("https://example.com/FeatureServer/0", batch_size=50)
assert [r.params["resultRecordCount"] for r in feature_page_requests(http)] == ["50", "50", "20"]
```

(and the same through `ops.from_arcgis`). That is the property the five NVE layers depend on.

Separately, the four bodies are copy-paste with one identifier and one literal varying; `@pytest.mark.parametrize("door", [ops.from_arcgis, extract_arcgis])` × `("batch_size", [50, None])` is the same coverage in one test.

docs/api/python-api.md
```diff
@@ -238,6 +238,7 @@ gpio.extract_arcgis(
 | `include_cols` | str | Comma-separated columns to include |
 | `exclude_cols` | str | Comma-separated columns to exclude |
 | `limit` | int | Maximum number of features |
+| `batch_size` | int | Features per request (default: server's `maxRecordCount`); lower it for layers the server cannot serialize a full page of |
```
cayetanobv: **The `ops.from_arcgis` reference row in this same file was not updated.**

`docs/api/python-api.md:1677` still reads:

```
| `ops.from_arcgis(service_url, token=None, where='1=1', bbox=None, include_cols=None, exclude_cols=None, limit=None)` | Fetch ArcGIS Feature Service |
```

This PR adds `batch_size` to `from_arcgis` and edits this file, but a reader of the `ops` reference cannot discover the parameter the PR exists to expose. (The row also predates `max_workers`, `output_crs`, `max_allowable_offset`, `repair_geometry` and `timeout`; since you are in here, worth bringing the whole row up to the real signature.)

cayetanobv: **CLAUDE.md: `docs/guide/` was not updated for an API change, and there is no CLI + Python example for batch size anywhere.**

`CLAUDE.md` → Git Workflow: "**PRs**: Update `docs/guide/` and `docs/api/python-api.md` if API changed." and Critical Rules: "**All documentation needs CLI + Python examples** - Use tabbed format".

Only `python-api.md` was touched. In `docs/guide/extract.md` the sole mention is prose at line 1202 ("Combine `--timeout` with `--batch-size` (smaller pages) ..."); there is no `=== "CLI"` / `=== "Python"` block for it, so a Python user reading the guide still cannot see that `batch_size=` exists. The PR body already contains the perfect example — the NVE table — e.g. under "Slow servers and timeouts":

```bash
gpio extract arcgis "<url>" out.parquet --batch-size 50 --timeout 180
```
```python
gpio.extract_arcgis("<url>", batch_size=50, timeout=180).write("out.parquet")
```

geoparquet_io/api/ops.py
```diff
@@ -2291,6 +2292,10 @@ def from_arcgis(
             output-CRS units (degrees on the default WGS84 path).
         timeout: Per-request HTTP timeout in seconds (default 60). Increase for
             layers with very large/complex geometries that are slow to serialize.
+        batch_size: Features requested per page. Default None uses the server's
+            advertised maxRecordCount. Lower it for layers the server cannot
+            serialize a full page of (JSON error 500, or a request timeout).
+            Appended last to keep the existing positional order stable.
```
cayetanobv: **Reviewer rationale is in the user-facing docstring, and the two front doors now disagree on where `batch_size` sits.**

"Appended last to keep the existing positional order stable." is a note to maintainers, not to a caller; it will render in API docs. Move it to a `#` comment beside the parameter.

The underlying asymmetry is worth a decision rather than a footnote: `extract_arcgis` (keyword-only) groups `batch_size` with `limit`/`max_workers`, matching core `arcgis_to_table`; `from_arcgis` puts it after `timeout`. Every future parameter faces the same fork. A `*` after `service_url` in `from_arcgis` (as `extract_arcgis` already has) is one deliberate `!` break that ends the problem; otherwise please state the convention (ops appends, table groups) in a comment so the next PR does not re-litigate it.

cayetanobv (commented):
**Recommendation: approve after the validation guard and the docs rows; the rest is judgment.** The forwarding itself is correct and minimal, CI is green across the matrix, and the four new tests pass locally. Six inline comments are attached; here is how I would rank them.

**Should land before merge**

1. **Reject `batch_size < 1` before any network call** (`api/table.py:454`). The CLI already refuses it via `IntRange(1, 5000)`; the API is now the only path where `0`, a negative, or a string from a pipeline config reaches core, and core silently swaps it for `min(2000, maxRecordCount)`, which is exactly the page size the caller was trying to escape. One `ValueError` next to the existing `max_workers` guard in `fetch_all_features` (core/arcgis.py:591) covers all three front doors. I know the PR body says "no core change", but this is the natural home; the alternative is the same two lines in both API functions.
2. **`ops.from_arcgis` reference row** (`docs/api/python-api.md:1677`) still shows a seven-parameter signature. Whoever reads the ops table cannot find the parameter this PR exists to add.
3. **A tabbed CLI + Python example for batch size in `docs/guide/extract.md`.** The "Slow servers and timeouts" section already says to combine `--timeout` with `--batch-size`; its Python tab is the place, and the NVE `batch_size=50, timeout=180` pairing from the PR body is the ready-made example.

**Worth a decision, not a blocker**

4. **Docstring wording.** "Default None uses the server's advertised maxRecordCount" is really `min(2000, maxRecordCount)` (`DEFAULT_PAGE_SIZE = 2000` at core/arcgis.py:54). A server advertising 5000 still gets 2000. The CLI help has the same imprecision, so this is inherited, but the new docstrings repeat it twice.
5. **The maintainer note in the `from_arcgis` Args docstring** ("Appended last to keep the existing positional order stable") will render in the API docs; a `#` comment says the same thing to the right audience. The parameter-order fork between the two front doors is real and your call; either a `*` in `from_arcgis` as a deliberate `!` break, or a one-line stated convention so the next parameter does not re-argue it.

**Optional hardening**

6. **Tests observe a mock, not the wire.** `stub_arcgis_service` + `feature_page_requests` in the same file make `resultRecordCount == ["50","50","20"]` a three-line assertion, and a parametrize over door × value collapses the four bodies to one. Related: the two drift guards that let this slip (the silent `continue` at `test_cli_api_default_parity.py:222` and the stale `_NETWORK` skip for `extract arcgis` at `test_cli_api_call_parity_scaffold.py:1483`, which predates the offline fake transport) are still open, so the next omitted parameter is equally invisible. Fine as a follow-up issue if you would rather keep this PR narrow.

Also: the PR is still a draft, so CodeRabbit skipped it; flip it when ready.

cayetanobv:
Review follow-ups landed in d8b6778: (1) `batch_size < 1`, strings, floats and bools are refused in `arcgis_to_table` before any request; (2) the `ops.from_arcgis` reference row carries its real signature; (3) tabbed CLI + Python batch-size example in `docs/guide/extract.md`; (4) docstrings say `maxRecordCount` capped at 2000; (5) the positional-order rationale moved to a `#` comment stating the convention; (6) the tests now assert `resultRecordCount` on the wire through both front doors. The two drift guards (silent `continue` in the default-parity suite and the stale `_NETWORK` skip for `extract arcgis`) are left for a follow-up.

